PR #14, opened 2020-09-23
Fix indentation in resourceratios template

tobylo:
Corrected the indentation and moved target as the first property which makes it easier to spot, as well as align with the test of the templates.


sozercan (approved):
LGTM

PR #17, opened 2020-10-08
Deny repos other then specified. This for controller kinds i.e. : deployments, statefulsets, DaemonSet..

ipnextgen:


maxsmythe:
Thanks for the contribution!

The commits for this PR will need DCO.

I'm also not seeing the CLA check being invoked, though maybe that will be nudged by my comments. That the check isn't running is something we'll need to fix on our end before merging.

Some thoughts on the PR generally....

Gatekeeper's main repo had debated doing something like this before but rejected the notion because label matching would be applied inconsistently. See https://github.com/open-policy-agent/gatekeeper/issues/298#issuecomment-567322270 and the following comments (the whole thread is on point, but those are the most salient bits).

I would like to see an approach where generator resources (e.g. Deployments -> pods) are hydrated outside of constraint templates, both to avoid the label behavior issue and so that all templates automatically reject things like deployments that create bad pods without the need for writing special logic for each template.

It's not clear to me where this code would live. Ideally I'd like to see these checks happen inside a CI/CD pipeline for a few reasons:

* No latency sensitivity
* It's possible to atomically check and reject an entire cluster config, avoiding partial applications
* Building resource expansion into Gatekeeper itself may tightly couple Gatekeeper with the APIs of whatever versions of K8s its meant to be deployed with. In CI/CD pipelines, users can choose which flavors of expansions they want to add, and it would be trivial for them to add their own.

Maybe it'd be possible to do a temporary expansion using some sort of pluggable middleware, that would at least avoid the API coupling issue.

If we do decide to go this route, I think it'd be best to develop the hydration Rego as a library shared across templates (`pods.containers` in this code) so it can be kept up-to-date across templates.

maxsmythe:
Closing, please re-open if there is follow-up.

PR #21, opened 2020-10-14
Add an allow list parameter, allowedSysctls, to the Forbidden sysctl constraint template

shomron:
The `K8sPSPForbiddenSysctls` constraint template allows limiting sysctls available to pods using a deny list. This change adds another option, `allowedSysctls`, which allows a constraint to specify an allow list in addition to, or instead of, the deny list.

The matching logic is as follows:

1. When specified, any sysctl not in the allow list is considered to be forbidden.
2. The allow list can be omitted.
3. The deny list takes precedence.


shomron:
A little bit of background on how PSPs handle a similar concern:

PSPs define `allowedUnsafeSysctls` and `forbiddenSysctls` fields. The semantics are a bit different:
1. If a sysctl is explicitly forbidden, it is blocked.
2. If a sysctl is in the hardcoded safe list, and not explicitly forbidden, it is allowed.
3. If a sysctl is not in the safe list, it must be explicitly allowed in `allowedUnsafeSysctls` to be allowed.

What I'm not 100% sure about is whether this logic runs on only sysctls listed explicitly in the PodSecurityContext, or whether the defaults are also considered in step 1. The relevant code is [here](https://github.com/kubernetes/kubernetes/blob/4000732bec17d16ad183e38774883f8a6ae93fd4/pkg/security/podsecuritypolicy/sysctl/mustmatchpatterns.go).

shomron:
Following up with additional context:

PSP's `allowedUnsafeSysctls` and `forbiddenSysctls` fields only affect kernel parameters specified via a pod's `securityContext.sysctl` field. They do _not_ affect a container's ability to set these parameters directly in any way. A _privileged_ container can set a forbidden sysctl by writing to `/proc/sys/`, regardless of what is set on the pod's `securityContext.sysctl` and regardless of effective PSP restrictions.

The [hardcoded safe list](https://github.com/kubernetes/kubernetes/blob/4000732bec17d16ad183e38774883f8a6ae93fd4/pkg/security/podsecuritypolicy/sysctl/mustmatchpatterns.go#L32-L39) only affects PSP validation of pod `securityContext.sysctl` fields and means that those parameters can be set without a PSP explicitly allowing them.

I believe that the semantics of `allowedSysctls` in this PR has the advantage over the PSP implementation in that it is more explicit and does not merge in a Kubernetes version-specific default set.

In both native PSPs as well as this constraint template, `forbiddenSysctls` takes precedence over `allowedSysctls` and behaves similarly.

shomron:
@ritazh @sozercan @maxsmythe let me know what you think!

maxsmythe:
+1 to the idea of making things explicit. I don't like the idea of trying to make sure a list of "safe" sysctls is up-to-date

ritazh:
+1 on adding `allowedSysctls`

library/pod-security-policy/forbidden-sysctls/samples/psp-forbidden-sysctls/constraint.yaml
```diff
@@ -11,3 +11,5 @@ spec:
     forbiddenSysctls:
     # - "*" # * may be used to forbid all sysctls
     - kernel.*
+    allowedSysctls:
+    - * # allows all sysctls. allowedSysctls is optional.
```
sozercan, line 15:
```suggestion
    - "*" # allows all sysctls. allowedSysctls is optional.
```

sozercan (approved):
Just a minor fix, LGTM otherwise.

sozercan:
It would be great if we can add an allowed example too.

library/pod-security-policy/forbidden-sysctls/README.md
```diff
@@ -1,5 +1,36 @@
 # Forbidden Sysctls security context policy
 
-Forbidden sysctls excludes specific sysctls. You can forbid a combination of safe and unsafe sysctls in the list. To forbid setting any sysctls, use `*` on its own. If a sysctl pattern ends with a `*` character, such as `kernel.*`, it'll match `*` with rest of the sysctl.
+The forbidden sysctls constraint allows one to limit the set of kernel parameters that can be modified by pods. This is accomplished by specifying a combination of allowed and forbidden sysctls using either of two parameters: `allowedSysctls` and `forbiddenSysctls`.
 
-By default, all safe sysctls are allowed. If you wish to use unsafe sysctls, make sure to whitelist `--allowed-unsafe-sysctls` kubelet flag on each node. For example, `--allowed-unsafe-sysctls='kernel.msg*,kernel.shm.*,net.*'`.
+## Parameters
+
+`allowedSysctls`: A list of explicitly allowed sysctls. Any sysctl not in this list will be considered forbidden. '*' and trailing wildcards are supported. If unspecified, no limitations are made by this parameter.
```
ritazh, line 7: Perhaps consider adding something explicit about the difference in behavior for empty vs unspecified.

chrisns:
any chance of getting this merged?

maxsmythe:
@shomron ready to merge?

shomron:
Ready as ever :)

maxsmythe:
@shomron ping :)

superbrothers:
Is there something wrong with this PR? If the changes are ok, why not merge it once and then add a `description`?

maxsmythe:
@shomron are you okay if I take over this PR?

shomron:
By all means 🙂

ordovicia:
Hi! I made a successor PR #253 which adds description for the new field. Please take a look.

sathieu:
Maybe this PR should be closed? https://github.com/open-policy-agent/gatekeeper-library/pull/253 was merged with the same goal.

Ping @maxsmythe  @shomron

PR #24, opened 2020-10-28
Adding GatewayClass namespaces validation

robscott:
We're working on the [next generation of Kubernetes Service APIs](https://github.com/kubernetes-sigs/service-apis) as part of Kubernetes sig-network. We wanted to provide a way for admins to restrict where certain classes of Gateways can be used (ie an external LB should only be created in prod namespaces). We [evaluated a lot of different options here](https://docs.google.com/document/d/1OmtEiT97i2_rwayiw0G2tW9GX9-BYhiwhVruicWertQ/edit), but we've settled on using policy agents like OPA + Gatekeeper for our initial alpha release. 

I'm not actually sure if you're open to community contributions yet, but wanted to start a conversation here. This example could certainly live in our repo, I just wanted to check in with maintainers here to see if it would also make sense in this library. Do you have any plans for something like the experimental directory that existed in helm/charts?

### DIFF
--- a/library/general/gatewayclassnamespaces/constraint.yaml
+++ b/library/general/gatewayclassnamespaces/constraint.yaml
@@ -1,0 +1,14 @@
+apiVersion: constraints.gatekeeper.sh/v1beta1
+kind: GatewayClassNamespaces
+metadata:
+  name: gatewayclassnamespaces
+spec:
+  match:
+    kinds:
+      - apiGroups: ["networking.x-k8s.io"]
+        kinds: ["Gateway"]
+  parameters:
+    gatewayClasses:
+      - name: external-lb
+        namespaces:
+        - prod
--- a/library/general/gatewayclassnamespaces/template.yaml
+++ b/library/general/gatewayclassnamespaces/template.yaml
@@ -1,0 +1,56 @@
+apiVersion: templates.gatekeeper.sh/v1beta1
+kind: ConstraintTemplate
+metadata:
+  name: gatewayclassnamespaces
+spec:
+  crd:
+    spec:
+      names:
+        kind: GatewayClassNamespaces
+      validation:
+        # Schema for the `parameters` field
+        openAPIV3Schema:
+          properties:
+            gatewayClasses:
+              type: array
+              items:
+                type: object
+                properties:
+                  name:
+                    type: string
+                    maxLength: 253
+                    minLength: 1
+                  namespaces:
+                    type: array
+                    items:
+                      type: string
+                required:
+                - value
+                - namespaces
+  targets:
+    - target: admission.k8s.gatekeeper.sh
+      rego: |
+        package gatewayclassnamespaces
+
+        has_key(x, k) {
+          _ = x[k]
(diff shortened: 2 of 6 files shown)


library/general/gatewayclassnamespaces/template.yaml
```diff
@@ -0,0 +1,36 @@
+apiVersion: templates.gatekeeper.sh/v1beta1
+kind: ConstraintTemplate
+metadata:
+  name: gatewayclassnamespaces
+spec:
+  crd:
+    spec:
+      names:
+        kind: GatewayClassNamespaces
```
maxsmythe, line 9: Could we add the schema for the parameters here?

library/general/gatewayclassnamespaces/constraint.yaml
```diff
@@ -0,0 +1,16 @@
...
+metadata:
+  name: gatewayclassnamespaces
+spec:
+  match:
+    kinds:
+      - apiGroups: ["networking.x-k8s.io"]
+        kinds: ["Gateway"]
+  parameters:
+    gatewayClassNamespaces:
+      external-lb:
```
maxsmythe, line 12: The k8s API guidelines discourages use of arbitrarily-keyed maps:

https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#lists-of-named-subobjects-preferred-over-maps

maxsmythe (commented):
Thanks for the submission and the context!

I think we're definitely open to submissions but we'll need to think a bit about what belongs where....

It's clear to me that templates against vanilla Kubernetes resources are a good fit as they are universal.

The question becomes harder to answer the more idiosyncratic the templates become. Should we host templates against Tekton resources? What about new, experimental personal project X?

Most importantly, how do we make sure that templates stay fresh?

TL;DR is that this definitely seems like it would be a fit (native K8s), but we should figure out the answers to the above questions and message them to users appropriately.

library/general/gatewayclassnamespaces/template.yaml
robscott: Updated, thanks for catching that!

library/general/gatewayclassnamespaces/constraint.yaml
robscott: Good call, turned this into a list.

robscott:
@maxsmythe Thanks for the great feedback here! Completely understand if it takes some time to figure out if/where this belongs. We don't mind hosting this in our repo as well, just trying to get a feel for if there's some kind of standard/recommended practice for APIs like this.

maxsmythe:
@ritazh @shomron @sozercan it looks like we've moved on to taking submissions more by default. Thoughts on this PR?

@robscott It's been a while since this PR was created (sorry about that), is this constrainttemplate still current WRT how the Gateway kind works?

robscott:
Thanks for following up on this @maxsmythe! It looks like this needs a rebase + an update to our latest API version, but otherwise it would still be accurate.

PR #31, opened 2020-11-06
 Establish /library and /src directory structure

ctab:
Implementing the directory structure for the [tooling proposal](https://docs.google.com/document/d/1uQlkIBQcgNNyth8o9ufYaVSfUq_JLJGy9fcuW3VB7vU):

1. Moved src.rego and src_test.rego files to `/src` (with same directory structure as `library/`)

2. Moving the sample constraints and any examples to `library/<group>/<policy>/samples/<constraint name>` directory. 

3. Renamed and slight edits to example files:
* Renamed examples to `example_disallowed.yaml`<sup>1</sup> and made an `example_allowed.yaml` for some cases where the example had a comment showing what to switch to make it pass.
* Example resources all now have "disallowed" or "allowed" in the name.
* Some of the examples depended on a pre-existing namespace "production"; I switched these to use "default" so they could be applied as is.

<sup>1</sup> These were originally `disallowed_example.yaml` and `allowed_example.yaml` in the proposal doc; I switched it around when I started making them and realized that `example_disallowed.yaml` and `example_allowed.yaml` would be next to each other when listing the directory contents, whereas the other format would split them around `constraint.yaml`.


maxsmythe (commented):
LGTM, @ritazh @sozercan thoughts?

sozercan:
we are missing some allowed examples for PSPs (like `users` or `volumes`) do we want to get to this in this PR or in a follow up?

ctab:
@sozercan that's true. I was hoping to handle that in a separate PR so it would be easier to see new code vs. existing code.

maxsmythe:
+1 to minimizing deltas in a PR of this scale.

sozercan (approved):

PR #32, opened 2020-11-16
streamline readme

grosser:


README.md
```diff
@@ -1,17 +1,14 @@
...
-    kubectl apply -f template.yaml
-
-## How to contribute to the library
-
-TODO.
+```bash
+cd library/general/httpsonly/
+kubectl apply -f template.yaml
+kubectl apply -f samples/ingress-https-only/constraint.yaml
+kubectl apply -f library/general/httpsonly/sync.yaml
```
sozercan, line 13: I believe this example doesn't need sync directly unless GK is running with OPA cache. Can we add a note here that this is optional?

grosser: done

sozercan (approved):

PR #33, opened 2020-11-16
add testing

grosser:
replaces https://github.com/open-policy-agent/gatekeeper/pull/821

needs https://github.com/open-policy-agent/opa/issues/2562 + https://github.com/open-policy-agent/opa/issues/2139 for nice coverage enforcement (or we can hack it ... I got the code already, but it's in ruby)

### DIFF
--- a/.github/workflows/actions.yaml
+++ b/.github/workflows/actions.yaml
@@ -1,0 +1,22 @@
+name: CI
+on:
+  push:
+    branches: [master]
+  pull_request:
+    branches: [master]
+jobs:
+  build:
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        os: ["ubuntu-latest", "macos-latest"]
+        opa: ["v0.24.0"]
+    name: ${{ matrix.os }} opa ${{ matrix.opa }}
+    steps:
+      - uses: actions/checkout@v2
+      - run: |
+        binary=$([[ "$OSTYPE" == "darwin"* ]] && echo "opa_darwin_amd64" || echo "opa_linux_amd64")
+        sudo curl -L -o /usr/local/bin/opa https://github.com/open-policy-agent/opa/releases/download/${{ matrix.opa }}/$binary
+        sudo chmod +x /usr/local/bin/opa
+        sh test.sh
+
--- a/.github/workflows/actions.yaml
+++ b/.github/workflows/actions.yaml
@@ -15,8 +15,7 @@ jobs:
     steps:
       - uses: actions/checkout@v2
       - run: |
-        binary=$([[ "$OSTYPE" == "darwin"* ]] && echo "opa_darwin_amd64" || echo "opa_linux_amd64")
-        sudo curl -L -o /usr/local/bin/opa https://github.com/open-policy-agent/opa/releases/download/${{ matrix.opa }}/$binary
-        sudo chmod +x /usr/local/bin/opa
-        sh test.sh
-
+          binary=$([[ "$OSTYPE" == "darwin"* ]] && echo "opa_darwin_amd64" || echo "opa_linux_amd64")
+          sudo curl -L -o /usr/local/bin/opa https://github.com/open-policy-agent/opa/releases/download/${{ matrix.opa }}/$binary
+          sudo chmod +x /usr/local/bin/opa
+          sh test.sh
--- a/src/general/httpsonly/src.rego
+++ b/src/general/httpsonly/src.rego
@@ -1,8 +1,8 @@
 package k8shttpsonly
 
 violation[{"msg": msg}] {
-  input.review.kind.kind == "Ingress"
-  re_match("^(extensions|networking.k8s.io)$", input.review.kind.group)
+  input.review.object.kind == "Ingress"
+  re_match("^(extensions|networking.k8s.io)/", input.review.object.apiVersion)
   ingress := input.review.object
   not https_complete(ingress)
   msg := sprintf("Ingress should be https. tls configuration and allow-http=false annotation are required for %v", [ingress.metadata.name])
--- a/test.sh
+++ b/test.sh
@@ -1,0 +1,14 @@
+#!/bin/bash
+
+set -eu
+
+for folder in src/*/*
+do
+  src=$folder/src.rego
+  test=$folder/src_test.rego
+  # TODO: enforce coverage
+  # needs https://github.com/open-policy-agent/opa/issues/2562 to see what was not covered
+  # needs https://github.com/open-policy-agent/opa/issues/2139 to see verbose output when using coverage
+  echo "opa test $src $test"
+  opa test $src $test
+done
--- a/README.md
+++ b/README.md
@@ -14,7 +14,17 @@ For example
 
 ## How to contribute to the library
 
-If you have a policy you would like to contribute to the library, please feel free to submit a pull request. Each new policy contribution should contain the following:
+### New policy
+
+If you have a policy you would like to contribute, please submit a pull request.
+Each new policy should contain:
 * A constraint template with a `description` annotation and the parameter structure, if any, defined in `spec.crd.spec.validation.openAPIV3Schema`
 * One or more sample constraints, each with an example of an allowed (`example_allowed.yaml`) and disallowed (`example_disallowed.yaml`) resource.
 * The rego source, as `src.rego` and unit tests as `src_test.rego` in the corresponding subdirectory under `src/`
+
+### Development
+
+* policy code and tests are maintained in `src/` folder and then manually copied into `library/`
+* run all tests with `./test.sh`
+* run single test with `opa test src/<folder>/src.rego src/<folder>/src_test.rego --verbose`
+* print results with `trace(sprintf("%v", [thing]))`
--- a/library/general/httpsonly/template.yaml
+++ b/library/general/httpsonly/template.yaml
@@ -16,8 +16,8 @@ spec:
         package k8shttpsonly
 
         violation[{"msg": msg}] {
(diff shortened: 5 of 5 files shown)


.github/workflows/actions.yaml
```diff
@@ -0,0 +1,20 @@
...
+  build:
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        os: ['ubuntu-latest', 'macos-latest']
+    name: ${{ matrix.os }}
+    steps:
+    - uses: actions/checkout@v2
+    - run:
+        sudo curl -L -o /usr/local/bin/opa https://github.com/open-policy-agent/opa/releases/download/v0.21.0/opa_linux_amd64
```
sozercan, line 17: can you make OPA version a variable?

test.sh
```diff
@@ -0,0 +1,29 @@
+#!/bin/bash
+
+set -eu
+
+# folder we create the code in + cleanup
+tmp=$(mktemp -d)
+trap "rm -fr $tmp" 0 2 3 15
+
+for yaml in library/*/*/template.yaml
```
sozercan, line 9: I believe idea is to generate yamls from rego, we might not need to go through yaml at all cc @ctab

grosser: there are no .rego files here, so would prefer to get this merged as is, separate PR can extract regos/deal with template building

.github/workflows/actions.yaml
grosser: done

test.sh
ctab: Hi @grosser,
Sorry for the confusion as we have been restructuring this repo. We do already have rego test files (and rego source files) for the policies in this repo in the `src/` directory. Example: `https://github.com/open-policy-agent/gatekeeper-library/blob/master/src/general/requiredlabels/src_test.rego`
Can you modify this script to use the existing rego source and test files?

grosser:
all done, looks better ?

.github/workflows/actions.yaml
```diff
@@ -0,0 +1,21 @@
...
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        os: ['ubuntu-latest', 'macos-latest']
+        opa: ['v0.21.0']
+    name: ${{ matrix.os }} opa ${{ matrix.opa }}
+    steps:
+    - uses: actions/checkout@v2
+    - run:
+        sudo curl -L -o /usr/local/bin/opa https://github.com/open-policy-agent/opa/releases/download/{{ matrix.opa }}/opa_linux_amd64
```
sozercan, line 18: you need to change to `${{ matrix.opa }}` and this will not work in macOS since this is a linux binary

library/general/requiredlabels/requiredlabels.test.rego
```diff
@@ -0,0 +1,12 @@
+package k8srequiredlabels
```
ctab, line 1: This duplicates the existing test file `src/general/requiredlabels/src_test.rego`.

src/general/httpsonly/src.rego
```diff
@@ -1,8 +1,8 @@
 package k8shttpsonly
 
 violation[{"msg": msg}] {
-  input.review.kind.kind == "Ingress"
-  re_match("^(extensions|networking.k8s.io)$", input.review.kind.group)
+  input.review.object.kind == "Ingress"
```
ctab, line 4: `input.review.kind` contains these values; there's no need for this change (likewise with the rego in the template)

grosser: the test says otherwise:
```
  out = {
    "object": {
      "kind": "Ingress",
      "apiVersion": "extensions/v1beta1",
```
so is the test wrong and this is not a kubernetes object but something else ?

library/general/requiredlabels/requiredlabels.test.rego
grosser: fixed

.github/workflows/actions.yaml
grosser: thx, fixed

src/general/httpsonly/src.rego
grosser: it looks to me that all these policies are for kubernetes objects, and that therefore the old `input.review.kind.group` was wrong
if that's not the case, we should call that out "# policy for xyz system" and fix the tests

grosser: @ctab I can merge the PR without fixing either the test or the source, so please lmk what solution you prefer

grosser: @ctab I can't merge the PR without fixing either the test or the source, so please lmk what solution you prefer

ctab: It does appear the existing test is wrong. Thanks for catching that. Could you add this to the review in the test?  (The object passed by Kubernetes contains these values so the source is valid.)

```
  "kind": {
    "group": "extensions",
    "version": "v1beta1",
    "kind": "Ingress"
  }
```

grosser: ah yes, input.review.kind would be that, but the object has kind and apiVersion

grosser: ... I actually prefer the current code because it only uses the standard kubernetes objects that users can see via `get -o json` so it would be easier to debug/understand ?

.github/workflows/actions.yaml
```diff
@@ -0,0 +1,21 @@
...
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        os: ['ubuntu-latest', 'macos-latest']
+        opa: ['v0.21.0']
+    name: ${{ matrix.os }} opa ${{ matrix.opa }}
+    steps:
+    - uses: actions/checkout@v2
+    - run:
+        sudo curl -L -o /usr/local/bin/opa https://github.com/open-policy-agent/opa/releases/download/${{ matrix.opa }}/opa_linux_amd64
```
sozercan, line 18: still needs a darwin binary for macOS

grosser: done

src/general/httpsonly/src.rego
ctab: I can appreciate that. My preference would be to change the test since the scope of this PR is getting testing to work and the code is working as-is. But this is fine also.

grosser:
@sozercan good to go ?

sozercan (approved):
LGTM. There was a minor indentation issue in the yaml.

test.sh
```diff
@@ -0,0 +1,14 @@
...
+
+for folder in src/*/*
+do
+  src=$folder/src.rego
+  test=$folder/src_test.rego
+  # TODO: enforce coverage
+  # needs https://github.com/open-policy-agent/opa/issues/2562 to see what was not covered
+  # needs https://github.com/open-policy-agent/opa/issues/2139 to see verbose output when using coverage
+  echo "opa test $src $test"
+  opa test $src $test
```
ctab, line 13: I would suggest using the opa docker image at the version currently used in Gatekeeper, rather than the locally installed binary which may not be current.
e.g.
```
  IMAGE_OPA=openpolicyagent/opa
  IMAGE_OPA_VERSION=0.24.0
```
...
```
  docker run -it -v "$(pwd)/${folder}":/work "${IMAGE_OPA}":"${IMAGE_OPA_VERSION}" test -v /work/src.rego /work/src_test.rego
```

ctab: Actually just saw the github workflow downloads it so that's fine. But it looks like 0.24 is the current version, could you change it to that?

grosser: updated

ctab: I still see 0.21.0 in actions.yaml instead of 0.24.0.
Also @sozercan had pushed[ a commit to this branch with formatting fixes ](https://github.com/grosser/gatekeeper-library/commit/c52df57652043fdcce01a3665c1d16960e3ba770)that appears to have been lost. Do you mind recreating that along with the version update?
Everything LGTM otherwise.

grosser: updated and brought back the formatting changes

.github/workflows/actions.yaml
```diff
@@ -0,0 +1,22 @@
...
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        os: ["ubuntu-latest", "macos-latest"]
+        opa: ["v0.24.0"]
+    name: ${{ matrix.os }} opa ${{ matrix.opa }}
+    steps:
+      - uses: actions/checkout@v2
+      - run:
+        binary=$([[ "$OSTYPE" == "darwin"* ]] && echo "opa_darwin_amd64" || echo "opa_linux_amd64")
```
sozercan, line 18: these needs to be indented

.github/workflows/actions.yaml
```diff
@@ -0,0 +1,22 @@
...
+  build:
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        os: ["ubuntu-latest", "macos-latest"]
+        opa: ["v0.24.0"]
+    name: ${{ matrix.os }} opa ${{ matrix.opa }}
+    steps:
+      - uses: actions/checkout@v2
+      - run:
```
sozercan, line 17: `run: |`

.github/workflows/actions.yaml
grosser: they are indented 2-space ... need more ?
... maybe simpler to suggest the formatting you want to see

sozercan: pushed to your branch. just for future reference, you can change `on` to be `on: [push, pull_request]` for testing purposes and test it on your branch.

ctab (approved):

grosser:
thx!

PR #34, opened 2020-12-03
Add integration test via example yaml files in each policy direcotry

### LINKED ISSUE #30: Integration testing via example yaml files in each policy directory
We can add integration testing based on each of the policy directories' templates, constraints, and example resource yaml files. Using the standard from https://github.com/open-policy-agent/gatekeeper-library/issues/25, we can require that all resources containing "_allowed*.yaml" pass and "_disallowed*.yaml" are rejected. (We'll need to ensure those files exist first -- https://github.com/open-policy-agent/gatekeeper-library/issues/26)

We have a couple options here. The kpt tool has a [Gatekeeper resource validator](https://googlecontainertools.github.io/kpt/guides/consumer/function/catalog/validators/) to evaluate a resource, template, and constraint against the OPA client. We can also spin up a kind cluster (similar to in the Gatekeeper repo) and apply the resources there. I imagine the kpt approach is faster but kind more representative of real-world usage.

Kevinma1995:
The integration test is done by spinning up a kind cluster (with make integration-bootstrap) and then iterate through each of the policies in the library folder to apply resources and verify results (with make test-integration)

Fixes #30


ctab:
Hey Kevin, sorry for the delay in reviewing this. This is really helpful, thanks for doing this! 
Can you update the branch, and also, rename the new file `library/general/externalip/samples/allowed-ip/example_denied.yaml` to `example_disallowed.yaml` so that it is picked up by the script?

Kevinma1995:
all done, please give it another pass

.github/workflows/workflow.yaml
```diff
@@ -0,0 +1,27 @@
...
+        run: |
+          mkdir -p $GITHUB_WORKSPACE/bin
+          echo "$GITHUB_WORKSPACE/bin" >> $GITHUB_PATH
+          make integration-bootstrap
+          make deploy
+
+      - name: Run integration test
+        run: |
+          make test-integration
+          echo -e '\n\n======= manager logs =======\n\n'
```
sozercan, line 26: can we use this so we can save all the logs? https://github.com/open-policy-agent/gatekeeper/blob/master/.github/workflows/workflow.yaml#L104-L115

test/bats/sync.yaml
```diff
@@ -0,0 +1,26 @@
+apiVersion: config.gatekeeper.sh/v1alpha1
+kind: Config
+metadata:
+  name: config
+  namespace: "gatekeeper-system"
+spec:
+  sync:
+    syncOnly:
```
sozercan, line 8: I think we just need `Ingress` and `Service` here for uniquness tests?

Makefile
```diff
@@ -0,0 +1,26 @@
...
+	sudo apt-get -o Acquire::Retries=30 update && sudo apt-get -o Acquire::Retries=30 install -y bats
+	# Download and install yq
+	sudo snap install yq
+	# Check for existing kind cluster
+	if [ $$(kind get clusters) ]; then kind delete cluster; fi
+	# Create a new kind cluster
+	TERM=dumb kind create cluster --image kindest/node:${KUBERNETES_VERSION} --config=test/kind_config.yaml
+
+deploy:
+	kubectl apply -f https://raw.githubusercontent.com/open-policy-agent/gatekeeper/release-3.1/deploy/gatekeeper.yaml
```
ritazh, line 23: we should make this `release-3.1` into a variable

sozercan, line 23: can we add a matrix to test release-3.2 also?

.github/workflows/workflow.yaml
```diff
@@ -0,0 +1,27 @@
+name: CI
```
sozercan, line 1: can we rename both actions so it's easy to tell them apart?

Kevinma1995: consolidated unit test to workflow.yaml

sozercan (approved):
Thanks for adding this! LGTM

test/bats/test.bats
```diff
@@ -0,0 +1,107 @@
...
+
+  kubectl run temp --generator=run-pod/v1  --image=tutum/curl -- tail -f /dev/null
+  kubectl wait --for=condition=Ready --timeout=60s pod temp
+  kubectl cp ${cert} temp:/cacert
+
+  wait_for_process ${WAIT_TIME} ${SLEEP_TIME} "kubectl exec -it temp -- curl -f --cacert /cacert --connect-timeout 1 --max-time 2  https://gatekeeper-webhook-service.gatekeeper-system.svc:443/v1/admitlabel"
+  kubectl delete pod temp
+}
+
+@test "constrainttemplates crd is established" {
```
maxsmythe, line 41: Should we be using something like gatekeeper-validate here so we don't need to:

* spin up a k8s cluster
* install Gatekeeper
* check for eventual consistency
* deal with cleanup

?

Kevinma1995: Thought around these two approaches is that kind is more representative of real-world usage.

maxsmythe: Fair enough. If the complexity gets to be too much, we can look at switching later.

Sorry for the late reply :/

maxsmythe (approved):
LGTM

sozercan:
@Kevinma1995 ready to merge?

Kevinma1995:
> @Kevinma1995 ready to merge?

Yes!

Issue #35, opened 2020-12-03
Improvement for SELinux Policy 


rawc0der:
Hello,

I am integrating some of these policies with the project I am working on and the rego policy for **K8sPSPSELinuxV2** is not actually adhering to the specification correctly.
Here is the troublemaker:
https://github.com/open-policy-agent/gatekeeper-library/blob/master/library/pod-security-policy/selinux/template.yaml

My findings so far are about the following sections.

### This verification for the `seLinux` context rule field is not implemented
 https://kubernetes.io/docs/concepts/policy/pod-security-policy/#selinux
```
SELinux
MustRunAs - Requires seLinuxOptions to be configured. Uses seLinuxOptions as the default. Validates against seLinuxOptions.
RunAsAny - No default provided. Allows any seLinuxOptions to be specified.

```  

### The verification is now handled only if there is a seLinuxOptions field set. It should validate the profiles only if `MustRunAs` is set.
```
  seLinux:
    rule: MustRunAs
```

### It should take into account the context set from the rule and ignore any profile validation if rule is:
```
  seLinux:
    rule: RunAsAny
```

I can try adjusting the resources and rego file for this policy and submit a PR.

EmandM:
Setting `RunAsAny` for this setting is the same as not applying this constraint. Any applied template & constraint increases the resources that Gatekeeper uses so to keep settings simple, and to encourage best practices, we have no intention of adding `RunAsAny` to this policy.

PR #36, opened 2020-12-03
improvement: enhance selinux policy to account for context rules

rawc0der:
This PR addresses the changes proposed in Issue #35. 

The specification on pod-security-policy for securityContext allows for multiple contexts under `SELinux`, such as `MustRunAs` and `RunAsAny` which should introduce different validation checks for this policy.

```
SELinux 
MustRunAs - Requires seLinuxOptions to be configured. Uses seLinuxOptions as the default. Validates against seLinuxOptions.
RunAsAny - No default provided. Allows any seLinuxOptions to be specified.
```
[Referenced here](https://kubernetes.io/docs/concepts/policy/pod-security-policy/#selinux)

I have changed most of the files and the test suite as well but am still thinking of various test scenarios.
If anyone (@sozercan, @EmandM) has any improvements in mind and has just a bit of time to review, I would be grateful.

### DIFF
--- a/library/pod-security-policy/selinux/samples/psp-selinux-v2/constraint.yaml
+++ b/library/pod-security-policy/selinux/samples/psp-selinux-v2/constraint.yaml
@@ -8,6 +8,7 @@ spec:
       - apiGroups: [""]
         kinds: ["Pod"]
   parameters:
+    seLinuxContext: MustRunAs
     allowedSELinuxOptions:
       - level: s0:c123,c456
         role: object_r
--- a/library/pod-security-policy/selinux/template.yaml
+++ b/library/pod-security-policy/selinux/template.yaml
@@ -73,7 +73,7 @@ spec:
             object[field]
         }
 
-        accept_context(rule, context) = true {
+        accept_context(rule, context) = false {
           rule == "RunAsAny"
         }
 
--- a/library/pod-security-policy/selinux/template.yaml
+++ b/library/pod-security-policy/selinux/template.yaml
@@ -11,6 +11,9 @@ spec:
         # Schema for the `parameters` field
         openAPIV3Schema:
           properties:
+            seLinuxContext:
+              type: string
+              enum: [RunAsAny, MustRunAs]
             allowedSELinuxOptions:
               type: array
               items:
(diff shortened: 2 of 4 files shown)


EmandM:
Adding `RunAsAny` to this policy allows all possible SELinux settings, which is identical to simply not applying the policy. Having a no-op policy (any policy that allows all) will change nothing on the cluster except increasing the resources that Gatekeeper uses. We wish to avoid this behaviour and encourage users to not add a constraint that results in no-op behaviour. Rather than add `RunAsAny`, we prefer for users to not apply the no-op policy.

rawc0der:
Ok. I understand that while that reasoning is actually aligned with the politics of Gatekeeper (no noop policies), I have actually encountered situations where this reasoning is too strict and does not allow flexibility. 
I actually refactored this policy for a use-case where resources are validated `offline`, in a `static analysis` fashion using **Conftest**.
Using the `RunAsAny` option would allow loosening up the policy on certain Namespaces or selected resources.

maxsmythe:
Would you be able to scope this down to appropriate namespaces using `match.namespaceSelector`, `match.excludedNamespaces` or `match.namespaces`?

https://github.com/open-policy-agent/gatekeeper#constraints

rawc0der:
Hey @maxsmythe,
Thanks for the hint. I can, although the use-case I'm more focused on is to be running Conftest validation against static resources and all the policy parameters are passed to Conftest in a single file just like a PodSecurityPolicy configuration. 
For some resources I want to turn off or loosen certain policies.. thus the idea of having `RunAsAny` context mode in the original rego definition.

maxsmythe:
Do you have a simplified example of this policy parameters setup so I can see how these are wired through conftest?

maxsmythe:
Sorry, I accidentally hit "close with comment" instead of "comment"

rawc0der:
NP,

The parameters passed to conftest as a `--data` object might have the following settings, provided there are multiple policy parameters in the same object ( I have just copy pasted from a working setup ):
# Enabled seLinux using MustRunAs
```
---
parameters:
  container-limits:
    cpu: '2'
    memory: 2000Mi
  psp-privileged:
    enabled: true
  psp-selinux:
    seLinuxContext: MustRunAs
    allowedSELinuxOptions:
      - level: s0:c123,c456
        role: object_r
        type: svirt_sandbox_file_t
        user: system_u
```
# Disabled seLinux using RunAsAny, which is actually what I am trying to achieve.
```
---
parameters:
  psp-privileged:
    enabled: false
  psp-selinux:
    seLinuxContext: RunAsAny
```
I understand that having this configuration has the same effect as turning off the policy but can be done easily in Gatekeeper's context. 
Using Conftest is more cumbersome to configure and I needed to leverage `seLinuxContext` property as an enabled flag to allow turning off the policy from configuration which might be read dynamically when doing CI builds, etc.

maxsmythe:
It looks like we dropped the ball following up on this. Feel free to re-open if this is still relevant and we can continue the discussion.

PR #39, opened 2020-12-04
External IP constraints

tallclair:
Add constraint template & samples for restricting service external IPs to an allowlist.

### DIFF
--- a/library/general/externalip/template.yaml
+++ b/library/general/externalip/template.yaml
@@ -1,0 +1,31 @@
+apiVersion: templates.gatekeeper.sh/v1beta1
+kind: ConstraintTemplate
+metadata:
+  name: k8sexternalips
+spec:
+  crd:
+    spec:
+      names:
+        kind: K8sExternalIPs
+      validation:
+        # Schema for the `parameters` field
+        openAPIV3Schema:
+          properties:
+            allowedIPs:
+              type: array
+              items:
+                type: string
+  targets:
+    - target: admission.k8s.gatekeeper.sh
+      rego: |
+        package k8sexternalips
+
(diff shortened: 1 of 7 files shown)


library/general/externalip/template.yaml
```diff
@@ -0,0 +1,31 @@
...
+              items:
+                type: string
+  targets:
+    - target: admission.k8s.gatekeeper.sh
+      rego: |
+        package k8sexternalips
+
+        violation[{"msg": msg}] {
+          input.review.kind.kind == "Service"
+          input.review.kind.group == ""
```
ritazh, line 25: let's make sure the rego here is consistent with the src.rego

tallclair: oops, this was just commented out in the source for debugging. fixed.

library/general/externalip/template.yaml
```diff
@@ -0,0 +1,31 @@
...
+          allowedIPs := {ip | ip := input.parameters.allowedIPs[_]}
```
maxsmythe, line 26: Do we want to use CIDRs here? We could use `net.cidr_contains_matches`:

https://www.openpolicyagent.org/docs/latest/policy-reference/#net

tallclair: Oh neat! I think both would be useful. Is it OK if we proceed with this as-is, and I'll follow up with a PR to add an `allowedCIDRs` parameter?

maxsmythe: That works!

maxsmythe (approved):
LGTM

ritazh (approved):
lgtm

PR #45, opened 2021-01-03
Cve 2020 8554

AbirHamzi:


YuvalAvra:
This policy would deny all usage of LoadBalancers IPs, which may be a bit too restrictive. Consider adding a whitelist option through an allowedUsers parameter. Something similar is done in the [externalIP constraint template](https://github.com/open-policy-agent/gatekeeper-library/tree/master/library/general/externalip) to whitelist IPs (allowedIPs). 

The system service account that normally updates Load Balancer IPs through services/status requests is `system:serviceaccount:kube-system:service-controller`, it should probably be included in the allowedUsers whitelist, if you choose to implement it.

AbirHamzi:
> This policy would deny all usage of LoadBalancers IPs, which may be a bit too restrictive. Consider adding a whitelist option through an allowedUsers parameter. Something similar is done in the [externalIP constraint template](https://github.com/open-policy-agent/gatekeeper-library/tree/master/library/general/externalip) to whitelist IPs (allowedIPs).
> 
> The system service account that normally updates Load Balancer IPs through services/status requests is `system:serviceaccount:kube-system:service-controller`, it should probably be included in the allowedUsers whitelist, if you choose to implement it.

I checked the following scenario and it works fine:
1. Create clusterip service
2. Update the service type to loadbalancer

After few seconds a public ip was assigned to the service, which means the policy has no impact on the service controller work. Please let me know if we have more scenarios to test.

YuvalAvra:
@AbirHamzi The service controller is able to assign a Load Balancer IP because the Rego in template.yaml misses the first assignment of a Load Balancer IP. You can verify that by patching the status of a LoadBalancer service just after creating it, before the service controller managed to assign it a Load Balancer IP.

To prevent this bypass, replace the following in template.yaml
```rego         
request.oldObject.status.loadBalancer.ingress != request.object.status.loadBalancer.ingress
```
With:
```rego
newLoadBalancerIPs := {ips | ips := request.object.status.loadBalancer.ingress[_].ip} - {oldips | oldips := input.review.oldObject.status.loadBalancer.ingress[_].ip}
count(newLoadBalancerIPs) > 0     
```

AbirHamzi:
> @AbirHamzi The service controller is able to assign a Load Balancer IP because template.yaml misses the first assignment of a Load Balancer IP. You can verify that by patching the status of a LoadBalancer service just after creating it, before the service controller managed to assign it a Load Balancer IP.
> 
> To prevent this bypass, replace the following line in template.yaml:
> 
> ```rego
> request.oldObject.status.loadBalancer.ingress != request.object.status.loadBalancer.ingress
> ```
> 
> With:
> 
> ```rego
> newLoadBalancerIPs := {ips | ips := request.object.status.loadBalancer.ingress[_].ip} - {oldips | oldips := input.review.oldObject.status.loadBalancer.ingress[_].ip}
> count(newLoadBalancerIPs) > 0     
> ```

Yes indeed you are right. I'm working on that, thank you for your feedback.

maxsmythe:
Thank you for the submission!

Is this the same CVE that this template was meant to address?

https://github.com/open-policy-agent/gatekeeper-library/pull/39

Referenced in this [k8s issue](https://github.com/kubernetes/kubernetes/issues/97076)

AbirHamzi:
> Thank you for the submission!
> 
> Is this the same CVE that this template was meant to address?
> 
> #39
> 
> Referenced in this [k8s issue](https://github.com/kubernetes/kubernetes/issues/97076)

@maxsmythe yes it's the same CVE, but  restricting service externalIPs/loadbalancerIPs to an allowlist is not enough because the hacker still can patch the service status and intercept the traffic.

library/cve/CVE-2020-8554/template.yaml
```diff
@@ -0,0 +1,36 @@
...
+              type: array
+              items: string
+  targets:
+    - target: admission.k8s.gatekeeper.sh
+      rego: |
+        package k8sblockservicepatch
+
+        violation[{"msg": msg}] {
+            input.review.kind.kind == "Service"
+            input.review.operation == "UPDATE"
```
maxsmythe, line 25: Is there a reason why we only want this template to work during UPDATE requests and not during CREATE or audit?

library/cve/CVE-2020-8554/template.yaml
```diff
@@ -0,0 +1,36 @@
...
+            input.review.subResource == "status"
+            input.review.requestSubResource == "status"
+            user := input.review.userInfo.username
+            allowed := {good | alloweduser:= input.parameters.users[_];  good = identical(user, alloweduser)}
```
maxsmythe, line 29: Is there some reason why certain users should be allowed to set disallowed IPs?

library/cve/CVE-2020-8554/template.yaml
```diff
@@ -0,0 +1,36 @@
...
+            not any(allowed)
+            msg := sprintf("CVE-2020-8554 was detected on service %v/%v by user: %v", [input.review.namespace, input.review.name, user])
```
maxsmythe, line 31: requiring there be a requesting user makes it so that the CVE cannot be validated at rest (e.g. during audit). This may mean pre-existing violations or violations that pass due to webhook failure may remain undetected.

If we merely validate that only acceptable IPs are used, this loophole closes.

maxsmythe (changes requested):
sorry for the slow reply... so the `status` field can be used to actuate changes to the system? I'm a bit surprised as K8s conventions imply status should merely reflect state, not change it (though conventions are not always consistently enforced).

Can this code be added to the template in #39 so that people only need to go to one place to address this CVE?

library/cve/CVE-2020-8554/template.yaml
AbirHamzi: @maxsmythe In fact,  it's because a patch request is an `UPDATE` operation that targets the subresource `status`. The following is an example of `input.review` :

```
{
    "kind": {
      "group": "",
      "version": "v1",
      "kind": "Service"
    },
    "subResource": "status",
    "requestSubResource": "status",
    "name": "mitm-lb",
    "operation": "UPDATE",
    "options": {
      "kind": "UpdateOptions",
      "apiVersion": "meta.k8s.io/v1"
    },
    "resource": {
      "group": "",
      "version": "v1",
      "resource": "services"
    },
    "requestKind": {
      "kind": "Service",
      "group": "",
      "version": "v1"
    },
    "oldObject": {
      "apiVersion": "v1",
      "kind": "Service",
      "metadata": {
        "resourceVersion": "880509",
        "creationTimestamp": "2021-01-02T13:23:08Z",
        "annotations": {
          "kubectl.kubernetes.io/last-applied-configuration": "{\"apiVersion\":\"v1\",\"kind\":\"Service\",\"metadata\":{\"annotations\":{},\"name\":\"mitm-lb\",\"namespace\":\"kubeproxy-mitm\"},\"spec\":{\"ports\":[{\"name\":\"http\",\"port\":80,\"targetPort\":8080},{\"name\":\"https\",\"port\":443,\"targetPort\":8443}],\"selector\":{\"app\":\"echoserver\"},\"type\":\"LoadBalancer\"}}\n"
        },
        "finalizers": [
          "service.kubernetes.io/load-balancer-cleanup"
        ],
        "name": "mitm-lb",
        "namespace": "kubeproxy-mitm",
        "uid": "9bf4b412-6188-4a37-9529-5d763a8ecfbe"
      },
      "spec": {
        "externalTrafficPolicy": "Cluster",
        "ports": [
          {
            "name": "http",
            "protocol": "TCP",
            "port": 80,
            "targetPort": 8080,
            "nodePort": 30379
          },
          {
            "name": "https",
            "protocol": "TCP",
            "port": 443,
            "targetPort": 8443,
            "nodePort": 32661
          }
        ],
        "selector": {
          "app": "echoserver"
        },
        "clusterIP": "10.0.223.26",
        "type": "LoadBalancer",
        "sessionAffinity": "None"
      },
      "status": {
        "loadBalancer": {
          "ingress": [
            {
              "ip": "20.74.10.194"
            }
          ]
        }
      }
    },
    "dryRun": false,
    "_unstable": {
      "namespace": {
        "metadata": {
          "resourceVersion": "832862",
          "creationTimestamp": "2021-01-01T16:09:10Z",
          "annotations": {
            "kubectl.kubernetes.io/last-applied-configuration": "{\"apiVersion\":\"v1\",\"kind\":\"Namespace\",\"metadata\":{\"annotations\":{},\"name\":\"kubeproxy-mitm\"}}\n"
          },
          "managedFields": [
            {
              "apiVersion": "v1",
              "time": "2021-01-01T16:09:10Z",
              "fieldsType": "FieldsV1",
              "fieldsV1": {
                "f:metadata": {
                  "f:annotations": {
                    ".": {},
                    "f:kubectl.kubernetes.io/last-applied-configuration": {}
                  }
                },
                "f:status": {
                  "f:phase": {}
                }
              },
              "manager": "kubectl-client-side-apply",
              "operation": "Update"
            }
          ],
          "name": "kubeproxy-mitm",
          "selfLink": "/api/v1/namespaces/kubeproxy-mitm",
          "uid": "5db33093-ca9e-43cb-b788-fb2161fd4273"
        },
        "spec": {
          "finalizers": [
            "kubernetes"
          ]
        },
        "status": {
          "phase": "Active"
        },
        "kind": "Namespace",
        "apiVersion": "v1"
      }
    },
    "uid": "66641aa3-880c-4cbb-992d-be742106b134",
    "requestResource": {
      "resource": "services",
      "group": "",
      "version": "v1"
    },
    "namespace": "kubeproxy-mitm",
    "userInfo": {
      "username": "masterclient",
      "groups": [
        "system:masters",
        "system:authenticated"
      ]
    },
    "object": {
      "apiVersion": "v1",
      "kind": "Service",
      "metadata": {
        "uid": "9bf4b412-6188-4a37-9529-5d763a8ecfbe",
        "resourceVersion": "880509",
        "creationTimestamp": "2021-01-02T13:23:08Z",
        "annotations": {
          "kubectl.kubernetes.io/last-applied-configuration": "{\"apiVersion\":\"v1\",\"kind\":\"Service\",\"metadata\":{\"annotations\":{},\"name\":\"mitm-lb\",\"namespace\":\"kubeproxy-mitm\"},\"spec\":{\"ports\":[{\"name\":\"http\",\"port\":80,\"targetPort\":8080},{\"name\":\"https\",\"port\":443,\"targetPort\":8443}],\"selector\":{\"app\":\"echoserver\"},\"type\":\"LoadBalancer\"}}\n"
        },
        "finalizers": [
          "service.kubernetes.io/load-balancer-cleanup"
        ],
        "managedFields": [
          {
            "fieldsType": "FieldsV1",
            "fieldsV1": {
              "f:spec": {
                "f:externalTrafficPolicy": {},
                "f:ports": {
                  ".": {},
                  "k:{\"port\":80,\"protocol\":\"TCP\"}": {
                    "f:port": {},
                    "f:protocol": {},
                    "f:targetPort": {},
                    ".": {},
                    "f:name": {}
                  },
                  "k:{\"port\":443,\"protocol\":\"TCP\"}": {
                    ".": {},
                    "f:name": {},
                    "f:port": {},
                    "f:protocol": {},
                    "f:targetPort": {}
                  }
                },
                "f:selector": {
                  ".": {},
                  "f:app": {}
                },
                "f:sessionAffinity": {},
                "f:type": {}
              },
              "f:metadata": {
                "f:annotations": {
                  ".": {},
                  "f:kubectl.kubernetes.io/last-applied-configuration": {}
                }
              }
            },
            "manager": "kubectl-client-side-apply",
            "operation": "Update",
            "apiVersion": "v1",
            "time": "2021-01-02T13:23:08Z"
          },
          {
            "fieldsType": "FieldsV1",
            "fieldsV1": {
              "f:metadata": {
                "f:finalizers": {
                  ".": {},
                  "v:\"service.kubernetes.io/load-balancer-cleanup\"": {}
                }
              }
            },
            "manager": "kube-controller-manager",
            "operation": "Update",
            "apiVersion": "v1",
            "time": "2021-01-02T13:23:12Z"
          },
          {
            "manager": "curl",
            "operation": "Update",
            "apiVersion": "v1",
            "time": "2021-01-02T13:49:53Z",
            "fieldsType": "FieldsV1",
            "fieldsV1": {
              "f:status": {
                "f:loadBalancer": {
                  "f:ingress": {}
                }
              }
            }
          }
        ],
        "name": "mitm-lb",
        "namespace": "kubeproxy-mitm"
      },
      "spec": {
        "selector": {
          "app": "echoserver"
        },
        "clusterIP": "10.0.223.26",
        "type": "LoadBalancer",
        "sessionAffinity": "None",
        "externalTrafficPolicy": "Cluster",
        "ports": [
          {
            "name": "http",
            "protocol": "TCP",
            "port": 80,
            "targetPort": 8080,
            "nodePort": 30379
          },
          {
            "port": 443,
            "targetPort": 8443,
            "nodePort": 32661,
            "name": "https",
            "protocol": "TCP"
          }
        ]
      },
      "status": {
        "loadBalancer": {
          "ingress": [
            {
              "ip": "20.74.17.228"
            }
          ]
        }
      }
    }
  }
```
You can catch this `input.review` only when you make a patch request using `Curl` like the following:

```
curl -k -v -XPATCH  -H "Accept: application/json" -H "Content-Type: application/merge-patch+json" 'http://127.0.0.1:8001/api/v1/namespaces/default/services/mitm-lb/status' -d '{"status":{"loadBalancer":{"ingress":[{"ip": "1.1.1.1"}]}}}'
```

`kubectl patch` can't patch the subresource status because of this [issue](https://github.com/kubernetes/kubernetes/issues/67455).

library/cve/CVE-2020-8554/template.yaml
AbirHamzi: @maxsmythe since a loadbalancer assigns `EXTERNAL-IPs` to a service by means of a patch request that updates `status.loadbalancer.ingress[_].ip`, we need to whitelist the loadbalancer's service account because it is a legitimate operation.

library/cve/CVE-2020-8554/template.yaml
AbirHamzi: @maxsmythe yes indeed, but I think whitelisting LoadBalancerIPs should be handled by the loadbalancer itself (Since we allow only the loadbalancer to assign `EXTERNAL-IPs`, I think we should not care about it). However, I ll add this policy to #39 and make a double check for allowed loadbalancerIPs.

library/cve/CVE-2020-8554/template.yaml
maxsmythe: Sorry for the (very) long response delay. LMK if this PR is still relevant.

One issue with allowing things by username is that this becomes essentially unauditable. If Gatekeeper is down (or this constraint is not yet installed), then we would not be able to detect violations after-the-fact, which is why we'd be interested in looking at the specific IPs allowed vs. who created them.

sathieu:
Is this PR still needed now we have https://github.com/open-policy-agent/gatekeeper-library/tree/master/library/general/externalip?

PR #46, opened 2021-01-07
exclude kube-system namespace in K8sPSPPrivilegedContainer example

katze120:
This is to allow system pods to be created on new nodes during node upgrade, repair, or auto-scaling.


maxsmythe (approved):
LGTM

ritazh (approved):

Issue #49, opened 2021-01-22
runAsNonRoot SecurityContext


sanderma:
The [psp users rule](https://github.com/open-policy-agent/gatekeeper-library/blob/master/src/pod-security-policy/users/src.rego) checks `runAsUser`, but ignores the `runAsNonRoot` SecurityContext.
Is there a specific reason for this?

I'd expect that if a user provides `runAsNonRoot` and leaves the `runAsUser` out, the review object is not denied.
In that case the kubelet will ensure that the image doesn't run as root.

```yaml
apiVersion: v1
kind: Pod
metadata:
  name: nginx-users-disallowed
  labels:
    app: nginx-users
spec:
  securityContext:
    runAsNonRoot: true
  containers:
    - name: nginx
      image: nginx
```

`kubectl describe pod -l app=nginx-users`

```
Normal   Pulling                 2s (x2 over 10s)    kubelet            Pulling image "nginx"
Warning  Failed                  2s                  kubelet            Error: container has runAsNonRoot and image will run as root
Normal   Pulled                  0s (x2 over 2s)     kubelet            Successfully pulled image "nginx"
```

I changed the rule a bit so when you set `MustRunAsNonRoot` it also allows setting the `securityContext/runAsNonRoot`:

```rego
# RunAsUser (separate due to "MustRunAsNonRoot")
get_user_violation(params, container) = msg {
  rule := params.rule
  provided_user := get_field_value("runAsUser", container, input.review)
  not accept_users(rule, provided_user)
  msg := sprintf("Container %v is attempting to run as disallowed user %v. Allowed runAsUser: %v", [container.name, provided_user, params])
}

get_user_violation(params, container) = msg {
  not get_field_value("runAsUser", container, input.review)
  not get_field_value("runAsNonRoot", container, input.review)
  params.rule != "RunAsAny"
  params.rule != "runAsNonRoot"
  msg := sprintf("Container %v is attempting to run without a required securityContext/%v", [container.name, params.rule])
}

get_user_violation(params, container) = msg {
  params.rule = "runAsNonRoot"
  not get_field_value("runAsUser", container, input.review)
  not get_field_value("runAsNonRoot", container, input.review)
  msg := sprintf("Container %v is attempting to run without a required securityContext/runAsNonRoot or securityContext/runAsUser != 0", [container.name])
}
```

Am I missing something, or should I do  a PR with this change (and added tests)? @

maxsmythe:
[...]

PR #50, opened 2021-01-27
Add runAsNonRoot securityContext check

sanderma:
This PR solves [Issue 49](https://github.com/open-policy-agent/gatekeeper-library/issues/49)

- It adds a check for the `runAsNonRoot` securityContext field.
- It adds tests to validate the rule is functioning.

### DIFF
--- a/library/pod-security-policy/users/template.yaml
+++ b/library/pod-security-policy/users/template.yaml
@@ -102,14 +102,12 @@ spec:
 
         get_user_violation(params, container) = msg {
           not get_field_value("runAsUser", container, input.review)
-          not get_field_value("runAsNonRoot", container, input.review)
-          params.rule != "RunAsAny"
-          params.rule != "runAsNonRoot"
-          msg := sprintf("Container %v is attempting to run without a required securityContext/%v", [container.name, params.rule])
+          params.rule = "MustRunAs"
+          msg := sprintf("Container %v is attempting to run without a required securityContext/runAsUser", [container.name])
         }
 
         get_user_violation(params, container) = msg {
-          params.rule = "runAsNonRoot"
+          params.rule = "MustRunAsNonRoot"
           not get_field_value("runAsUser", container, input.review)
           not get_field_value("runAsNonRoot", container, input.review)
           msg := sprintf("Container %v is attempting to run without a required securityContext/runAsNonRoot or securityContext/runAsUser != 0", [container.name])
--- a/library/pod-security-policy/users/template.yaml
+++ b/library/pod-security-policy/users/template.yaml
@@ -102,8 +102,17 @@ spec:
 
         get_user_violation(params, container) = msg {
           not get_field_value("runAsUser", container, input.review)
+          not get_field_value("runAsNonRoot", container, input.review)
           params.rule != "RunAsAny"
-          msg := sprintf("Container %v is attempting to run without a required securityContext/runAsUser. Allowed runAsUser: %v", [container.name, params])
+          params.rule != "runAsNonRoot"
+          msg := sprintf("Container %v is attempting to run without a required securityContext/%v", [container.name, params.rule])
+        }
+
+        get_user_violation(params, container) = msg {
+          params.rule = "runAsNonRoot"
+          not get_field_value("runAsUser", container, input.review)
+          not get_field_value("runAsNonRoot", container, input.review)
+          msg := sprintf("Container %v is attempting to run without a required securityContext/runAsNonRoot or securityContext/runAsUser != 0", [container.name])
         }
 
         accept_users("RunAsAny", provided_user) {true}
--- a/src/pod-security-policy/users/src_test.rego
+++ b/src/pod-security-policy/users/src_test.rego
@@ -213,7 +213,22 @@ test_user_input_container_run_as_rule_non_root_ignore_ranges {
   results := violation with input as input
   count(results) == 0
 }
-
+test_user_input_container_run_as_rule_non_root_runasnonroot_true {
+  input := {
+    "review": review(runAsNonRoot(true), [ctr("cont1", null)], null),
+    "parameters": runAsUser(rule("MustRunAsNonRoot", null))
+  }
+  results := violation with input as input
+  count(results) == 0
+}
+test_user_input_container_run_as_rule_non_root_runasnonroot_false {
+  input := {
+    "review": review(runAsNonRoot(false), [ctr("cont1", null)], null),
+    "parameters": runAsUser(rule("MustRunAsNonRoot", null))
+  }
+  results := violation with input as input
+  count(results) == 1
+}
 
 
 ## runAsGroup ##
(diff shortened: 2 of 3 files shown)


sanderma:
cla/linuxfoundation is signed.

maxsmythe:
Thanks for the PR!

What is the expected config to invoke "run as non root"? The example constraint in the library currently looks like this:

```yaml
kind: K8sPSPAllowedUsers
metadata:
  name: psp-pods-allowed-user-ranges
spec:
  match:
    kinds:
      - apiGroups: [""]
        kinds: ["Pod"]
  parameters:
    runAsUser:
      rule: MustRunAs # MustRunAsNonRoot # RunAsAny 
      ranges:
        - min: 100
          max: 200
    runAsGroup:
      rule: MustRunAs # MayRunAs # RunAsAny 
      ranges:
        - min: 100
          max: 200
    supplementalGroups:
      rule: MustRunAs # MayRunAs # RunAsAny 
      ranges:
        - min: 100
          max: 200
    fsGroup:
      rule: MustRunAs # MayRunAs # RunAsAny 
      ranges:
        - min: 100
          max: 200
```

Did we want to have a "run as any non-root user" constraint look like the following?


```yaml
kind: K8sPSPAllowedUsers
metadata:
  name: psp-pods-allowed-user-ranges
spec:
  match:
    kinds:
      - apiGroups: [""]
        kinds: ["Pod"]
  parameters:
    runAsUser:
      rule:  MustRunAsNonRoot
```

library/pod-security-policy/users/template.yaml
```diff
@@ -102,8 +102,17 @@ spec:
...
           not get_field_value("runAsUser", container, input.review)
+          not get_field_value("runAsNonRoot", container, input.review)
           params.rule != "RunAsAny"
-          msg := sprintf("Container %v is attempting to run without a required securityContext/runAsUser. Allowed runAsUser: %v", [container.name, params])
+          params.rule != "runAsNonRoot"
+          msg := sprintf("Container %v is attempting to run without a required securityContext/%v", [container.name, params.rule])
+        }
+
+        get_user_violation(params, container) = msg {
+          params.rule = "runAsNonRoot"
```
maxsmythe, line 112: I think this should be `MustRunAsNonRoot`?

library/pod-security-policy/users/template.yaml
```diff
@@ -102,8 +102,17 @@ spec:
 
         get_user_violation(params, container) = msg {
           not get_field_value("runAsUser", container, input.review)
+          not get_field_value("runAsNonRoot", container, input.review)
           params.rule != "RunAsAny"
-          msg := sprintf("Container %v is attempting to run without a required securityContext/runAsUser. Allowed runAsUser: %v", [container.name, params])
+          params.rule != "runAsNonRoot"
```
maxsmythe, line 107: `MustRunAsNonRoot`?

library/pod-security-policy/users/template.yaml
```diff
@@ -102,8 +102,17 @@ spec:
 
         get_user_violation(params, container) = msg {
           not get_field_value("runAsUser", container, input.review)
+          not get_field_value("runAsNonRoot", container, input.review)
```
maxsmythe, line 105: I don't think we need to test this here, as the `params.rule != "runAsNonRoot"` line below means that this check will only run for `MustRunAs`, which does require a specific user be provided.

`RunAsNonRoot` with an undefined user is checked in the extra `get_user_violation` you've added below.

src/pod-security-policy/users/src_test.rego
```diff
@@ -213,7 +213,22 @@ test_user_input_container_run_as_rule_non_root_ignore_ranges {
...
-
+test_user_input_container_run_as_rule_non_root_runasnonroot_true {
+  input := {
+    "review": review(runAsNonRoot(true), [ctr("cont1", null)], null),
+    "parameters": runAsUser(rule("MustRunAsNonRoot", null))
+  }
+  results := violation with input as input
+  count(results) == 0
+}
+test_user_input_container_run_as_rule_non_root_runasnonroot_false {
```
maxsmythe, line 224: I think you'd get an unexpected no violations if you added a test:

```
test_user_input_container_must_run_as_rule_runasnonroot_true {
  input := {
    "review": review(runAsNonRoot(true), [ctr("cont1", null)], null),
    "parameters": runAsUser(rule("MustRunAs", [range(100, 200)]))
  }
  count(results) == 1
}
```

In that scenario a user would expect a violation b/c they are requiring a specific user ID range.

maxsmythe (changes requested):

src/pod-security-policy/users/src_test.rego
sanderma: Added test:
```rego
test_user_one_container_run_in_range_seccont_runasnonroot {
  input := { "review": review(null, [ctr("cont1", runAsNonRoot(true))], null), "parameters": user_mustrunas_100_200 }
  count(results) == 1
}
```

library/pod-security-policy/users/template.yaml
sanderma: Like so?

If `runAsUser` is present, it should always be checked, the `accept_users` func sorts out if it's ok:

```rego
get_user_violation(params, container) = msg {
  rule := params.rule
  provided_user := get_field_value("runAsUser", container, input.review)
  not accept_users(rule, provided_user)
  msg := sprintf("Container %v is attempting to run as disallowed user %v. Allowed runAsUser: %v", [container.name, provided_user, params])
}
```

When we ask for a specific user range via `MustRunAs`, `runAsUser` field should be present:

```rego
get_user_violation(params, container) = msg {
  not get_field_value("runAsUser", container, input.review)
  params.rule = "MustRunAs"
  msg := sprintf("Container %v is attempting to run without a required securityContext/runAsUser", [container.name])
}
```

If `MustRunAsNonRoot` and user doesn't add `runAsUser`, or `runAsNonRoot`: fail:

```rego
get_user_violation(params, container) = msg {
  params.rule = "MustRunAsNonRoot"
  not get_field_value("runAsUser", container, input.review)
  not get_field_value("runAsNonRoot", container, input.review)
  msg := sprintf("Container %v is attempting to run without a required securityContext/runAsNonRoot or securityContext/runAsUser != 0", [container.name])
}
```

library/pod-security-policy/users/template.yaml
sanderma: Fixed

library/pod-security-policy/users/template.yaml
sanderma: Fixed

maxsmythe (approved):
This LGTM @EmandM, LGTY?

ritazh (approved):

library/pod-security-policy/users/template.yaml
```diff
@@ -102,8 +102,15 @@ spec:
 
         get_user_violation(params, container) = msg {
           not get_field_value("runAsUser", container, input.review)
-          params.rule != "RunAsAny"
-          msg := sprintf("Container %v is attempting to run without a required securityContext/runAsUser. Allowed runAsUser: %v", [container.name, params])
+          params.rule = "MustRunAs"
+          msg := sprintf("Container %v is attempting to run without a required securityContext/runAsUser", [container.name])
```
sanderma, line 106: I see I left the param field out. The formatting was a bit off, but very useful.

maxsmythe:
Ready to merge?

sanderma:
> Ready to merge?

I don't see any other things to change at this point.

maxsmythe:
Fair enough! Merging.

PR #54, opened 2021-01-29
Add new template to disallow using the latest tag in a container image

arapulido:
It is fairly common policy in Kubernetes to disallow using the `latest` tag in some clusters (especially staging and production).

This PR adds a new template for a policy that will check that tags for images are included, and that the tag is not `latest`


library/general/disallowlatest/template.yaml
```diff
@@ -0,0 +1,43 @@
...
+      names:
+        kind: K8sDisallowLatest
+  targets:
+    - target: admission.k8s.gatekeeper.sh
+      rego: |
+        package k8sdisallowlatest
+
+        violation[{"msg": msg}] {
+          container := input.review.object.spec.containers[_]
+          latest := [re_match(":latest", container.image)]
```
maxsmythe, line 19: Performing a suffix match might be safer here. Either that or we should anchor the regex to the end of the string using "$", though we should strip whitespace if we do that.

library/general/disallowlatest/template.yaml
```diff
@@ -0,0 +1,43 @@
...
+          all(latest)
+          msg := sprintf("container <%v> uses a latest image tag <%v>", [container.name, container.image])
+        }
+
+        violation[{"msg": msg}] {
+          container := input.review.object.spec.containers[_]
+          tag := [re_match(":", container.image)]
```
maxsmythe, line 26: The `contains()` builtin is probably best to use here:

https://www.openpolicyagent.org/docs/latest/policy-reference/#strings

library/general/disallowlatest/template.yaml
```diff
@@ -0,0 +1,43 @@
...
+          not all(tag)
+          msg := sprintf("container <%v> didn't specify an image tag <%v>", [container.name, container.image])
+        }
+
+        violation[{"msg": msg}] {
+          container := input.review.object.spec.initContainers[_]
```
maxsmythe, line 32: We can avoid duplicating logic by creating a list of all containers before iterating over them.

https://github.com/open-policy-agent/gatekeeper-library/blob/ce24dd6802b8c845f80a27731b9095cc0864726f/src/pod-security-policy/read-only-root-filesystem/src.rego

is an example of how this has been done elsewhere.

maxsmythe (changes requested):
Thanks for the PR!

This fills a significant gap.

I'm wondering if we could adjust the constraint template to take a list of disallowed tags as a parameter. That would allow this template to cover both the "latest" use case and any other tags users may want to block.

arapulido:
Thanks a lot for the quick review! I will work on those changes asap

library/general/disallowedtags/template.yaml
```diff
@@ -0,0 +1,45 @@
...
+              items:
+                type: string
+  targets:
+    - target: admission.k8s.gatekeeper.sh
+      rego: |
+        package k8sdisallowedtags
+
+        violation[{"msg": msg}] {
+          container := input_containers[_]
+          tags := [forbid | tag = input.parameters.tags[_] ; forbid = endswith(container.image, tag)]
```
maxsmythe, line 28: Should we create an object that maps disallowed tags so that we don't need to scroll through the list every time when there are multiple containers?

@tsandall is there some way to make sure that a map of tags would be created at the entry point of evaluating the rule so we guarantee we aren't paying the cost of construction * the # of images?

In practice, this may be fine as the # of images and the # of disallowed tags may not be very large.

maxsmythe, line 28: We should add a `:` prefix to the tag to make sure we are denying the full tag and not just a tag suffix.

For example: `myimage:shangrilatest` would be blocked under the current constraint as-written.

maxsmythe (changes requested):

library/general/disallowedtags/template.yaml
arapulido: Very true! I will fix this.

arapulido: Added a commit to address this. I will wait on the response on the object creation to address the other one.

maxsmythe (approved):
LGTM

Thanks for this! Waiting on second reviewer now.

sozercan:
Hi @arapulido 👋 thanks for the PR! Can you copy rego and add tests in `src`? example: https://github.com/open-policy-agent/gatekeeper-library/tree/master/src/general/allowedrepos

library/general/disallowedtags/samples/container-image-must-not-have-latest-tag/example_disallowed2.yaml
```diff
@@ -0,0 +1,12 @@
...
+metadata:
+  name: opa-disallowed-2
+spec:
+  containers:
+    - name: opa
+      image: openpolicyagent/opa:latest
+      args:
+        - "run"
+        - "--server"
+        - "--addr=localhost:8080"
```
sozercan, line 12: nit: newline

arapulido:
> Hi @arapulido 👋 thanks for the PR! Can you copy rego and add tests in `src`? example: https://github.com/open-policy-agent/gatekeeper-library/tree/master/src/general/allowedrepos

Hello @sozercan 👋  Nice to see you around :) Sure, I will work on those

sozercan (approved):
LGTM

PR #61, opened 2021-03-11
Handle false security context values correctly

### LINKED ISSUE #60: mustRunAsNonRoot evaluations will error out in some cases
This can be reproduced by adding this test to https://github.com/open-policy-agent/gatekeeper-library/blob/master/src/pod-security-policy/users/src_test.rego

```
test_wcs {
  input := {"review": review({"runAsNonRoot": true}, [ctr("cont1", runAsNonRoot(false))], null), "parameters": user_mustrunasnonroot }
  count(results) == 0
}
```

You should see: 
```
data.k8spspallowedusers.test_wcs: ERROR (1.205236ms)
  /opa-policies/src.rego:90: eval_conflict_error: functions must not produce multiple outputs for same inputs
```

The review object isn't anything too special, just a Pod with PodSecurityContext setting runAsNonRoot=true, but with a container overriding that.

The error is triggered by this line in the policy code: https://github.com/open-policy-agent/gatekeeper-library/blob/master/src/pod-security-policy/users/src.rego#L39

It will run the functions on both L84 and L90, and will obtain different results from them (false and true, respectively).

The root of the problem is probably that L91 is not a good enough guard. The function on L90 will trigger not only when "no container level field exists", it will also trigger when "the container level field exists but is set to false".

maxsmythe:
Fixes #60 

### DIFF
--- a/library/pod-security-policy/users/template.yaml
+++ b/library/pod-security-policy/users/template.yaml
@@ -161,7 +161,7 @@ spec:
 
         # If no container level exists, use pod level
         get_field_value(field, container, review) = out {
-          not get_seccontext_field(field, container)
+          not has_seccontext_field(field, container)
           review.kind.kind == "Pod"
           pod_value := get_seccontext_field(field, review.object.spec)
           out := pod_value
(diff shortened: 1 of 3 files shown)


sozercan (approved):
LGTM

PR #62, opened 2021-03-12
add mutation examples for PSP

sozercan:


library/pod-security-policy/users/mutation.yaml
```diff
@@ -0,0 +1,18 @@
...
+metadata:
+  name: k8spspusers
+spec:
+  applyTo:
+  - groups: [""]
+    kinds: ["Pod"]
+    versions: ["v1"]
+  location: "spec.containers[name:*].securityContext.runAsNonRoot"
+  parameters:
+    # pathTests:
```
sozercan, line 12: https://github.com/open-policy-agent/gatekeeper/issues/1176

maxsmythe:
What do we want to do about the fact that mutation is still in ~alpha?

library/pod-security-policy/capabilities/mutation.yaml
```diff
@@ -0,0 +1,13 @@
...
+  name: k8spspcapabilities
+spec:
+  applyTo:
+  - groups: [""]
+    kinds: ["Pod"]
+    versions: ["v1"]
+  location: "spec.containers[name:*].securityContext.capabilities.add"
+  parameters:
+    assign:
+      value: ["NEW_CAPABILITY"] # default add capability
```
maxsmythe, line 13: is `NEW_CAPABILITY` an actual thing? If these are toy examples, we should be careful where we put them.

library/pod-security-policy/users/mutation-mustRunAsNonRoot.yaml
```diff
@@ -0,0 +1,18 @@
...
+  name: k8spsprunasnonroot
+spec:
+  applyTo:
+  - groups: [""]
+    kinds: ["Pod"]
+    versions: ["v1"]
+  location: "spec.containers[name:*].securityContext.runAsNonRoot"
+  parameters:
+    pathTests:
+    # - subPath: "spec.containers[name:*].securityContext.runAsUser"
```
maxsmythe, line 13: Is there any harm in setting runAsNonRoot even if runAsUser is also set?

maxsmythe (commented):
Couple of questions on the specifics of these mutations.

Also, I'm not sure how I feel about the placement of these in the library. On one hand, it's good to have an example. On the other hand, we are putting actual mutations (which have production impact) next to templates (which need constraints to be created before they do a thing).

Maybe the right thing to do here is to put everything in the `samples` folder?

sozercan:
@maxsmythe I am targeting `mutation` branch here for now (I think we discussed this in the community call at some point), so they are not next to the templates. I just kept the same layout (`library/pod-security-policy/`).

library/pod-security-policy/capabilities/mutation.yaml
sozercan: Yes, it's just an example. I was hesitant on adding an actual capability, just in case someone adds it without checking.

library/pod-security-policy/users/mutation-mustRunAsNonRoot.yaml
sozercan: Kubelet gives `CreateContainerConfigError` if they are conflicting: (runAsUser=0 and runAsNonRoot=true)

```
  Warning  Failed     17s                kubelet            Error: container's runAsUser breaks non-root policy (pod: "nginx-run-as-root_default(350bd4bc-10c2-4299-912f-a9651dfbadab)", container: nginx)
```

maxsmythe: I'm thinking this is a job for the match criteria.

If you have a constraint that is enforcing run-as-non-root, you'll want this defaulting set, at which point you'd like the request to be rejected anyway (though it would be nice to know which would win, the violation message or the conflict message, as the former is more user-friendly).

If users want to allow run as root, then they probably have no opinion as to whether/how `runAsNonRoot` is set.

If they want to force run as root to be set, they would have two mutations so they could default both values.

This does point to a general syntactic weakness when dealing with [union](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#unions) type objects, however.

maxsmythe: Another option would be to default the `securityContext` object in its entirety

maxsmythe:
That mutation branch strategy sounds familiar to me. I think I'd still put these in the `examples` subfolder, per an above comment

library/pod-security-policy/users/mutation-mustRunAsNonRoot.yaml
sozercan: Wouldn't setting `securityContext` be same issue because of non-prefix path tests?

What is the next step here? Should we remove the commented out part and leave it up to validation to check for `runAsUser` to be non-zero?

maxsmythe: The two different paths share a common root at `spec.containers[name:*].securityContext`

Yeah, we can delete the comment, or rejigger the example to avoid the potential issue.

sozercan:
```
  location: "spec.containers[name:*].securityContext"
  parameters:
    pathTests:
    - subPath: "spec.containers[name:*].securityContext.runAsUser"
      condition: MustNotExist
    - subPath: "spec.containers[name:*].securityContext.runAsNonRoot"
      condition: MustNotExist
    assign:
      value:
        runAsNonRoot: true
```

like this? this also gives non-prefix error

I removed the path test for now.

maxsmythe: More like this:

```
  location: "spec.containers[name:*].securityContext"
  parameters:
    assign:
      value:
        runAsNonRoot: true
```

The path test is unnecessary as the glob expansion won't take effect if there are no containers, and we want to set all security contexts to "run as non-root"

sozercan: > we want to set all security contexts to "run as non-root"

If we want similar functionality with PSP, it's set all security contexts unless it's defined explicitly.

Example scenario:
- User has a certain workload (example, uses port 80) that can run as root only, so user sets it accordingly
- Gatekeeper won't mutate since path tests are set (otherwise, workload won't run at all)
- User adds exclusion for that workload so it's not denied, but rest of the workloads are still part of mutation and validation

maxsmythe: Then you could add a pathTest like this:

```
  location: "spec.containers[name:*].securityContext"
  parameters:
    pathTests:
    - subPath: "spec.containers[name:*].securityContext"
      condition: MustNotExist
    assign:
      value:
        runAsNonRoot: true
```

sozercan:
```
    - subPath: "spec.containers[name:*].securityContext"
      condition: MustNotExist
```

I think this path test is too broad since securityContext can have other things like `capabilities` that's not related to users.

maxsmythe: IMO that's up to the admin... I don't think this can be one size fits all.

maxsmythe: Also, because it's saying `MustNotExist`, it won't override any security context provided by the user

sozercan:
@maxsmythe @ritazh do we need any other changes in this PR?

maxsmythe:
I think the pathtests thing is the only open discussion.

ritazh:
> What do we want to do about the fact that mutation is still in ~alpha?

I prefer we merge these to main branch and create a mutation folder under PSP.

PR #65, opened 2021-03-17
Adding missing kustomization resources

vfarcic:
Adding missing kustomization resources


maxsmythe (approved):
LGTM, thanks for this!

PR #70, opened 2021-04-06
Adding some library policies for Windows workloads

### LINKED ISSUE #4: RunAsUserName Policy Constraint for windows pods
**Describe the solution you'd like**
A policy that blocks `ContainerAdministrator` from being set on the WindowsOptions podspec.  It is generally a good idea to run your containers with `ContainerUser` for windows pods.  The users are not shared between the Container and host but the `ContainerAdministrator` does have additional privileges with in the container.  In the PR for https://github.com/kubernetes/kubernetes/pull/92355  an agree was made block `ContainerAdministrator` if `RunAsNonRoot` was specified.

The pod sec looks like ([can also be set per container](https://kubernetes.io/docs/tasks/configure-pod-container/configure-runasusername/#set-the-username-for-a-container)):

```
spec:
  securityContext:
    windowsOptions:
      runAsUserName: "ContainerUser"
```

There are also username limitations to be aware of: https://kubernetes.io/docs/tasks/configure-pod-container/configure-runasusername/#windows-username-limitations

**Anything else you would like to add:**
[Miscellaneous information that will assist in solving the issue.]

The following has more information on Windows Security Policies which are currently being defined:
kubernetes/kubernetes#64801 (comment)

**Environment:**

- Gatekeeper version:
- Kubernetes version: (use `kubectl version`):

marosset:
Fixes #4 

This PR adds the following new policies
- A policy to enforce that all pods have specified node selectors set
- A policy to disallow pods that use specified RunAsUserNames in their securityContext options
- A policy that enforces best practices around resource requests/limits for Windows pods 
- Add docs for recommendations on using OS specific policies in a cluster

These library policies will use `kubernetes.io/os` node selectors to determine if a pod is a Windows pod.
Once PSP v2 settles on a way to identify the OS for a pod I can update these library polices to use the same mechanism in a future PR.

Documentation will be added to describe usages patterns following https://github.com/open-policy-agent/gatekeeper-library/pull/70#issuecomment-824460250

Signed-off-by: Mark Rossetti <marosset@microsoft.com>

### DIFF
--- a/OS-specific-policies.md
+++ b/OS-specific-policies.md
@@ -1,0 +1,47 @@
+# OS Specific Policies
+
+Kubernetes allows for pods/containers targeting different operating systems to be scheduled into the same cluster and different operating systems and can behave differently depending on what OS the pods/containers are targeting.
+
+This guide will help you configure policies that are meant only to specific operating systems.
+
+## Usage
+
+### Determine how your cluster will identify target OS
+
+Unfortunately there is no definitive way identify which OS a pod is targeted for.
+If you would like to enforce OS-specific policies in your cluster you should first determine how your cluster will identify the target OS for pods.
+
+Policies in this repository will use [nodeSelectors](https://v1-18.docs.kubernetes.io/docs/concepts/scheduling-eviction/assign-pod-node/#nodeselector) against the well-known [kubernetes.io\os](https://kubernetes.io/docs/reference/labels-annotations-taints/#kubernetes-io-os) label for pods. This is the simplest way to identifying the target OS for a pod but other mechanisms like [Runtime Classes](https://kubernetes.io/docs/concepts/containers/runtime-class/) exist as well and will follow the same procedure for enabling policy enforcement.
+
+### Add a policy requiring all pods to identify an OS
+
+After you have selected how you would like your cluster to identify the target OS for pods create a policy to ensure **every** pod admitted into your cluster specifies an OS.
+
+Example:
+
+```bash
+cd library/general/requirednodeselectors
+kubectl apply -f template.yaml
+kubectl apply -f samples/require-os-node-selector/constraint.yaml
+```
+
+### Add OS specific policies
+
+1. Include a check for your targeted OS in the rego for your policy. This check should skip policy enforcement if the OS does not match and enforce policy enforcement if the OS does match.
+
+    Example: The function `isWindowsPod` in the [windows-runasusername](./library/pod-security-policy/windows-runasusername/template.yaml) template.
+
+1. Add OS specific policies to your cluster.
+
+    Example:
+
+    ```bash
+    cd library/general/windows-container-resources
+    kubectl apply -f template.yaml
+    kubectl apply -f samples/constraint.yaml
+    ```
+
+Some examples of OS specific policies include:
+
+- [windows-container-resources](./library/general/windows-container-resources)
+- [windows-runasusername](./library/pod-security-policy/windows-runasusername)
--- a/README.md
+++ b/README.md
@@ -15,6 +15,9 @@ kubectl apply -f samples/ingress-https-only/constraint.yaml
 kubectl apply -f library/general/httpsonly/sync.yaml # optional: when GK is running with OPA cache
 ```
 
+Note: Some policies should only be enforced for pods targeting a specific OS.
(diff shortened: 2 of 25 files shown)


jsturtevant:
> Note: if there is not a nodeSelector for kubernetes.io/os : windows this policy will never trigger a violation.

Would there be a way to create a different policy such as "if there are Windows nodes in the cluster then you must specify node selector of type `kubernetes.io/os`"?    It could potentially remove the "I deployed my workload on the wrong os" and make these hard to bypass.

maxsmythe:
We talked about this some in the meeting today. There are a lot of problems with trying to make pod constraints only apply to pods running on certain OS's out of the box:

- There are apparently multiple potential ways to limit a pod to a particular OS
- There is no built-in requirement for users to make any of these specifications
- Unless you force users to adopt a specific policy, baking in any solution risks leaving some users out in the cold

We didn't get a chance to come to a final conclusion about what the best path forward would be. My proposed idea:

- Constraint templates do not try to solve this problem, let users limit the scope of constraints via namespace/ namespace selectors / label selectors
- Users can create a second constraint template that understands how they indicate "windows" (or whatever OS) for their particular cluster and the requirements around it (e.g. it must always be specified, if not specified assume X, etc.).
- This second constraint template can be used to limit the application of the "os" pod label (or use of special namespace, or however users divide their workloads) to only those pods that meet the appropriate criteria

ritazh:
This PR is pending "A standardized identifier for Windows pods" to ensure OS-specific policies are enforcing workloads on those OSes only.

marosset:
I didn't know there was a community meeting today and that this was going to be discussed.
I'll join next week.

jsturtevant:
Discussed this at meeting on 4/21.  To summarize (essentially the same as @maxsmythe's comment above):

Ideally we would have something on the pod specs that would identify OS.  In the current versions of Kubernetes there is no standard solution, so this could be different depending on the cluster and configuration.  A few examples could be node selector, a label, runtime classes, namespaces, etc.

Since we don't know the cluster configuration, to get this working properly today, we need to use a two Constraint approach:

1.  The first constraint would ensure that all pods have an OS identifier regardless of OS. This could be runtimeclass, nodeselector, label, etc. This ensures that a user couldn't by-pass a policy by leaving off the identifier (ie if windows pod was scheduled without a runtime class how would the policy for windows know to trigger evaluation).  This will be specific to the way a cluster administrator sets up the node. We can give an example but it will be up to the cluster admin to make sure this applied.
2. The second constraint would be the one for a given policy, in this case the RunAsUserName, that would check for the Windows (or in other policies it might be Linux)

In the future if Kubernetes provides a value on the podspec for OS type we could use that as the first constraint.  This is being discussed a future option in the PSP vNext KEP:  https://github.com/kubernetes/enhancements/pull/2582/files#r617105381

marosset:
I added a few more policies to this PR and will work on adding documentation next

marosset:
Sorry for putting this off for so long!

marosset:
I'm looking for some feedback on the approach I'm using to determine the target OS for pods.
The two most common ways are to use nodeSelectors against `kubernetes.io\os` or runtime classes (which usually also have nodeSelector for os).

For now I think it would be best to have library policies use nodeSelectors directly.
Our data/surveys show that an overwhelming majority of Windows nodes joined to Kubernetes clusters are using Docker EE/dockershim.
There is a default runtime class handler that can be used with dockershim [but the name is simply `docker`](https://kubernetes.io/docs/concepts/containers/runtime-class/#dockershim) and there is no way to distinguish the target OS from that.

Once dockershim is completely removed for Kubernetes (scheduled for 1.21 release) I think it would make sense to update these library policies to use runtime classes.

I feel that this approach is best for the current userbase but am open to any and all feedback here.

paprickar:
Hey everyone,

I am in need for these policies for all kind of deployments (not only pods).

Therefore I created following PR against "marosset:windows-runasusername" branch (<https://github.com/marosset/gatekeeper-library/pull/1>).

Would be great to get Feedback / if I am missing something here.

FYI:
I am currently testing these changes against Daemonsets, StatefulSet etc. (Deployments are covered). If wanted I then can contribute example disallow yaml definitions as well.

Best
Patrick

PR #72, opened 2021-04-16
add mutation pod security policies

sozercan:

Re-opening since I can't change target branch of #62 

As discussed on the community call, policies are moved to `library/experimental/mutation/pod-security-policy/`

### DIFF
--- a/library/experimental/mutation/pod-security-policy/allow-privilege-escalation/samples/mutation.yaml
+++ b/library/experimental/mutation/pod-security-policy/allow-privilege-escalation/samples/mutation.yaml
@@ -1,0 +1,16 @@
+apiVersion: mutations.gatekeeper.sh/v1alpha1
+kind: Assign
+metadata:
+  name: k8spspdefaultallowprivilegeescalation
+spec:
+  applyTo:
+  - groups: [""]
+    kinds: ["Pod"]
+    versions: ["v1"]
+  location: "spec.containers[name:*].securityContext.allowPrivilegeEscalation"
+  parameters:
+    pathTests:
+    - subPath: "spec.containers[name:*].securityContext.allowPrivilegeEscalation"
+      condition: MustNotExist
+    assign:
+      value: false
--- a/library/experimental/mutation/pod-security-policy/allow-privilege-escalation/samples/default-allow-privilege-escalation/example.yaml
+++ b/library/experimental/mutation/pod-security-policy/allow-privilege-escalation/samples/default-allow-privilege-escalation/example.yaml
@@ -1,0 +1,10 @@
+apiVersion: v1
+kind: Pod
+metadata:
+  name: nginx-default-privilege-escalation
+  labels:
+    app: nginx-default-privilege-escalation
+spec:
+  containers:
+  - name: nginx
+    image: nginx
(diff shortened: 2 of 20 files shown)


maxsmythe (approved):
LGTM

library/experimental/mutation/pod-security-policy/allow-privilege-escalation/samples/mutation.yaml
```diff
@@ -0,0 +1,16 @@
...
+spec:
+  applyTo:
+  - groups: [""]
+    kinds: ["Pod"]
+    versions: ["v1"]
+  location: "spec.containers[name:*].securityContext.allowPrivilegeEscalation"
+  parameters:
+    pathTests:
+    - subPath: "spec.containers[name:*].securityContext.allowPrivilegeEscalation"
+      condition: MustNotExist
```
ritazh, line 14: > MustNotExist - the path must not exist or do not mutate

What if `securityContext.allowPrivilegeEscalation: true` is set already?

sozercan: I believe in PSP if it's explicitly defined it doesn't mutate.
https://github.com/kubernetes/kubernetes/blob/master/pkg/security/podsecuritypolicy/provider.go#L185-L194

ritazh (approved):

nwesoccer:
Is there a way to have mutation location reference multiple things?  For example setting securityContext.allowPrivilegeEscalation for both all containers and all initContainers.  In fact, most of these securityContext settings can be applied to both containers and initContainers.  The validation web-hook and [constraint templates](https://github.com/open-policy-agent/gatekeeper-library/blob/master/library/pod-security-policy/allow-privilege-escalation/template.yaml#L33) check both.  I know we could duplicate all of these for both, but that seems like a lot of duplication.

Thanks

maxsmythe:
This comment seems similar to https://github.com/open-policy-agent/gatekeeper/issues/1175 in spirit (though they are inspired by different things -- templates vs. contaners/initContainers).

Also interesting is that a solution for one doesn't address a solution for the other (and may even make it harder).

Are there any resources other than Pods that have this same paradigm of two different paths, same struct?

We could accept a list of paths and move `pathTests` to be next to their associated parent path.

I'm wondering if that over-fits on the pod use case (as important as it is).

It also muddies the waters a bit... if I have multiple kinds/versions, which sets of paths are valid for which of the kinds/versions?

I think these issues speak to the need for some higher-order mutation object that can generate mutators rather than a need to make these very granular mutators more general.

I potentially see the need for:

* Multiple mutators ingested atomically (think adding both a volume and a volume mount at the same time... we don't want to add the mount but not the volume)
* Mutator "expansion" -- turning a template for a mutator into one that can modify both containers and initContainers or one that modifies different paths on different kinds (but sets them to the same value).

@shomron curious if you have any thoughts?

nwesoccer:
Seems the seccomp supports `/pod` but I'm having trouble finding documentation that mentions the apparmor annotations supporting it (metadata.annotations."apparmor.security.beta.kubernetes.io/pod).  Also, the apparmor constraint does not consider it valid either.

Issue #64, opened 2021-03-17
Check timeoutSeconds for readinessProbe and livenessProbe


ritazh:
This [Kubernetes PR](https://github.com/kubernetes/kubernetes/pull/94115) enabled `ExecProbeTimeout` feature flag, which ensures kubelet will respect exec probe timeouts. If `timeoutSeconds` is not specified, the timeout will default to 1 second. A policy to check if `timeoutSeconds` is set for readinessProbe and livenessProbe will ensure proper timeout is set to avoid any breakage.

sozercan:
> Note:
Before Kubernetes 1.20, the field timeoutSeconds was not respected for exec probes: probes continued running indefinitely, even past their configured deadline, until a result was returned.

> This defect was corrected in Kubernetes v1.20. You may have been relying on the previous behavior, even without realizing it, as the default timeout is 1 second. As a cluster administrator, you can disable the feature gate ExecProbeTimeout (set it to false) on each kubelet to restore the behavior from older versions, then remove that override once all the exec probes in the cluster have a timeoutSeconds value set.
If you have pods that are impacted from the default 1 second timeout, you should update their probe timeout so that you're ready for the eventual removal of that feature gate.

https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/#configure-probes

reetasingh:
@ritazh @sozercan I have created a PR for this https://github.com/open-policy-agent/gatekeeper-library/pull/74. 
but the integration tests are failing because pod is getting created with default timeoutSeconds.  unit tests have all passed. how do I test this behaviour and make sure integration tests pass?

PR #74, opened 2021-04-17
Add policy to check timeoutSeconds field is present in probes

reetasingh:
Adress https://github.com/open-policy-agent/gatekeeper-library/issues/64


maxsmythe:
This LGTM, not sure why the e2e tests are failing

reetasingh:
> This LGTM, not sure why the e2e tests are failing

@maxsmythe thanks for reviewing this.
I think its because by default kubernetes sets default timeoutSeconds=1 . for my example_disallowed.yaml even though I dont give timeoutSeconds field. kubernetes it setting default value and hence it is getting allowed

@ritazh @sozercan any ideas on how to address this so that the e2e tests pass?

ritazh:
> I think its because by default kubernetes sets default timeoutSeconds=1 

That's right. Perhaps this policy can instead take in a parameter for timeoutSeconds and validate the request values against the parameter value.

maxsmythe:
@reetasingh still interested in this PR?

PR #81, opened 2021-05-18
Add anchoring to allowedrepos example

### LINKED ISSUE #80: using `startwith` in allowedrepos policy may allow for bypass
The [Allowed Repos policy](https://github.com/open-policy-agent/gatekeeper-library/blob/master/library/general/allowedrepos/template.yaml) checks that containers are from approved locations.

It uses startswith as a function to conduct [this check](https://github.com/open-policy-agent/gatekeeper-library/blob/master/library/general/allowedrepos/template.yaml#L28).

Taking this approach could allow for bypass as an attacker could just start their image name with one of the approved strings.

As shown in [this example ](https://play.openpolicyagent.org/p/iHgFVrQu08) where the allowed string is "openpolicyagent" , "openpolicyagentnotreally" will be allowed.

Using an approach which matches the whole account name or other string, would help mitigate this kind of risk.

maxsmythe:
Fixes #80 

### DIFF
--- a/library/general/allowedrepos/samples/repo-must-be-openpolicyagent/constraint.yaml
+++ b/library/general/allowedrepos/samples/repo-must-be-openpolicyagent/constraint.yaml
@@ -11,4 +11,4 @@ spec:
       - "default"
   parameters:
     repos:
-      - "openpolicyagent"
+      - "openpolicyagent/"


ritazh (approved):

npapageorgopoulos:
how can I add multiple repos, and what if one of these is docker hub which doesn't need repository uri... for instance I can provide only image name like the disallowed example

PR #91, opened 2021-07-09
fix: fix privilege-escalation check to handle securityContexts with missing allowPrivilegeEscalation element

scoof:
This fixes the logic for when securityContext is present, but doesn't contain allowPrivilegeEscalation.

It seems like some tests were actually testing these conditions, but their assertions were wrong, so I have fixed those as well.

### DIFF
--- a/src/pod-security-policy/allow-privilege-escalation/src.rego
+++ b/src/pod-security-policy/allow-privilege-escalation/src.rego
@@ -7,10 +7,9 @@ violation[{"msg": msg, "details": {}}] {
 }
 
 input_allow_privilege_escalation(c) {
-    not has_field(c, "securityContext")
-}
-input_allow_privilege_escalation(c) {
-    not c.securityContext.allowPrivilegeEscalation == false
+    has_field(c, "securityContext")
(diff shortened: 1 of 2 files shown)


scoof:
CLA signed

maxsmythe:
Thanks for the PR!

From the PSP docs it looks like `allowPrivilegeEscalation` is assumed to be true if undefined:


https://kubernetes.io/docs/concepts/policy/pod-security-policy/#privilege-escalation
```
DefaultAllowPrivilegeEscalation - Sets the default for the allowPrivilegeEscalation option. The default behavior without this is to allow privilege escalation so as to not break setuid binaries. If that behavior is not desired, this field can be used to default to disallow, while still permitting pods to request allowPrivilegeEscalation explicitly.
```

Is that not the case?

It looks like this PR changes things to assume that `allowedPrivilegeEscalation` is `false` if undefined.

We should make sure our expectations match the true behavior.

scoof:
Well, that was embarressing. LMGBTY

scoof:
(I was confused by testing on a cluster with DefaultAllowPrivilegeEscalation=false)

Issue #96, opened 2021-08-01
Incorrect handling of "/" in k8spsphostfilesystem policy


philsphicas:
When considering whether filesystem paths match, the special case of the root path "/" is not handled correctly.

For example, the following test fails.

```rego
test_path_matches {
    path_matches("/", "/foo")
}
```

PR #97, opened 2021-08-01
Fix k8spsphostfilesystem path matching

philsphicas:
This change updates the path matching logic to correctly handle "/",
which should match all paths.

Fixes https://github.com/open-policy-agent/gatekeeper-library/issues/96

### DIFF
--- a/library/pod-security-policy/host-filesystem/template.yaml
+++ b/library/pod-security-policy/host-filesystem/template.yaml
@@ -78,10 +78,18 @@ spec:
         # This allows "/foo", "/foo/", "/foo/bar" etc., but
         # disallows "/fool", "/etc/foo" etc.
         path_matches(prefix, path) {
-            a := split(trim(prefix, "/"), "/")
-            b := split(trim(path, "/"), "/")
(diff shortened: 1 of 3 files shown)


philsphicas:
I signed it

maxsmythe (changes requested):
Thanks for the PR!

The code looks good, can you make sure the changes are also included in:

https://github.com/open-policy-agent/gatekeeper-library/blob/master/library/pod-security-policy/host-filesystem/template.yaml

?

Unfortunately, we haven't yet automated the pipeline for doing this.

philsphicas:
> The code looks good, can you make sure the changes are also included in:
> 
> https://github.com/open-policy-agent/gatekeeper-library/blob/master/library/pod-security-policy/host-filesystem/template.yaml
> 

Done.

Thanks for reviewing, sorry I missed that.

maxsmythe (approved):
Absolutely. No worries on missing that, things should be more automated than they currently are, thank you for the PR!

ritazh (approved):

PR #99, opened 2021-08-03
add extra control for .securityContext field to privileged container rule

developer-guy:
Signed-off-by: Batuhan Apaydın <batuhan.apaydin@trendyol.com>

### DIFF
--- a/library/pod-security-policy/privileged-containers/template.yaml
+++ b/library/pod-security-policy/privileged-containers/template.yaml
@@ -16,8 +16,16 @@ spec:
 
         violation[{"msg": msg, "details": {}}] {
             c := input_containers[_]
-            c.securityContext.privileged
+            input_allow_privileged_container(c)
             msg := sprintf("Privileged container is not allowed: %v, securityContext: %v", [c.name, c.securityContext])
+        }
+
+        input_allow_privileged_container(c) {
+            not has_field(c, "securityContext")
+        }
+
+        input_allow_privileged_container(c) {
+            not c.securityContext.privileged == false
         }
 
         input_containers[c] {
@@ -26,4 +34,9 @@ spec:
 
         input_containers[c] {
             c := input.review.object.spec.initContainers[_]
+        }
+
+        # has_field returns whether an object has a field
+        has_field(object, field) = true {
+            object[field]
         }
--- a/src/pod-security-policy/privileged-containers/src_test.rego
+++ b/src/pod-security-policy/privileged-containers/src_test.rego
@@ -13,7 +13,7 @@ test_input_container_privileged_not_allowed {
 test_input_container_many_not_privileged_allowed {
     input := { "review": input_review_many}
     results := violation with input as input
-    count(results) == 0
+    count(results) > 0
 }
 test_input_container_many_mixed_privileged_not_allowed {
     input := { "review": input_review_many_mixed}
--- a/src/pod-security-policy/privileged-containers/src.rego
+++ b/src/pod-security-policy/privileged-containers/src.rego
@@ -1,9 +1,18 @@
 package k8spspprivileged
 
+
 violation[{"msg": msg, "details": {}}] {
     c := input_containers[_]
-    c.securityContext.privileged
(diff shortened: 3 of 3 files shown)


library/pod-security-policy/privileged-containers/template.yaml
```diff
@@ -16,14 +16,27 @@ spec:
 
         violation[{"msg": msg, "details": {}}] {
             c := input_containers[_]
-            c.securityContext.privileged
+            input_allow_privileged_container(c)
             msg := sprintf("Privileged container is not allowed: %v, securityContext: %v", [c.name, c.securityContext])
```
maxsmythe, line 20: if `c.securityContext` is undefined, this line will not match, which would render the changes you made moot.

maxsmythe (changes requested):
Thanks for the PR!

A couple of questions:

- It looks like the default value if `securityContext` is not supplied is `false`, can you elaborate on why we'd wan't to reject if a security context is undefined?
- We should add unit tests to make sure your change is operating as expected.

developer-guy:
> Thanks for the PR!
> 
> A couple of questions:
> 
> * It looks like the default value if `securityContext` is not supplied is `false`, can you elaborate on why we'd wan't to reject if a security context is undefined?
> * We should add unit tests to make sure your change is operating as expected.

Hello @maxsmythe, thanks for the quick response. I did it on purpose because I saw that other rules related to under `.securityContext` fields all have this control such as [read-only-file-system](https://github.com/open-policy-agent/gatekeeper-library/tree/master/library/pod-security-policy/read-only-root-filesystem), [allow-privilege-escalation](https://github.com/open-policy-agent/gatekeeper-library/tree/master/library/pod-security-policy/allow-privilege-escalation), so I added this control for privileged-container rule too.

library/pod-security-policy/privileged-containers/template.yaml
developer-guy: thank you so much for warning me about this @maxsmythe , I added unit tests and fixed the message to avoid this problem.

developer-guy:
Any updates @maxsmythe 🙏🙋🏻‍♂️

src/pod-security-policy/privileged-containers/src_test.rego
```diff
@@ -13,7 +13,7 @@ test_input_container_privileged_not_allowed {
 test_input_container_many_not_privileged_allowed {
     input := { "review": input_review_many}
     results := violation with input as input
-    count(results) == 0
+    count(results) > 0
```
ritazh, line 16: why is this test updated?
```
"spec": {
            "containers": input_containers_many,
            "initContainers": input_containers_one
```
both have `c.securityContext.privileged == false`

library/pod-security-policy/privileged-containers/template.yaml
```diff
@@ -16,14 +16,27 @@ spec:
...
         }
 
+        input_allow_privileged_container(c) {
+            not has_field(c, "securityContext")
```
ritazh, line 24: if `securityContext` is not provided, k8s would default it to `false` which should not trigger this violation.

maxsmythe:
Sorry I dropped the ball on replying. Rita's point is well-taken. It may be that `read-only-filesystem` does not have a secure default when `securityContext` is undefined, but `allowPrivilegeEscalation` does, which would make failing this constraint template on an empty `securityContext` inaccurate.

library/pod-security-policy/privileged-containers/template.yaml
developer-guy: yeah, I have just checked this and you are right! thank you so much @ritazh

developer-guy:
> Sorry I dropped the ball on replying. Rita's point is well-taken. It may be that `read-only-filesystem` does not have a secure default when `securityContext` is undefined, but `allowPrivilegeEscalation` does, which would make failing this constraint template on an empty `securityContext` inaccurate.

ok now I got it, I'm closing this PR, thank you so much @maxsmythe

PR #106, opened 2021-09-12
Added policy to disallow unauthenticated access to cluster resources

michaelkotelnikov:
The [NSA / CISA Kubernetes Hardening Guidance (August 2021)](https://media.defense.gov/2021/Aug/03/2002820425/-1/-1/1/CTR_KUBERNETES%20HARDENING%20GUIDANCE.PDF) [page 19] states that -

"
In Kubernetes 1.6 and newer, anonymous requests are enabled by default. When RBAC is enabled, anonymous requests require explicit authorization of the system:anonymous user or system:unauthenticated group. Anonymous requests should be disabled by passing the --anonymous-auth=false option to the API server. Leaving anonymous requests enabled could allow a cyber actor to access cluster resources without authentication.
"

The recommendation states that anonymous authentication should be disabled by default. Yet, many environments leave `--anonymous-auth` enabled for various reasons. The purpose of this policy is to control and reduce the usage of anonymous requests in Kubernetes environments.

Roles / ClusterRoles that the organization considers "safe" for anonymous requests will be placed in `constraint.yaml`, while others will be blocked on creation.

***Scenario 1 - Gatekeeper policy is not deployed*** - 

```
# as user-1, local namespace admin -

$ cat disallowed-binding.yaml
apiVersion: rbac.authorization.k8s.io/v1
kind: RoleBinding
metadata:
  name: dangerous-rolebinding
  namespace: application-1
roleRef:
  apiGroup: rbac.authorization.k8s.io
  kind: ClusterRole
  name: admin
subjects:
- apiGroup: rbac.authorization.k8s.io
  kind: Group
  name: system:unauthenticated

$ kubectl apply -f /tmp/disallowed-binding.yaml 
rolebinding.rbac.authorization.k8s.io/dangerous-rolebinding created

$ kubectl get pods --as="system:anonymous" -n application-1
NAME                         READY   STATUS    RESTARTS   AGE
webserver-7cccc68cc6-dmkt5   1/1     Running   0          8m28s
```

***Scenario 2 - Gatekeeper policy is deployed*** -

```
# as user-1, local namespace admin -

kubectl apply -f disallowed-binding.yaml 
Error from server ([denied by no-anonymous] Unauthenticated user reference is not allowed in RoleBinding dangerous-rolebinding ): error when creating "/tmp/disallowed-binding.yaml": admission webhook "validation.gatekeeper.sh" denied the request: [denied by no-anonymous] Unauthenticated user reference is not allowed in RoleBinding dangerous-rolebinding
```

Feel free to share feedback or edit the contents of the PR :)


michaelkotelnikov:
For CLA authorization - I signed it

library/general/disallowanonymous/template.yaml
```diff
@@ -0,0 +1,33 @@
+apiVersion: templates.gatekeeper.sh/v1beta1
+kind: ConstraintTemplate
+metadata:
+  name: k8sdisallowanonymous
+  annotations:
+    description: Disallows associating ClusterRole and Role resources to the system:anonymous user and system:unauthenticated group.
```
maxsmythe, line 6: Should we make this more generic by adding parameters that allow constraint authors to specify which users/groups should be disallowed?

maxsmythe (changes requested):
Thanks for the PR! Sorry for the slow turnaround :/

One thought about how we could make this constraint template more generic.

library/general/disallowanonymous/template.yaml
michaelkotelnikov: Thanks for reviewing @maxsmythe :)

I like the idea, I thought about way to implement the solution without over complicating the process for the constraint author. Came up with something like - 

```
kind: K8sDisallowAnonymous
metadata:
  name: no-anonymous
spec:
  match:
    kinds:
      - apiGroups: ["rbac.authorization.k8s.io"]
        kinds: ["ClusterRoleBinding"]
      - apiGroups: ["rbac.authorization.k8s.io"]
        kinds: ["RoleBinding"]
  parameters:
    roleMapping:
      - Users:
          - group-1
          - user-1 
        allowedRoles:
          - cluster-role-1
          - cluster-role-2
      - Users:
          - group-2
          - user-2
        allowedRoles:
          - cluster-role-3
          - cluster-role-4
```

I'm not sure if I'm confident with this though. This kind of ignores `ServiceAccounts` and the fact that they're namespace scoped. We might want the constraint to affect `ServiceAccounts` in specific namespaces (since their names may be the same across namespaces). On the other hand, taking the ServiceAccount Namespace into account will create a complex constraint that might be confusing for a constraint author.

Maybe the constraint should not be generic? After all, the original idea is to focus on an implementation of a specific "security guideline" that helps handling anonymous authentications (represented using the anonymous user and unauthenticated group).

maxsmythe: Yeah, there are a lot of edge cases depending on how generic we make this. Might be better to leave well enough alone for now.

library/general/disallowanonymous/template.yaml
```diff
@@ -0,0 +1,33 @@
...
+spec:
+  crd:
+    spec:
+      names:
+        kind: K8sDisallowAnonymous
```
maxsmythe, line 11: It looks like this CT accepts parameters. We'll need a schema for that. Here's an example schema:

https://github.com/open-policy-agent/gatekeeper-library/blob/de1818ffdb03b2005dc8ef4beb23bfbadb85a63f/library/general/allowedrepos/template.yaml#L13-L22

michaelkotelnikov: @maxsmythe Added a schema to CT.

maxsmythe (commented):
LGTM after one more change:

Can we add a suite.yaml to this so it can be integrated with the new testing framework?

Example suite.yaml:

https://github.com/open-policy-agent/gatekeeper-library/blob/master/library/general/allowedrepos/suite.yaml

maxsmythe:
Also, docs for how to run `gator test`, if you're interested: https://open-policy-agent.github.io/gatekeeper/website/docs/gator

michaelkotelnikov:
@maxsmythe Added [suite.yaml](https://github.com/open-policy-agent/gatekeeper-library/blob/6a99ecdd1d23e25d44a8883951324a341afd3c2b/library/general/disallowanonymous/suite.yaml) and changed the directory structure at [samples](https://github.com/open-policy-agent/gatekeeper-library/tree/6a99ecdd1d23e25d44a8883951324a341afd3c2b/library/general/disallowanonymous/samples)

maxsmythe (approved):
LGTM!

@ritazh @sozercan @shomron LGTY?

src/general/disallowanonymous/src_test.rego
```diff
@@ -0,0 +1,248 @@
...
+    count(results) == 1
+}
+
+test_allowed_role_anonymous_user_clusterrolebinding {
+    input := {"review": clusterrolebinding([{"name": "system:anonymous", "kind": "User"}], "role-2"), "parameters": {"allowedRoles": ["role-2"]}}
+    results := violation with input as input
+    count(results) == 0
+}
+
+test_multiple_subjects_anonymous_user_clusterrolebinding {
```
ritazh, line 27: Do we have a test with just `system:authenticated`?
```rego
input := {"review": clusterrolebinding([{"name": "system:authenticated", "kind": "Group"}], "role-2"), "parameters": {"allowedRoles": ["role-1"]}}
count(results) == 0
```

michaelkotelnikov: Thanks @ritazh. I added the test.

PR #109, opened 2021-09-14
First stab at field descriptions for the PSP templates

julianKatz:


library/pod-security-policy/apparmor/template.yaml
```diff
@@ -15,6 +15,7 @@ spec:
           type: object
           properties:
             allowedProfiles:
+              description: "An array of AppArmor profiles. Examples: runtime/default, unconfined."
```
maxsmythe, line 18: backtick on values (`runtime/default`)

library/pod-security-policy/apparmor/template.yaml
```diff
@@ -15,6 +15,7 @@ spec:
...
               type: array
               items:
                 type: string
```
maxsmythe, line 21: Should we have a docstring for the individual item? Is that an option?

library/pod-security-policy/capabilities/template.yaml
```diff
@@ -16,10 +16,12 @@ spec:
           properties:
             allowedCapabilities:
               type: array
+              description: "An allow-list of Linux capabilities."
               items:
                 type: string
             requiredDropCapabilities:
               type: array
+              description: "A disallow-list of Linux capabilities."
```
maxsmythe, line 24: This may not be how this works. IIRC pods allow you to specify capabilities that are explicitly dropped, but then they can be re-added.

This supports someone saying "drop all capabilities, but re-add the NET capability"

library/pod-security-policy/flexvolume-drivers/template.yaml
```diff
@@ -16,10 +16,12 @@ spec:
           properties:
             allowedFlexVolumes:
               type: array
+              description: "An array of AllowedFlexVolume objects with the single field: 'driver'."
```
maxsmythe, line 19: No need to list the fields of objects, that's done by the schema.

library/pod-security-policy/fsgroup/template.yaml
```diff
@@ -3,7 +3,7 @@ kind: ConstraintTemplate
 metadata:
   name: k8spspfsgroup
   annotations:
-    description: Controls allocating an FSGroup that owns the Pod's volumes.
+    description: "Controls allocating an FSGroup that owns the Pod's volumes.  Corresponds to the `fsGroup` field in a PodSecurityPolicy."
```
maxsmythe, line 6: I wish there could be an "extended discription", so we can have a blurb and then more detailed instructions.

library/pod-security-policy/fsgroup/template.yaml
```diff
@@ -15,9 +15,15 @@ spec:
...
             rule:
+              description: "An FSGroup rule name.  Example: MayRunAs."
+              enum:
+                - MayRunAs
+                - MustRunAs
+                - RunAsAny
               type: string
             ranges:
               type: array
+              description: "GID ranges affected by the rule."
```
maxsmythe, line 26: docstring for min/max?

library/pod-security-policy/host-filesystem/template.yaml
```diff
@@ -16,13 +16,16 @@ spec:
           properties:
             allowedHostPaths:
               type: array
+              description: "An array of hostpath objects, representing paths and read/write configuration."
               items:
                 type: object
                 properties:
                   readOnly:
                     type: boolean
+                    description: "when set to true, will allow host volumes matching the pathPrefix only if all volume mounts are readOnly."
```
maxsmythe, line 25: all volume mounts or all matching volume mounts?

library/pod-security-policy/proc-mount/template.yaml
```diff
@@ -16,6 +16,10 @@ spec:
           properties:
             procMount:
               type: string
+              description: "Defines the strategy for the security exposure of certain paths in `/proc` by the container runtime."
+              enum:
+                - Default
```
maxsmythe, line 21: It'd be cool if we could document the significance of specific enum values.

maxsmythe (commented):
added some starting comments, but couldn't get through the whole thing yet

library/pod-security-policy/selinux/template.yaml
```diff
@@ -3,7 +3,7 @@ kind: ConstraintTemplate
 metadata:
   name: k8spspselinuxv2
   annotations:
-    description: Controls the SELinux context of the container.
+    description: "Defines an allow-list of seLinuxOptions configurations for pod containers.  Corresponds to a PodSecurityPolicy with SELinux rule `MustRunAs` enabled and specific `seLinuxOptions` specified."
```
maxsmythe, line 6: can we make these longer descriptions multi-lined via the `|` operator (like the rego field below). This should help readability for people using vim (or GitHub code review)

WRT text...

"and specific `seLinuxOptions` specified."

"specific" isn't very descriptive. Do you mean the options as provided to the parameters of the CT?

library/pod-security-policy/selinux/template.yaml
```diff
@@ -16,17 +16,19 @@ spec:
...
             allowedSELinuxOptions:
               type: array
+              description: "An allow-list of seLinuxOptions configurations."
               items:
                 type: object
+                description: "A single, allowed configuration of seLinuxOptions for a pod container."
                 properties:
                   level:
-                      type: string
+                    type: string
```
maxsmythe, line 25: missing descriptions for these fields

library/pod-security-policy/users/template.yaml
```diff
@@ -15,55 +15,87 @@ spec:
...
                 rule:
                   type: string
+                  description: "A strategy for applying the runAsUser restriction."
+                  enum:
+                    - MustRunAs
+                    - MustRunAsNonRoot
+                    - RunAsAny
                 ranges:
                   type: array
+                  description: "A list of ranges affected by the rule."
```
maxsmythe, line 29: Ranges of what? User IDs?

library/pod-security-policy/users/template.yaml
```diff
@@ -15,55 +15,87 @@ spec:
...
                   items:
                     type: object
+                    description: "The range of user IDs affected by the rule."
                     properties:
                       min:
```
maxsmythe, line 34: docstring needed. ex: is this inclusive or exclusive?

applies to all the min/max fields in this resource.

library/pod-security-policy/users/template.yaml
```diff
@@ -15,55 +15,87 @@ spec:
...
                     type: object
+                    description: "The range of group IDs affected by the rule."
                     properties:
                       min:
                         type: integer
                       max:
                         type: integer
             supplementalGroups:
               type: object
+              description: "Controls which groups are applied to the first process run in any pod container, in addition to the container's primary GID."
```
maxsmythe, line 62: Is there a reason we're calling out the "first process"?

library/pod-security-policy/users/template.yaml
```diff
@@ -15,55 +15,87 @@ spec:
...
                     type: object
+                    description: "The range of group IDs affected by the rule."
                     properties:
                       min:
                         type: integer
                       max:
                         type: integer
             fsGroup:
               type: object
+              description: "Controls the supplemental group applied to some volumes."
```
maxsmythe, line 84: "some volumes": which volumes?

library/pod-security-policy/volumes/template.yaml
```diff
@@ -15,6 +15,7 @@ spec:
           type: object
           properties:
             volumes:
+              description: "`volumes` is an array of volume types, recorded as strings.  All volume types can be enabled using `*`."
```
maxsmythe, line 18: No need to list the type ("recorded as strings"), it's already in the schema metadata.

maxsmythe (changes requested):
finished the review, added some more comments.

library/pod-security-policy/apparmor/template.yaml
julianKatz: Done

library/pod-security-policy/apparmor/template.yaml
julianKatz: Yes it is!  Would you prefer I put the examples there?

library/pod-security-policy/capabilities/template.yaml
julianKatz: I see what you mean.  This random [blog post](https://kubesec.io/basics/containers-securitycontext-capabilities-drop-index-all/#example) gives the same example.  This is pretty vague in the k8s docs, but I guess it's a domain that most users of the feature might already have knowledge about.

julianKatz: I've updated these descriptions to better fit my new understanding.  PTAL

library/pod-security-policy/flexvolume-drivers/template.yaml
julianKatz: Shortened.

library/pod-security-policy/fsgroup/template.yaml
julianKatz: Yea I was thinking about that.  Particularly something that would allow for some nice markdown syntax.

library/pod-security-policy/fsgroup/template.yaml
julianKatz: Added some short descriptions.  PTAL

library/pod-security-policy/host-filesystem/template.yaml
julianKatz: So, hilariously, I copied this [from the k8s libraries themselves](https://github.com/kubernetes/api/blob/96087243fecba390391364dd42abe2f3ff21c1e3/policy/v1beta1/types.go#L328).

> all volume mounts or all matching volume mounts?

So this would be all of the volume mounts that match that prefix.  So, if you had another allowed prefix with `readOnly: false` the mounts matching _that_ prefix would _not_ be required to be `readOnly`.

---

Another version of that description could be:

`When set to true, all mounts covered by the pathPrefix must be mounted readOnly.`

Thoughts?

library/pod-security-policy/proc-mount/template.yaml
julianKatz: So `Enum` [is a slice](https://github.com/kubernetes/apiextensions-apiserver/blob/master/pkg/apis/apiextensions/types_jsonschema.go) of a `JSON` [objects](https://github.com/kubernetes/apiextensions-apiserver/blob/0149c93f156a9bb358230f5fbdca12516c3e42d4/pkg/apis/apiextensions/types_jsonschema.go#L129).  That is actually an `interface{}` type, but I think that's just b/c it has to correspond to whatever is declared in `type:`.

So, I think our only option would be to write details about the enum values in the `description` on line 19.  Proposed version:

```Defines the strategy for the security exposure of certain paths in `/proc` by the container runtime. Setting to `Default` will use the runtime defaults, where `Unmasked` bypasses the default behavior.```

File: library/pod-security-policy/selinux/template.yaml (hunk 9 above)
Comment (julianKatz) [reply library/pod-security-policy/selinux/template.yaml:6]: Will change all of the templates to use the `|` operator.  It's definitely cleaner and more readable.

Comment (julianKatz) [reply library/pod-security-policy/selinux/template.yaml:6]: r.e. "specific, the idea I'm going for is the following:

If you use this template, you configure at least one set of `seLinuxOptions` that is defined as a legal configuration of options for a container to run with.  Were you to apply a container with a different set of Linux Options, that container would be rejected by g8r.

To achieve the same effect with a PSP, you'd have to create an `seLinux` entry with rule `MustRunAs` and the specific set of allowed SELinux options included under it.  That would look like:

```yaml
spec:
  seLinux:
    rule: RunAsAny
    seLinuxOptions:
      [values]
```

So I'm really just trying to say that you'd have to configure that field with some specific values of your choosing.  So I'm not being particularly _specific_ when I say specific.  I'm just trying to say that you have to configure.

How could I say it better?

maxsmythe: One thing I forgot about is newlines... I think `>` is the correct character if you want to have newlines for readability but not for content:

https://yaml.org/spec/1.2.1/#style/block/folded

library/pod-security-policy/apparmor/template.yaml
maxsmythe: Yep!

library/pod-security-policy/fsgroup/template.yaml
maxsmythe: Yeah, interesting to think how it would interact with shorter descriptions and examples...

File an issue for a feature request?

library/pod-security-policy/capabilities/template.yaml
maxsmythe: Change LGTM!

library/pod-security-policy/fsgroup/template.yaml
maxsmythe: LGTM

library/pod-security-policy/host-filesystem/template.yaml
maxsmythe: Ah, def. don't just use PSP words:

1) Because we should at least cite them if we do
2) Because the Rego may not be written exactly like the PSP Go code... particularly if the behavior isn't documented as part of the API

Looking at the template, here's what it seems like is happening:

- It's possible to use paths on the host node as volumes that can be mounted to the pods (`spec.volumes`)
   - in order to do this, you must specify that the volume is a host path type, and which host path you want to use as a volume, this path is what `pathPrefix` matches against
- A volume can be mounted to one or more paths in the pod
  - Each volume mount must reference a volume defined in `spec.volumes`
  - Each volume mount specifies where that volume gets mounted inside the pod (this value is not checked)
  - Each volume mount can be mounted as read only
  
  
It appears that the `readOnly` parameter allows users to say "for every volume mount that references this volume must mount it as read only"

library/pod-security-policy/proc-mount/template.yaml
maxsmythe: Yeah, I think you're right. Pitty, the ability to have flexible formatting options around enum docstrings could be useful.

library/pod-security-policy/selinux/template.yaml
```diff
@@ -16,17 +16,19 @@ spec:
           properties:
             allowedSELinuxOptions:
               type: array
+              description: "An allow-list of seLinuxOptions configurations."
               items:
                 type: object
+                description: "A single, allowed configuration of seLinuxOptions for a pod container."
```
maxsmythe, line 22: No need to call out that it is "single"

library/pod-security-policy/selinux/template.yaml
maxsmythe: re: specific... TBH I wouldn't mention it. It's enough to say that it is "requiring SELinux configs"

Referencing the PSP policy could be helpful for users already familiar with PSP, to help guide them when choosing replacement policies, but we shouldn't do it to the point where it confuses anyone unfamiliar with PSP and reading the docs for the first time.

library/pod-security-policy/selinux/template.yaml
julianKatz: These are actually hard to find a succinct description of without saying something meaningless like "An SELinux level".  

Kubernetes has these same kind of meaningless descriptions in [their code](https://github.com/kubernetes/api/blob/87c4113e35a1b376e4434139109105243860ea52/core/v1/types.go#L6220). 

Is something like `An SELinux level.` ok for this use-case?  Not sure if we need to be responsible for teaching people how to do SE Linux, particularly given that I don't know anything about it 😂

library/pod-security-policy/users/template.yaml
julianKatz: added

library/pod-security-policy/users/template.yaml
julianKatz: All inclusive.  Updated the descriptions.

library/pod-security-policy/users/template.yaml
julianKatz: I guess "main process" would be a better way to describe this.

I stole this from the [kubernetes `core` code](https://github.com/kubernetes/api/blob/87c4113e35a1b376e4434139109105243860ea52/core/v1/types.go#L3329).

My understanding is that other processes are forked off of that first process, so they can only have groups that it would have?  Is that right?

library/pod-security-policy/users/template.yaml
julianKatz: Also [stolen from k8s docs](https://kubernetes.io/docs/concepts/policy/pod-security-policy/#volumes-and-file-systems:~:text=FSGroup%20%2D%20Controls%20the%20supplemental%20group,any%20fsGroup%20ID%20to%20be%20specified.).

Based on [this blogpost](https://medium.com/kubernetes-tutorials/defining-privileges-and-access-control-settings-for-pods-and-containers-in-kubernetes-2cef08fc62b7#d6f3:~:text=.spec.securityContext.fsGroup%20%E2%80%94%20The%20field%20defines%20a,owning%20GID%20will%20be%20the%20fsGroup.) and [these docs](https://kubernetes.io/docs/tasks/configure-pod-container/security-context/#pods-security-security-context-yaml:~:text=Since%20fsGroup%20field%20is%20specified%2C%20all,volume%20will%20be%20Group%20ID%202000.), it looks like it sets both the volume mounted to the contianer AND any files created within it by the Pod containers have this GID.

This is complicated, and I'm not sure how to think about ROI on figuring out how it works in detail.  Is there something that's kind of vague but not wrong that we could say?  It seems like the k8s docs punted as well

library/pod-security-policy/volumes/template.yaml
julianKatz: Removed.

library/pod-security-policy/selinux/template.yaml
julianKatz: removed.

library/pod-security-policy/fsgroup/template.yaml
julianKatz: Where does this feature request go?  Isn't this part of JSONSchema?  Or are you talking about extended description for the whole resource?  Couldn't we just add another annotation for that?

maxsmythe: The second (another annotation), just an FR against the library.

library/pod-security-policy/selinux/template.yaml
maxsmythe: Yeah, a reference to some other (presumably authoritative) concept is sufficient. Teaching others about implementation details of systems we don't maintain is likely out-of-scope.

library/pod-security-policy/users/template.yaml
maxsmythe: Please reword anything inspired from others' work.

library/pod-security-policy/users/template.yaml
maxsmythe: Please reword anything inspired from others' work.

library/pod-security-policy/users/template.yaml
maxsmythe: Also, I do like "main process" better. You can do escalation via methods like `su`, so it's not strictly true that it's an inheritance tree, but I do think that's the default behavior.

library/pod-security-policy/users/template.yaml
maxsmythe: Also, we should be documenting these things based of the Rego written, for accuracy.

Finally, I'm asking "which volumes", because as a user, that would be my first question when I read something like "some volumes.

library/pod-security-policy/fsgroup/template.yaml
julianKatz: Issue created: https://github.com/open-policy-agent/gatekeeper-library/issues/112

library/pod-security-policy/host-filesystem/template.yaml
julianKatz: how about:

```
when set to true, any container volumeMounts matching the pathPrefix must include `readOnly: true`
```

library/pod-security-policy/proc-mount/template.yaml
julianKatz: updated with my proposed description

library/pod-security-policy/selinux/template.yaml
julianKatz: Are you saying I should get rid of the "Corresponds to a PodSecurityPolicy" parts of all the descriptions?  @mkorfgoogle told me he liked those.

julianKatz: Also, newest version:

```
      Defines an allow-list of seLinuxOptions configurations for pod
      containers.  Corresponds to a PodSecurityPolicy requiring SELinux configs.
```

library/pod-security-policy/selinux/template.yaml
julianKatz: Ok.  I've added some short descriptions like that to the 4 fields.

library/pod-security-policy/users/template.yaml
julianKatz: New version:

```
              description: "Controls which groups are applied to a pod container's main process, in addition to the container's primary GID."
```

library/pod-security-policy/users/template.yaml
julianKatz: I think you make a good point in [another comment](https://github.com/open-policy-agent/gatekeeper-library/pull/109#discussion_r712661299) that we shouldn't try to teach people these features, but rather how to restrict them via the CT.

That being said, I think it makes sense to change the descriptions in this CT to things like:

```yaml
fsGroup:
  type: object
  description: "Controls the fsGroup values allowed in a Pod or container level SecurityContext."
```

This is simpler and is actually what we validate in the rego.

Going to go ahead and change these, let me know if I should revert.

library/pod-security-policy/users/template.yaml
julianKatz: After reading other comments, I think it's better to go a [simpler route](https://github.com/open-policy-agent/gatekeeper-library/pull/109#discussion_r714288228), where we don't try to teach the feature in the CT.  Rather, we should say specifically what policy the CT enforces and no more.

library/pod-security-policy/users/template.yaml
```diff
@@ -18,7 +18,7 @@ spec:
           properties:
             runAsUser:
               type: object
-              description: "Controls which user ID pod containers can be run with."
+              description: "Controls which user ID values allowed in a Pod or container level SecurityContext."
```
maxsmythe, line 21: nit: "are allowed", here and below

library/pod-security-policy/fsgroup/template.yaml
```diff
@@ -15,15 +17,23 @@ spec:
           type: object
           properties:
             rule:
+              description: "An FSGroup rule name.  Example: MayRunAs."
+              enum:
+                - MayRunAs
```
maxsmythe, line 22: When we are parsing schemas, we can probably list allowed enum values.

maxsmythe (approved):
LGTM with a few nits!

library/pod-security-policy/users/template.yaml
julianKatz: fixed

library/pod-security-policy/fsgroup/template.yaml
julianKatz: removed `. Example: MayRunAs.`

library/pod-security-policy/selinux/template.yaml
```diff
@@ -16,17 +18,23 @@ spec:
           properties:
             allowedSELinuxOptions:
               type: array
+              description: "An allow-list of seLinuxOptions configurations."
```
willbeason, line 21: seLinux -> SELinux?

library/pod-security-policy/proc-mount/template.yaml
```diff
@@ -16,6 +18,14 @@ spec:
           properties:
             procMount:
               type: string
+              description: >-
+                Defines the strategy for the security exposure of certain paths
+                in `/proc` by the container runtime. Setting to `Default` will
```
mkorfgoogle, line 23: Avoid future tense. "Setting to 'Default' uses the runtime defaults, where ....

library/pod-security-policy/read-only-root-filesystem/template.yaml
```diff
@@ -3,7 +3,10 @@ kind: ConstraintTemplate
 metadata:
   name: k8spspreadonlyrootfilesystem
   annotations:
-    description: Requires the use of a read only root file system.
+    description: >-
+      Requires the use of a read only root file system by pod containers.
```
mkorfgoogle, line 7: Minor nit, hyphenate read-only

library/pod-security-policy/users/template.yaml
```diff
@@ -15,60 +18,100 @@ spec:
...
                     properties:
                       min:
                         type: integer
+                        description: "The minimum user ID in the range, inclusive."
                       max:
                         type: integer
+                        description: "The maximum user ID in the range, inclusive."
             runAsGroup:
               type: object
+              description: "Controls which group ID values are allowed in a Pod or container level SecurityContext."
```
mkorfgoogle, line 45: I don't think Pod is capitalized, and container-level should be hyphenated when used as an adjective. "Controls which group ID values are allowed in a pod or container-level SecurityContext."

library/pod-security-policy/users/template.yaml
```diff
@@ -15,60 +18,100 @@ spec:
...
                     properties:
                       min:
                         type: integer
+                        description: "The minimum group ID in the range, inclusive."
                       max:
                         type: integer
+                        description: "The maximum group ID in the range, inclusive."
             supplementalGroups:
               type: object
+              description: "Controls the supplementalGroups values are allowed in a Pod or container level SecurityContext."
```
mkorfgoogle, line 69: "Controls the supplementalGroups values that are allowed in a Pod or container-level SecurityContext."

library/pod-security-policy/users/template.yaml
```diff
@@ -15,60 +18,100 @@ spec:
...
                     properties:
                       min:
                         type: integer
+                        description: "The minimum group ID in the range, inclusive."
                       max:
                         type: integer
+                        description: "The maximum group ID in the range, inclusive."
             fsGroup:
               type: object
+              description: "Controls the fsGroup values are allowed in a Pod or container level SecurityContext."
```
mkorfgoogle, line 93: "Controls the fsGroup values that are allowed in a Pod or container-level SecurityContext."

library/pod-security-policy/allow-privilege-escalation/template.yaml
```diff
@@ -3,7 +3,9 @@ kind: ConstraintTemplate
 metadata:
   name: k8spspallowprivilegeescalationcontainer
   annotations:
-    description: Controls restricting escalation to root privileges.
+    description: >-
+      Controls restricting escalation to root privileges. Corresponds to the
+      `allowPrivilegeEscalation` field in a PodSecurityPolicy.
```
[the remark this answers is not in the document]
maxsmythe: For PSP specifically, the main purpose is to provide guidance for people who are referencing the PSP fields/ documentation.

The PSP documentation references its fields by non-fully-qualified name:

https://kubernetes.io/docs/concepts/policy/pod-security-policy/

So we should follow their conventions.

Since PSP is deprecated, we may choose to remove these references entirely once most/all users are no longer in a transitional phase.

If we are talking about related fields within the same constraint template, that seems reasonable to me, though I'd be a bit concerned about verbosity. Maybe make all paths relative to `parameters`?

library/pod-security-policy/apparmor/template.yaml
```diff
@@ -15,6 +17,7 @@ spec:
           type: object
           properties:
             allowedProfiles:
+              description: "An array of AppArmor profiles. Examples: `runtime/default`, `unconfined`."
```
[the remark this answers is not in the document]
maxsmythe: +1, though maybe add that to the template description instead of the field description?

library/pod-security-policy/host-filesystem/template.yaml
```diff
@@ -16,13 +18,16 @@ spec:
...
               type: array
+              description: "An array of hostpath objects, representing paths and read/write configuration."
               items:
                 type: object
                 properties:
-                  readOnly:
-                    type: boolean
                   pathPrefix:
                     type: string
+                    description: "The path prefix that the host volume must match."
```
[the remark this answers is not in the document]
maxsmythe: Rather than invoking K8s resource specifics (which we are attempting to insulate users from), could we change this to:

"The prefix that the host path backing the host volume must match"

In general: prefer intent-based descriptions vs. schema-based descriptions, b/c the schema may change across resource versions while the intent stays the same

library/pod-security-policy/capabilities/template.yaml
```diff
@@ -16,10 +19,12 @@ spec:
           properties:
             allowedCapabilities:
               type: array
+              description: "A list of Linux capabilities that can be added to a container."
```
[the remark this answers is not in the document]
maxsmythe: In general, I think background conceptual context should go in the constraint's description annotation, since that is where it would read most naturally in a markdown doc.

For people accessing the fields via `kubectl explain`, we should assume that they have seen the top-level `DESCRIPTION` output.

That adds a second note though...

`metadata.annnotations.description` has two key flaws:

1. The name is wrong: annotations should be namespaced to gatekeeper.sh
2. The annotation doesn't show up with `kubectl explain`

And one benefit, namely that annotations come at the top of the YAML

We should really be doing the following:

1. Putting what we are currently putting in `metadata.annotations.description` in the description for the top-level field of the CT's jsonschema (`spec.crd.spec.validation.openAPIV3Schema.description`)
2. When generating a CRD from the CT, using that root `description` field, as the root `description` field for the CRD's schema

That will work with `kubectl explain`

Unfortunately, I don't want to just get rid of the current `description` annotation in case anyone relies on them.

For now, can we copy/paste the description annotation and just have duplicated strings (we can do this after all editorial comments are dealt with).

We can deal with the actual CRD generation in step (2) later.

library/pod-security-policy/users/template.yaml
willbeason: Pod capitalized refers to the Kubernetes resource, so it is correct for it to be capitalized.

library/pod-security-policy/host-filesystem/template.yaml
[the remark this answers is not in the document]
willbeason: Then what about a link to the schema? It can be a pain to find an example of the schema we're referencing.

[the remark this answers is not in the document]
maxsmythe: That sounds like it would be covered by adding "examples" documentation that shows example allowed/disallowed resources, which we can source from test data?

library/pod-security-policy/apparmor/template.yaml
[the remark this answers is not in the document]
julianKatz: Agreed with @maxsmythe on full template description.

New version:

```yaml
    description: >-
      Configures an allow-list of AppArmor profiles for use by containers.
      This corresponds to specific annotations applied to a PodSecurityPolicy.
      For information on AppArmor, see
      https://kubernetes.io/docs/tutorials/clusters/apparmor/
```

LMK if anyone dislikes otherwise will resolve

library/pod-security-policy/capabilities/template.yaml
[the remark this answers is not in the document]
julianKatz: I've made an issue (open-policy-agent/frameworks#140) requesting better support for `kubectl explain`.

A manual test I did suggests development work will be needed to get the description plumbed through:

```
❯ kubectl delete constrainttemplates k8srequiredlabels
constrainttemplate.templates.gatekeeper.sh "k8srequiredlabels" deleted

# APPLIES THE FOLLOWING:
#
# apiVersion: templates.gatekeeper.sh/v1beta1
# kind: ConstraintTemplate
# metadata:
#   name: k8srequiredlabels
# spec:
#   crd:
#     spec:
#       names:
#         kind: K8sRequiredLabels
#       validation:
#         # Schema for the `parameters` field
#         openAPIV3Schema:
#           description: "CARNE ASADA BURRITO"
#           properties:
#             labels:
#               type: array
#               items:
#                 type: string
#   targets:
#     - target: admission.k8s.gatekeeper.sh
#       rego: |
#         package k8srequiredlabels
# 
#         violation[{"msg": msg, "details": {"missing_labels": missing}}] {
#           provided := {label | input.review.object.metadata.labels[label]}
#           required := {label | label := input.parameters.labels[_]}
#           missing := required - provided
#           count(missing) > 0
#           msg := sprintf("you must provide labels: %v", [missing])
#         }
❯ kubectl apply -f example/templates/k8srequiredlabels_template.yaml
constrainttemplate.templates.gatekeeper.sh/k8srequiredlabels unchanged

❯ kubectl explain k8srequiredlabels
KIND:     K8sRequiredLabels
VERSION:  constraints.gatekeeper.sh/v1beta1

DESCRIPTION:
     <empty>

FIELDS:
   apiVersion   <string>
     APIVersion defines the versioned schema of this representation of an
     object. Servers should convert recognized schemas to the latest internal
     value, and may reject unrecognized values. More info:
     https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#resources

   kind <string>
     Kind is a string value representing the REST resource this object
     represents. Servers may infer this from the endpoint the client submits
     requests to. Cannot be updated. In CamelCase. More info:
     https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#types-kinds

   metadata     <Object>
     Standard object's metadata. More info:
     https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#metadata

   spec <Object>

   status       <>
```

I'll leave this Action Item for myself once I'm done with prose edits:

- [ ] Copy all `metadata.annotations.description` values to `spec.crd.spec.validation.openAPIV3Schema.description`

library/pod-security-policy/forbidden-sysctls/template.yaml
```diff
@@ -16,6 +18,7 @@ spec:
           properties:
             forbiddenSysctls:
               type: array
+              description: "A disallow-list of sysctls. `*` forbids all sysctls."
```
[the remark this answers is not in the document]
julianKatz: @maxsmythe and I talked in [another comment](https://github.com/open-policy-agent/gatekeeper-library/pull/109#discussion_r712661299) about overextending ourselves in teaching users "how to PSP" in these templates.  With that in mind, I think straying into safe vs. unsafe sysctls is a bit beyond the CT's scope.

Part of why I say that is the existence of the [allowedUnsafeSysctls](https://kubernetes.io/docs/concepts/policy/pod-security-policy/#sysctl:~:text=allowedUnsafeSysctls%20%2D%20allows%20specific%20sysctls%20that,these%20are%20not%20listed%20in%20forbiddenSysctls.) PSP field, which is an additional part of this feature set.  I feel like digging into the unsafe ones would require us to explain this whole story.

library/pod-security-policy/fsgroup/template.yaml
```diff
@@ -15,15 +17,23 @@ spec:
           type: object
           properties:
             rule:
+              description: "An FSGroup rule name."
```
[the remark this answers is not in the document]
julianKatz: I've added a link in the CT description to the Volumes and File Systems section

library/pod-security-policy/selinux/template.yaml
julianKatz: New version:

```yaml
                description: "An allowed configuration of SELinux options for a pod container."
```

LMK if you disagree

library/pod-security-policy/fsgroup/template.yaml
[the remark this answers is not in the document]
julianKatz: I'm going to add this link to all the templates that are related to that section.

library/pod-security-policy/users/template.yaml
julianKatz: Looks like [k8s docs](https://kubernetes.io/docs/concepts/workloads/pods/) use it like a proper noun.

library/pod-security-policy/selinux/template.yaml
willbeason: Sounds good.

library/pod-security-policy/fsgroup/template.yaml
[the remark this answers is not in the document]
willbeason: Sounds good.

library/pod-security-policy/forbidden-sysctls/template.yaml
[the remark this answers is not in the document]
willbeason: LGTM

library/pod-security-policy/capabilities/template.yaml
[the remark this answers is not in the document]
willbeason: LGTM

library/pod-security-policy/apparmor/template.yaml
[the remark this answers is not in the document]
willbeason: SG

library/pod-security-policy/allow-privilege-escalation/template.yaml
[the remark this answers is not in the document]
willbeason: I'm fine either way.

willbeason (approved):

library/pod-security-policy/allow-privilege-escalation/template.yaml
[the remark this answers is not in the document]
maxsmythe: I lean more towards matching the PSP docs (meaning what's written is fine as-is)

library/pod-security-policy/allow-privilege-escalation/template.yaml
```diff
@@ -3,12 +3,24 @@ kind: ConstraintTemplate
...
+      `allowPrivilegeEscalation` field in a PodSecurityPolicy. For more
+      information, see
+      https://kubernetes.io/docs/concepts/policy/pod-security-policy/#privilege-escalation
 spec:
   crd:
     spec:
       names:
         kind: K8sPSPAllowPrivilegeEscalationContainer
+      validation:
+        # Schema for the `parameters` field
```
ritazh, line 17: This template does not have parameters. Would this comment confuse people; expecting it to have parameters?

sozercan: this applies to privileged policy too

julianKatz: removed for both

sozercan (approved):
LGTM

PR #110, opened 2021-09-16
Treat "ALL" capability as special

jbrook:
When checking required drop capabilities, if the container drops "ALL" that should not trigger a violation even if other specifc capabilites are in the requiredDropCapabilities parameter.

### DIFF
--- a/src/pod-security-policy/capabilities/src_test.rego
+++ b/src/pod-security-policy/capabilities/src_test.rego
@@ -242,6 +242,11 @@ test_input_drop_literal_all_with_all_param {
    results := violation with input as input
    count(results) == 0
 }
+test_input_drop_literal_with_all_param {
+   input := { "review": input_init_review([cdrop(["one"])]), "parameters": {"requiredDropCapabilities": ["one", "ALL"]}}
+   results := violation with input as input
+   count(results) == 1
+}
 test_input_drop_literal_all_x2 {
    input := { "review": input_init_review([cdrop(["ALL", "two"])]), "parameters": {"requiredDropCapabilities": ["one", "two"]}}
    results := violation with input as input
--- a/src/pod-security-policy/capabilities/src_test.rego
+++ b/src/pod-security-policy/capabilities/src_test.rego
@@ -104,19 +104,24 @@ test_input_drop_undefined_x2 {
    count(results) == 2
 }
 test_input_drop_literal_all {
+   input := { "review": input_review([cdrop(["ALL"])]), "parameters": {"requiredDropCapabilities": ["one", "two"]}}
+   results := violation with input as input
+   count(results) == 0
+}
+test_input_drop_literal_all_lower {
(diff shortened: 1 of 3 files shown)


src/pod-security-policy/capabilities/src_test.rego
```diff
@@ -103,6 +103,21 @@ test_input_drop_undefined_x2 {
    results := violation with input as input
    count(results) == 2
 }
+test_input_drop_literal_all {
+   input := { "review": input_review([cdrop(["all"])]), "parameters": {"requiredDropCapabilities": ["one", "two"]}}
```
maxsmythe, line 107: `ALL` instead of `all` here and below?

src/pod-security-policy/capabilities/src_test.rego
```diff
@@ -103,6 +103,21 @@ test_input_drop_undefined_x2 {
...
+   results := violation with input as input
+   count(results) == 1
```
maxsmythe, line 109: We would expect `count(results) == 0` if there were no violations

maxsmythe (changes requested):
Thank you for the PR! I think there are a couple test bugs, and the CLA.

src/pod-security-policy/capabilities/src_test.rego
jbrook: I have now made the "ALL" comparison case insensitive.

src/pod-security-policy/capabilities/src_test.rego
jbrook: Fixed

src/pod-security-policy/capabilities/src_test.rego
maxsmythe: Just to verify... kubernetes interprets "ALL" in a case-insensitive manner?

maxsmythe (approved):
LGTM!

@ritazh @shomron @sozercan LGTY?

Also, we discovered that the capabilities fields are case-insensitive, so updated the template to reflect that

src/pod-security-policy/capabilities/src_test.rego
```diff
@@ -207,6 +227,27 @@ test_input_drop_undefined_x2 {
...
+   input := { "review": input_init_review([cdrop(["ALL"])]), "parameters": {"requiredDropCapabilities": ["one", "two"]}}
+   results := violation with input as input
+   count(results) == 0
+}
+test_input_drop_literal_all_lower {
+   input := { "review": input_init_review([cdrop(["all"])]), "parameters": {"requiredDropCapabilities": ["one", "two"]}}
+   results := violation with input as input
+   count(results) == 0
+}
+test_input_drop_literal_all_with_all_param {
```
ritazh, line 240: can we add a test for test_input_drop_literal_with_all_param where dropped does not include `all` and count(results) > 0?

maxsmythe: @ritazh added the test

ritazh (approved):
Thank you for the PR!

PR #111, opened 2021-09-21
Add example suite.yaml files for gator test

willbeason:
These can be run individually, or all at once with:

    gator test ./...

Signed-off-by: Will Beason <willbeason@google.com>


maxsmythe (approved):
LGTM

julianKatz (approved):

library/general/block-endpoint-edit-default-role/suite.yaml
```diff
@@ -0,0 +1,13 @@
+kind: Suite
+apiVersion: test.gatekeeper.sh/v1alpha1
+tests:
+- name: allowed-repos
```
ritazh, line 4: nit: update the test name?

library/general/block-nodeport-services/suite.yaml
```diff
@@ -0,0 +1,14 @@
+kind: Suite
+apiVersion: test.gatekeeper.sh/v1alpha1
+tests:
+- name: allowed-repos
```
ritazh, line 4: ditto

library/general/containerlimits/suite.yaml
```diff
@@ -0,0 +1,13 @@
+kind: Suite
+apiVersion: test.gatekeeper.sh/v1alpha1
+tests:
+- name: allowed-repos
```
ritazh, line 4: ditto

library/general/allowedrepos/suite.yaml
```diff
@@ -0,0 +1,13 @@
+kind: Suite
+apiVersion: test.gatekeeper.sh/v1alpha1
+tests:
+- name: allowed-repos
+  template: template.yaml
+  constraint: samples/repo-must-be-openpolicyagent/constraint.yaml
+  cases:
+  - name: example-allowed
+    object: samples/repo-must-be-openpolicyagent/example_allowed.yaml
```
ritazh, line 9: Question: if there's no assertion, does it default to `violations: no`? if so, when does a user want to explicitly use assertions with `violation: no`?

maxsmythe: I think this thread is most relevant:

https://github.com/open-policy-agent/gatekeeper/pull/1519#discussion_r697701073

Agreed that zero assertions => implied assertion that the object will pass validation.

IMO it should be that zero assertions => zero tests against the object => the test can never fail

maxsmythe: If we leave the behavior as-is, we risk winding up with weird edge cases if the tests become more complex (e.g. imagine we are not testing validation)

library/general/block-nodeport-services/suite.yaml
willbeason: Done.

library/general/block-endpoint-edit-default-role/suite.yaml
willbeason: Done.

library/general/allowedrepos/suite.yaml
ritazh: wdyt @willbeason ?

willbeason: I made the change you suggested.

ritazh: thanks! but does that mean gator doesnt require assertions?

ritazh (approved):

library/general/allowedrepos/suite.yaml
maxsmythe: Will told me today he wanted to change it so that a test with no assertions is always invalid and will fail.

PR #116, opened 2021-09-29
Enrich some template under library/general/

julianKatz:
Signed-off-by: juliankatz <juliankatz@google.com>


library/general/containerlimits/template.yaml
```diff
@@ -3,8 +3,11 @@ kind: ConstraintTemplate
 metadata:
   name: k8scontainerlimits
   annotations:
-    description: Requires containers to have memory and CPU limits set and within
-      a specified maximum amount.
+    description: >-
+      Requires containers to have memory and CPU limits set and constrains
+      limits to set maximum values.
```
maxsmythe, line 8: double use of the word "set" with different meanings can be confusing.

"to be within the specified maximum values"

library/general/disallowedtags/template.yaml
```diff
@@ -3,8 +3,11 @@ kind: ConstraintTemplate
 metadata:
   name: k8sdisallowedtags
   annotations:
-    description: Requires container images to have an image tag different
-      from the ones in a specified list.
+    description: >-
+      Requires container images to have an image tag different from the ones in
+      a specified list.
```
maxsmythe, line 8: "the specified list"

library/general/externalip/template.yaml
```diff
@@ -16,7 +19,9 @@ spec:
           properties:
             allowedIPs:
               type: array
+              description: "An allow-list of external IP addresses."
               items:
+                description: "A single IP address."
```
maxsmythe, line 24: Do we docstring individual items in a list when the list is of a primitive type?

library/general/httpsonly/template.yaml
```diff
@@ -3,8 +3,15 @@ kind: ConstraintTemplate
 metadata:
   name: k8shttpsonly
   annotations:
-    description: Requires Ingress resources to be HTTPS only; TLS configuration should
-      be set and `kubernetes.io/ingress.allow-http` annotation equals false.
+    description: >-
+      Requires Ingress resources to be HTTPS only.
+
+      Ingress resources must:
+      - include a valid TLS configuration
```
maxsmythe, line 10: Will this list format correctly after YAML wrapping?

maxsmythe (commented):
Mostly LGTM, couple comments

library/general/externalip/template.yaml
willbeason: +1 I think the description for the array is sufficient.

library/general/httpsonly/template.yaml
willbeason: No, it will not. With `>` newlines are replaced with spaces so this list will become one line.

willbeason (approved):

library/general/containerlimits/template.yaml
julianKatz: Done

library/general/httpsonly/template.yaml
julianKatz: This one isn't long enough to warrant complex formatting, but it looks like we can use `|-` to keep things like newlines (we do this with rego).  It'll be worth some testing in the future to see if we'd like to pursue more formatting in our descriptions.

library/general/externalip/template.yaml
julianKatz: removed

library/general/disallowedtags/template.yaml
julianKatz: Done

maxsmythe (approved):
LGTM!

library/general/httpsonly/template.yaml
willbeason: +1 for `|-`

Do we prever `>-` for some reason?

maxsmythe: Wrapping long lines in the YAML that we don't want to be wrapped universally.

Otherwise we are coupling the GitHub code review line width (which is rather short), with HTTP text wrapping (where the flow should be determined by the screen width, see [responsive web design](https://en.wikipedia.org/wiki/Responsive_web_design))

sozercan:
@julianKatz looks like these are failing integration tests

ritazh (approved):

mkorfgoogle:
LGTM

PR #124, opened 2021-10-13
Add examples mutations for initContainers

sathieu:


sathieu:
NB: This needs to be done for k8spspdefaultallowprivilegeescalation too.

PR #130, opened 2021-11-05
Add gator test suites for general/

willbeason:
The only significant changes I made to existing tests were:

1. Adding tests when an allowed example did not exist
2. Specifying Namespace on namespace-scoped objects used in referential
constraints, since gator test can't infer which is which
3. Breaking apart multi-object files used for referential constraints to match the gator test format

These may be tested by running "gator test ./..." from the repository
root


julianKatz (approved):

sozercan:
are we planning on adding gator suites in CI?

library/general/allowedrepos/suite.yaml
```diff
@@ -9,7 +9,23 @@ tests:
...
+    object: samples/repo-must-be-openpolicyagent/disallowed_container.yaml
     assertions:
     - violations: yes
+      message: container
+  - name: initcontainer-disallowed
+    object: samples/repo-must-be-openpolicyagent/disallowed_initcontainer.yaml
+    assertions:
+    - violations: 1
+      message: initContainer
+    - violations: 0
```
sozercan, line 22: do we want to make violations consistent? i see some `yes` and `no` in the other suites

test/bats/test.bats
```diff
@@ -90,6 +90,14 @@ setup() {
           fi
         done
 
+        for inventory in "$sample"/example_inventory*.yaml; do
```
sozercan, line 93: what is the inventory used for?

maxsmythe: populating external data

maxsmythe (approved):

library/general/allowedrepos/suite.yaml
willbeason: I wanted to have working examples I can point to from documentation, so I made tests use the full range of violations.

willbeason:
> are we planning on adding gator suites in CI?

Yep, we want to replace the existing bash tests with gator test. This is faster (5 seconds vs. 10+ minutes) and can run locally without setup.

PR #131, opened 2021-11-11
Add docstrings for K8sRequiredProbes

maxsmythe:


willbeason (approved):

library/general/requiredprobes/template.yaml
```diff
@@ -14,10 +14,12 @@ spec:
...
           properties:
             probes:
+              description: "A list of probes that are required (ex: `readinessProbe`)"
               type: array
               items:
                 type: string
             probeTypes:
+              description: "The probe must implement one of the `probeType` in order to satisfy the constraint (ex. `tcpSocket` satisfies `['tcpSocket', 'exec']`)"
               type: array
               items:
```
julianKatz, line 24: we can add an `enum` with the allowed values from [the docs](https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/#configure-probes)

julianKatz: `readinessProbe`, `livenessProbe`

julianKatz (approved):

library/general/requiredprobes/template.yaml
maxsmythe: Unfortunately that would be a backwards-incompatible change.

maxsmythe: Also, those particular values would go under `probe` as this constraint template is written.

julianKatz (approved):

sozercan (approved):

PR #138, opened 2021-11-17
Add suite.yaml files for half of PSP templates

julianKatz:
Signed-off-by: juliankatz <juliankatz@google.com>


library/pod-security-policy/capabilities/suite.yaml
```diff
@@ -0,0 +1,11 @@
+kind: Suite
+apiVersion: test.gatekeeper.sh/v1alpha1
+tests:
+  - name: capabilities
+    template: template.yaml
+    constraint: samples/capabilities-demo/constraint.yaml
+    cases:
```
willbeason, line 7: Add an allowed example (ditto others - it's important to verify that we document what is allowed)

willbeason (changes requested):

library/pod-security-policy/capabilities/suite.yaml
julianKatz: Done!

willbeason (approved):

build/gator/Dockerfile
```diff
@@ -0,0 +1,15 @@
+FROM alpine
```
sozercan, line 1: can we use a multi stage build to copy the binary to a distroless/static container?

sozercan: and let's parametrize the version

julianKatz: > can we use a multi stage build to copy the binary to a distroless/static container?

While I'm a big fan of distroless containers for usecases where something needs to be either lightweight or highly-secure, I find them to be a pain for usecases like this one.  As soon as something doesn't work, I'm stuck debugging an image that doesn't have a shell.

Is this a change you feel strongly about?

julianKatz: > and let's parametrize the version as a build arg and Makefile/CI can use GATEKEEPER_VERSION we can update

Sounds good.  Will do.

maxsmythe (approved):
LGTM

build/gator/Dockerfile
julianKatz: from slack: @sozercan suggested using the [distroless debug image](https://github.com/GoogleContainerTools/distroless#debug-images), which solves my debugging problem.  That's a good compromise.  Will change.

julianKatz: SO, it turns out that our binary needs to be statically compiled (i.e. with `CGO_ENABLED=0`) for it to work inside a distroless container.  See this [example](https://github.com/GoogleContainerTools/distroless/blob/main/examples/go/Dockerfile) from the distroless project.

The most recent release [didn't do that](https://github.com/open-policy-agent/gatekeeper/blob/master/Makefile#L402).

I've made a [gatekeeper issue](https://github.com/open-policy-agent/gatekeeper/issues/1732) to do the static build, and an [issue](https://github.com/open-policy-agent/gatekeeper-library/issues/153) in this repository to use it once it's available.

For now, I think we should punt on this and move forward.  This is just available in the `Makefile` for people who dont' want to download `gator` themselves as a convenience.  There's nothing running continuously that needs to be resistant to hacking, AFAICT.

sozercan (approved):

PR #147, opened 2021-12-02
Explain fields for requiredlabel/annotation CTs

willbeason:


library/general/requiredannotations/template.yaml
```diff
@@ -18,13 +19,20 @@ spec:
...
                 type: object
                 properties:
                   key:
                     type: string
+                    description: >-
+                      The required annotation.
                   allowedRegex:
                     type: string
+                    description: >-
+                      A regular expression the annotation's value must match. The value
```
maxsmythe, line 34: We should probably mention that this field is optional, and that the regex wont be checked if `allowedRegex` is undefined.

willbeason: Done!

library/general/requiredannotations/template.yaml
```diff
@@ -18,13 +19,21 @@ spec:
...
+                      must contain at least one match for the regular expression.
+                      If unset, or set to empty string, matches all values.
```
maxsmythe, line 36: nit: I think we can ignore the empty-string case... is an empty string regex one that matches everything?

maxsmythe (approved):
LGTM

library/general/requiredannotations/template.yaml
willbeason: Sort of - the wording is weird. The empty string regex has matches in all existent strings. It doesn't "match" the text (like ".*" would), but it "has matches" in the text.

julianKatz (approved):
If Max's comment is resolved, LGTM

library/general/requiredannotations/template.yaml
maxsmythe: That sounds like it's covered by "must contain at least one match for the regular expression"

For the undefined case, you can prefix the first sentence with "When defined, a regular expression..."

Basically, it'd be good to make it clear that users can ignore this field unless they want to leverage it.

willbeason: Done.

julianKatz:
ready to merge?

sozercan (approved):

PR #148, opened 2021-12-04
Add image-based container exemptions to PSP policies.

maxsmythe:
This allows for exempting containers from constraints via the container's image prefix, which should make it possible to inject sidecars from trusted sources without worrying about policy conflicts.

### DIFF
--- a/library/general/containerlimits/template.yaml
+++ b/library/general/containerlimits/template.yaml
@@ -18,6 +18,13 @@ spec:
         openAPIV3Schema:
           type: object
           properties:
+            exemptImagePrefixes:
+              description: >-
+                Any container that uses an image that matches a prefix in this list will be excluded
+                from enforcement.
+              type: array
+              items:
+                type: string
             cpu:
               description: "The maximum allowed cpu limit on a Pod, exclusive."
               type: string
@@ -28,6 +35,8 @@ spec:
     - target: admission.k8s.gatekeeper.sh
       rego: |
         package k8scontainerlimits
+
+        import data.lib.exempt_container.is_exempt
 
         missing(obj, field) = true {
           not obj[field]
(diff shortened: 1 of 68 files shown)


library/general/containerlimits/template.yaml
```diff
@@ -210,10 +226,21 @@ spec:
...
         }
+      libs:
+        - |
+          package lib.exempt_container
+
+          is_exempt(container) {
+            exempt_image_prefixes := object.get(object.get(input, "parameters", {}), "exemptImagePrefixes", [])
+            img := container.image
+            prefix := exempt_image_prefixes[_]
+            startswith(img, prefix)
```
shomron, line 245: Are there any ways this might be circumvented by playing tricks with Docker image reference parsing idiosyncrasies? e.g. `library/alpine` vs.`alpine`, or `foobar/alpine` vs. `foo.bar/alpine`?

sozercan:
in the community call, we talked about using similar format as `exemptNamespacePrefix`, i think this is same convention as that?

maxsmythe:
@sozercan yeah, we should make it use the same syntax as namespace prefixing.

library/general/containerlimits/template.yaml
maxsmythe: Is there some convention people use to make sure the prefix of the docker container is absolutely resolved (i.e. represents the host repository).

There may be some element of users needing to know how to do the right thing here.

maxsmythe: It looks like it'd be on users to specify a fully-qualified image name:

https://www.redhat.com/en/blog/be-careful-when-pulling-images-short-name

I'm not sure there is anything we can do to make it better?

shomron: It might be worth documenting a warning that highlights the potential ambiguity and urges users to fully qualify their prefixes and images.

maxsmythe: That's a good idea, we can add it to the description.

maxsmythe:
@sozercan @shomron LMK if the changes LGTY

sozercan (approved):
Thank you! LGTM

Issue #120, opened 2021-10-08
Allow specifying allowed images for disallowedtags


sathieu:
Allow to disallow all `latest` images, except a whitelist.

Expected usage:

 ```yaml
apiVersion: constraints.gatekeeper.sh/v1beta1
kind: K8sDisallowedTags
metadata:
  name: container-image-must-not-have-latest-tag
spec:
  match:
    kinds:
      - apiGroups: [""]
        kinds: ["Pod"]
    namespaces:
      - "default"
  parameters:
    tags:
    - latest
    allowedImages:
    - foo/bar:latest
```

PR #151, opened 2021-12-08
Exempt allowImages for K8sDisallowedTags

becky-hd:
Signed-off-by: Becky Huang <beckyhd@google.com>

Support the images in the allowlist `allowedImages` being exempted from K8sDisallowedTags constraint.

Fixes [#120 ](https://github.com/open-policy-agent/gatekeeper-library/issues/120)


library/general/disallowedtags/template.yaml
```diff
@@ -23,20 +23,29 @@ spec:
...
+                type: string
   targets:
     - target: admission.k8s.gatekeeper.sh
       rego: |
         package k8sdisallowedtags
 
         violation[{"msg": msg}] {
           container := input_containers[_]
+          images := [allowed | image = input.parameters.allowedImages[_]; allowed = (image == container.image) ]
+          not all(images)
```
maxsmythe, line 39: It looks like what you're trying to do is scroll through the list of exceptions, marking true if an exception is matched, false otherwise, then if there is at least one `true`, the exception is granted. Is that accurate?

If so, I think the correct word would be `not any(images)`

Also, nit on the variable name `images`, since it's not storing a list of images, maybe: `allowed_images_matches`?

maxsmythe: Then again `not any()` would fail in the case where there were no allowed images

maxsmythe: Maybe building the allowedImages check in a function would be easier? You could use the piecewise function behavior to emulate or-style logic.

becky-hd: I updated tests with 2 containers checking:
- `exempt-image:disallowed-tag` and `another-image:accept-tag` should have no violation
- `exempt-image:disallowed-tag` and `another-image:disallowed-tag` should be violated

strangely they both passed the test under `not any(matched_allowed_images)` and  `not all(matched_allowed_images)`

maxsmythe: Sorry, was overthinking with `not any`, if there are no allowed images, this should return `true` always, which is what we want.

maxsmythe: The difference in behavior between `any` and `all` only shows up if there are multiple allowed images.

becky-hd: Updated the tests with 2 allowed images, turns out `not all` is not a valid.

maxsmythe (approved):
LGTM

sozercan:
@becky-hd looks like `library/general/.DS_Store` is being checked in, can we remove that file?

becky-hd:
> 

oops, removed.

sozercan:
@becky-hd  can we update the rego src and add tests for this too? https://github.com/open-policy-agent/gatekeeper-library/tree/master/src/general/disallowedtags

sozercan (changes requested):
Waiting for src and src_test rego

sozercan:
@becky-hd should we align this with #148?

becky-hd:
> @becky-hd should we align this with #148?

Would it make sense to use `data.lib.exempt_container.is_exempt` for image exemption here? @maxsmythe

maxsmythe:
@becky-hd 

I think it makes sense.

sozercan:
@becky-hd looks like opa binary is being pushed

sozercan:
LGTM otherwise

becky-hd:
> @becky-hd looks like opa binary is being pushed

thanks for the catch, it is removed.

`opa test src/general/disallowedtags/*.rego` locally threw some errors, running test for other existing lib(e.g. src/general/containerlimits/) got  the same errors, has anyone had the same issue?

src/general/disallowedtags/lib_exempt_container.rego:3: rego_unsafe_var_error: var _ is unsafe
src/general/disallowedtags/lib_exempt_container.rego:17: rego_unsafe_var_error: expression is unsafe

src/general/disallowedtags/src_test.rego
```diff
@@ -92,6 +92,21 @@ test_input_denied_mixed_container_latest {
     results := violation with input as input
     count(results) == 1
 }
+test_input_denied_mixed_container_with_some_exempt_image {
+    input := { "review": input_init_review(array.concat(input_container_exempt, input_container_denied_latest)), "parameters": {"tags": ["latest", "testing"], "allowedImages": ["exempt:latest"]}}
```
sozercan, line 96:
```suggestion
    input := { "review": input_init_review(array.concat(input_container_exempt, input_container_denied_latest)), "parameters": {"tags": ["latest", "testing"], "exemptImages": ["exempt:latest"]}}
```

src/general/disallowedtags/src_test.rego
```diff
@@ -92,6 +92,21 @@ test_input_denied_mixed_container_latest {
...
+    results := violation with input as input
+    count(results) == 2
+}
+test_input_denied_dual_container_with_all_exempt_image {
+    input := { "review": input_init_review(array.concat(input_container_exempt, input_container_denied_latest)), "parameters": {"tags": ["latest", "testing"], "allowedImages": ["exempt:latest", "exempt:testing"]}}
```
sozercan, line 101:
```suggestion
    input := { "review": input_init_review(array.concat(input_container_exempt, input_container_denied_latest)), "parameters": {"tags": ["latest", "testing"], "exemptImages": ["exempt:latest", "exempt:testing"]}}
```

src/general/disallowedtags/src_test.rego
```diff
@@ -92,6 +92,21 @@ test_input_denied_mixed_container_latest {
...
+    results := violation with input as input
+    count(results) == 1
+}
+test_input_allowed_dual_container_with_exempt_image {
+    input := { "review": input_init_review(input_container_exempt), "parameters": {"tags": ["latest", "testing"], "allowedImages": ["exempt:latest", "exempt:testing"]}}
```
sozercan, line 106:
```suggestion
    input := { "review": input_init_review(input_container_exempt), "parameters": {"tags": ["latest", "testing"], "exemptImages": ["exempt:latest", "exempt:testing"]}}
```

sozercan (changes requested):

src/general/disallowedtags/src_test.rego
becky-hd: done.

sozercan (approved):
LGTM

library/general/disallowedtags/suite.yaml
```diff
@@ -11,6 +11,10 @@ tests:
     object: samples/container-image-must-not-have-latest-tag/example_allowed.yaml
     assertions:
     - violations: no
+  - name: exempt-allowed
+    object: samples/container-image-must-not-have-latest-tag/example_allowed_excluded.yaml
+    assertions:
+    - violations: no
   - name: example-disallowed
```
julianKatz, line 18: better names for these cases would make the suite.yaml much more helpful.  Right now, I have no way to know what's being tested other than to go look at the objects and grok the template.  Names that say _what_ we're testing would be helpful.

becky-hd: Renamed test names and objects.

julianKatz (approved):

maxsmythe (approved):
LGTM

sozercan:
@becky-hd looks like there's a conflict now

maxsmythe:
looks like `make generate` needs to be run to generate the templates.

becky-hd:
> looks like `make generate` needs to be run to generate the templates.

Thanks for pointing that out, I updated its `constraint.tmpl`.

PR #158, opened 2021-12-15
add reorder parameter to example kustomize command

wkrause13:
If you run the example kustomize build and apply command, the constrainttemplate is created after the constraint. The first time this is run, it results in the error below: 

```
constrainttemplate.templates.gatekeeper.sh/k8sreplicalimits created
Error from server (NotFound): error when creating "STDIN": the server could not find the requested resource
```

Using the --reorder none parameter resolves this issue

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,7 +22,7 @@ resources:
 
 Then define your constraints in a file called `constraints.yaml` in the same directory. Example constraints can be found in the "samples" folders.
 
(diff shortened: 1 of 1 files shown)


maxsmythe:
Thanks for the PR!

Is the kustomize command including any constraints? It should only be referencing the library's constraint templates

PR #160, opened 2021-12-17
Create block-wildcard-ingress constraint

mac-chaffee:
Mentioned the reasoning here: https://github.com/kubernetes-sigs/multi-tenancy/issues/1524 , which I've copied below:

> Users can put wildcards (*) in their Ingresses, but I think this can lead to security issues in a multi-tenant environment: https://kubernetes.io/docs/concepts/services-networking/ingress/#hostname-wildcards

> Wildcards in hostnames mean that one tenant could create an ingress for "myservice.example.com" and another tenant could create an ingress for "\*.example.com" or just "\*" and wait. As long as the "myservice.example.com" ingress exists, there's no issue. But if that Ingress is ever removed, even for a second, the traffic will be sent to the other tenant's wildcard Ingress instead, where they can intercept credentials/secrets/etc.

> Additionally, if users leave the "host" field blank in their Ingress, it will automatically be converted into the wildcard "*", which means this situation may even happen by accident.

> Gatekeeper has an example policy which enforces unique Ingress hosts amongst tenants, but it doesn't handle wildcards: https://github.com/open-policy-agent/gatekeeper-library/blob/master/library/general/uniqueingresshost/template.yaml Maybe adding a benchmark here can help raise awareness of this risk.

> Additional reference: https://youtu.be/Bxnu3llBN20?t=538

This PR adds a constraint to check for wildcards in Ingress hostnames.

CAVEAT: If you create an Ingress with multiple hostnames where one of them is invalid but the others are valid, the error message will say ALL the hostnames are invalid. Any Rego experts know of a way around that? If not, It's probably rare and harmless enough to be fine IMO.

Also PS: `gator` is extremely cool! I had already written the ConstraintTemplate, but the total time it took for me to create the test case having ZERO knowledge of `gator` was like 20 minutes, which is so nice!

### DIFF
--- a/library/general/block-wildcard-ingress/suite.yaml
+++ b/library/general/block-wildcard-ingress/suite.yaml
@@ -11,7 +11,15 @@ tests:
     object: samples/block-wildcard-ingress/example_allowed.yaml
     assertions:
     - violations: 0
-  - name: example-disallowed
-    object: samples/block-wildcard-ingress/example_disallowed.yaml
+  - name: blank-host
+    object: samples/block-wildcard-ingress/disallowed/blank_host.yaml
     assertions:
-    - violations: 3
+    - violations: 1
+  - name: host-omitted
+    object: samples/block-wildcard-ingress/disallowed/host_omitted.yaml
+    assertions:
+    - violations: 1
+  - name: wildcard-host
+    object: samples/block-wildcard-ingress/disallowed/wildcard_host.yaml
+    assertions:
+    - violations: 1
--- a/library/general/block-wildcard-ingress/suite.yaml
+++ b/library/general/block-wildcard-ingress/suite.yaml
@@ -14,4 +14,4 @@ tests:
   - name: example-disallowed
     object: samples/block-wildcard-ingress/example_disallowed.yaml
     assertions:
-    - violations: 4
+    - violations: 3
--- a/library/general/block-wildcard-ingress/suite.yaml
+++ b/library/general/block-wildcard-ingress/suite.yaml
@@ -1,0 +1,17 @@
+kind: Suite
+apiVersion: test.gatekeeper.sh/v1alpha1
+metadata:
+  name: block-wildcard-ingress
+tests:
+- name: block-wildcard-ingress
+  template: template.yaml
+  constraint: samples/block-wildcard-ingress/constraint.yaml
+  cases:
+  - name: example-allowed
+    object: samples/block-wildcard-ingress/example_allowed.yaml
+    assertions:
+    - violations: 0
+  - name: example-disallowed
+    object: samples/block-wildcard-ingress/example_disallowed.yaml
+    assertions:
+    - violations: 4
--- a/library/general/block-wildcard-ingress/template.yaml
+++ b/library/general/block-wildcard-ingress/template.yaml
@@ -1,0 +1,36 @@
+apiVersion: templates.gatekeeper.sh/v1beta1
+kind: ConstraintTemplate
+metadata:
+  name: k8sblockwildcardingress
+  annotations:
+    description: >-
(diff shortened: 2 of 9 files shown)


library/general/block-wildcard-ingress/template.yaml
```diff
@@ -0,0 +1,33 @@
+apiVersion: templates.gatekeeper.sh/v1beta1
+kind: ConstraintTemplate
+metadata:
+  name: k8sblockwildcardingress
+  annotations:
+    description: >-
+      Users should not be able to create Ingresses with a blank or wildcard (*) hostname since that would enabled them to intercept traffic for other services in the cluster, even if they don't have access to those services.
```
ritazh, line 7: nit:

```suggestion
      Users should not be able to create Ingresses with a blank or wildcard (*) hostname since that would enable them to intercept traffic for other services in the cluster, even if they don't have access to those services.
```

library/general/block-wildcard-ingress/template.yaml
```diff
@@ -0,0 +1,33 @@
...
+          object.get(input.review.object.spec.rules[_], "host", "") == ""
+        }
+        is_wildcard = true {
+          contains(object.get(input.review.object.spec.rules[_], "host", ""), "*")
+        }
+
+
+        violation[{"msg": msg}] {
+          input.review.kind.kind == "Ingress"
+          hostname := object.get(input.review.object.spec.rules[_], "host", null)
```
ritazh, line 30: 👍

ritazh (approved):
Thanks for the PR @mac-chaffee!
LGTM

library/general/block-wildcard-ingress/template.yaml
```diff
@@ -0,0 +1,33 @@
...
+        is_wildcard = true {
+          object.get(input.review.object.spec.rules[_], "host", "") == ""
+        }
+        is_wildcard = true {
+          contains(object.get(input.review.object.spec.rules[_], "host", ""), "*")
+        }
+
+
+        violation[{"msg": msg}] {
+          input.review.kind.kind == "Ingress"
```
julianKatz, line 29: would DRY things up a bit to have a variable like:

```
rules := input.review.object.spec.rules[_]
```

This would be used in the `hostename :=` line and could be passed into your two `is_wildcard` functions as a parameter.  This makes the code a bit easier to follow than its current state, where `input.review` is effectively accessed as a global variable.

library/general/block-wildcard-ingress/suite.yaml
```diff
@@ -0,0 +1,17 @@
...
+tests:
+- name: block-wildcard-ingress
+  template: template.yaml
+  constraint: samples/block-wildcard-ingress/constraint.yaml
+  cases:
+  - name: example-allowed
+    object: samples/block-wildcard-ingress/example_allowed.yaml
+    assertions:
+    - violations: 0
+  - name: example-disallowed
```
julianKatz, line 14: Rather than a single `example_disallowed.yaml`, you can separate the negative test cases into three separate objects.  Each object can have a descriptive name like `host-name-omitted.yaml`, and could even be in a folder like `library/general/block-wildcard-ingress/samples/block-wildcard-ingress/disallowed/host-name-omitted.yaml`.

This would mean that if one of the edge cases you test was broken by a future change to the template, a test for that specific edge case would fail. It's also easier for users to understand the specific cases that are disallowed, since they are spelled out explicitly in the tests.  I had to go read your rego to understand what about each ingress rule was being tested.

library/general/block-wildcard-ingress/template.yaml
mac-chaffee: That ended up fixing the bug I mentioned where it was printing one message per host even when some of the hosts were valid :)

ritazh: Good catch!

julianKatz (approved):

PR #166, opened 2022-01-24
feat: automate changes from src.rego to template.yaml

### LINKED ISSUE #19: Duplicated code in template.yaml and src.rego
Right now the code in `src.rego` is duplicated in `template.yaml`. Of course we have to keep `src.rego` around for the tests to work. But it would be great if we could inject the code of `src.rego` into `.spec.targets[0].rego` of the `template.yaml` file.

I assume that's not possible with `kustomize`? Maybe [`yq`](https://github.com/mikefarah/yq)?

### LINKED ISSUE #27: Build script to insert src.rego into template.yaml files
Per the [tooling proposal doc](https://docs.google.com/document/d/1uQlkIBQcgNNyth8o9ufYaVSfUq_JLJGy9fcuW3VB7vU/), we should set up a build script to automatically insert/update the Rego source in template.yaml files from the original src.rego. (We will need to update the README.md with instructions to run this script after any Rego code updates.)

The script should ideally also insert a version annotation (e.g. the SHA256 hash of the Rego code) which would let users know if the template they are using is current with the repo.

### LINKED ISSUE #100: Automate changes from src.rego to template.yaml
This will make contribution and testing easier.

chewong:
Fixes #19
Fixes #27
Fixes #100 

This PR creates `./scripts/generate.sh` to automate changes from src.rego to template.yaml.

Some noticeable changes:
- Converted YAML multiline character from `>-` to `|` to preserve the multiline description when running the generate script.
- Converted tabs to spaces for some files.
- Added/removed trailing spaces and newline character for some files.
- Added `make generate` as a GitHub Actions CI job.

Credit to @hi-artem for the [poc](https://github.com/open-policy-agent/gatekeeper-library/issues/100#issuecomment-1018102989).


Makefile
```diff
@@ -5,8 +5,9 @@ KUSTOMIZE_VERSION ?= 3.7.0
 GATEKEEPER_VERSION ?= release-3.5
 BATS_VERSION ?= 1.3.0
 GATOR_VERSION ?= 3.7.0
+GOMPLATE_VERSION ?= 3.10.0
 
-integration-bootstrap:
+integration-bootstrap: yq
```
sozercan, line 10:
```suggestion
integration-bootstrap: yq gomplate
```

chewong: gomplate is not a prerequisite of integration test so I don't need we need to add it here.

sozercan (approved):
LGTM

scripts/generate.sh
```diff
@@ -0,0 +1,75 @@
+#!/usr/bin/env bash
```
maxsmythe, line 1: It looks like we would still need to manually copy the libs file over. Is that the case?

maxsmythe (commented):
Thanks for the PR!

I'm wondering if we should have a golang-template-formatted template.yaml next to the Rego in the `src` directory and use that to generate the output.

Having a clean separation between manual and generated output may simplify reviews by highlighting which changes are determined by the autogen pipeline and which are manual changes.

It should also simplify the script by removing the need to clobber parts of the YAML.

Having some solution for libs that doesn't involve copy/paste would be good, though the solution for that isn't obvious, so might be worth addressing in a later PR.

scripts/generate.sh
chewong: By manually copying the libs file, did you mean having duplicated `lib_*.rego` in `src/<policies-that-import-the-libs>?

chewong:
> Having a clean separation between manual and generated output may simplify reviews by highlighting which changes are determined by the autogen pipeline and which are manual changes.

Agreed. Let me revise the PR.

Makefile
```diff
@@ -42,3 +41,22 @@ test-gator-dockerized: __build-gator
 .PHONY: build-gator
 __build-gator:
 	docker build --build-arg GATOR_VERSION=$(GATOR_VERSION) -f build/gator/Dockerfile -t gator-container .
+
+.PHONY: yq
+yq:
+	# Download and install yq
+	sudo snap install yq
```
maxsmythe, line 48: Is snap available to people running this locally?

Also, it looks like yq may no longer be used?

scripts/generate_constraint_tmpl.sh
```diff
@@ -0,0 +1,69 @@
+#!/usr/bin/env bash
```
maxsmythe, line 1: It looks like the makefile is calling gomplate directly. Is this script still used?

maxsmythe (commented):
Thanks for rearranging things! I think we are close. I have a couple of cleanup comments.

scripts/generate_constraint_tmpl.sh
chewong: This is a script that helped me reverse generate constraint.tmpl from existing template.yaml. I don't think we need it anymore.

Makefile
chewong: I was simply moving this line from `make integration-bootstrap`. It also looks like the entire Makefile assumes that we are running in the GitHub Actions environment (e.g. the use of `GITHUB_WORKSPACE`). I can open a follow-up PR to make it more developer-friendly.

chewong: Yes, yq is no longer being used. Let me remove that.

sozercan (approved):

maxsmythe (approved):
LGTM

Thanks for doing this!

chewong:
Rebased. PTAL.

maxsmythe:
still LGTM

ritazh (approved):
Thanks @chewong for adding this! 🎉

Issue #79, opened 2021-05-13
seccomp policy doesn't take account of new format.


raesene:
The [seccomp policy](https://github.com/open-policy-agent/gatekeeper-library/blob/master/library/pod-security-policy/seccomp/template.yaml) checks for a valid seccomp profile being set in annotations for the workload. Since Kubernetes 1.19, seccomp policy settings have moved into the securityContext section of the manifest (docs [here](https://kubernetes.io/docs/tutorials/clusters/seccomp/#create-a-pod-with-a-seccomp-profile-for-syscall-auditing)).

If the current version of the policy is used, it may block workloads which should be allowed.

PR #167, opened 2022-01-25
Rework PSP seccomp policy to support the securityContext configuration option

alain-baxter:
Reworked the Rego policy code for k8spspseccomp so that it will support seccomp configuration found in the pods securityContext while still keeping support for the annotation based seccomp configuration that has been deprecated (to be removed in kubernetes v1.25).

This new policy code was tested with the original unit tests first to ensure backwards compatibility, however I have refactored and expanded the unit tests to cover all locations where a seccomp profile can be found.

The messaging to the user is improved, since there are now 4 locations where a profile can be detected for each container (pod annotation, container annotation, pod securityContext, and container securityContext) I made it clear which location the profile was detected in for each container. I also continue to return a violation when a container has no profile detected from any source as per the original policy, however this is a bit more explicit in the new policy code.

Gator tests were also confirmed to be passing with new messages, I have fixed one which was referencing the incorrect example file.

Fixes: https://github.com/open-policy-agent/gatekeeper-library/issues/79

Signed-off-by: Alain Baxter <alain.baxter@sada.com>

### DIFF
--- a/library/pod-security-policy/seccomp/template.yaml
+++ b/library/pod-security-policy/seccomp/template.yaml
@@ -206,7 +206,6 @@ spec:
         # Container profile as defined in pods securityContext
         get_profile(container) = {"profile": profile, "file": file, "location": location} {
         	not has_securitycontext_container(container)
-        	not has_annotation(get_container_annotation_key(container.name))
         	profile := input.review.object.spec.securityContext.seccompProfile.type
         	file := object.get(input.review.object.spec.securityContext.seccompProfile, "localhostProfile", "")
         	location := "pod securityContext"
--- a/library/pod-security-policy/seccomp/template.yaml
+++ b/library/pod-security-policy/seccomp/template.yaml
@@ -73,182 +73,182 @@ spec:
         pod_annotation_key = "seccomp.security.alpha.kubernetes.io/pod"
 
         naming_translation = {
-        	# securityContext -> annotation
-        	"RuntimeDefault": ["runtime/default", "docker/default"],
-        	"Unconfined": ["unconfined"],
-        	"Localhost": ["localhost"],
-        	# annotation -> securityContext
-        	"runtime/default": ["RuntimeDefault"],
-        	"docker/default": ["RuntimeDefault"],
-        	"unconfined": ["Unconfined"],
-        	"localhost": ["Localhost"],
+            # securityContext -> annotation
+            "RuntimeDefault": ["runtime/default", "docker/default"],
+            "Unconfined": ["unconfined"],
+            "Localhost": ["localhost"],
+            # annotation -> securityContext
+            "runtime/default": ["RuntimeDefault"],
+            "docker/default": ["RuntimeDefault"],
+            "unconfined": ["Unconfined"],
+            "localhost": ["Localhost"],
         }
 
         violation[{"msg": msg}] {
-        	not input_wildcard_allowed_profiles
-        	allowed_profiles := get_allowed_profiles
-        	container := input_containers[name]
-        	not is_exempt(container)
-        	result := get_profile(container)
-        	not allowed_profile(result.profile, result.file, allowed_profiles)
-        	msg := get_message(result.profile, result.file, name, result.location, allowed_profiles)
+            not input_wildcard_allowed_profiles
+            allowed_profiles := get_allowed_profiles
+            container := input_containers[name]
+            not is_exempt(container)
+            result := get_profile(container)
+            not allowed_profile(result.profile, result.file, allowed_profiles)
+            msg := get_message(result.profile, result.file, name, result.location, allowed_profiles)
         }
 
         get_message(profile, file, name, location, allowed_profiles) = message {
-        	not profile == "Localhost"
-        	message := sprintf("Seccomp profile '%v' is not allowed for container '%v'. Found at: %v. Allowed profiles: %v", [profile, name, location, allowed_profiles])
+            not profile == "Localhost"
+            message := sprintf("Seccomp profile '%v' is not allowed for container '%v'. Found at: %v. Allowed profiles: %v", [profile, name, location, allowed_profiles])
         }
 
         get_message(profile, file, name, location, allowed_profiles) = message {
-        	profile == "Localhost"
-        	message := sprintf("Seccomp profile '%v' with file '%v' is not allowed for container '%v'. Found at: %v. Allowed profiles: %v", [profile, file, name, location, allowed_profiles])
+            profile == "Localhost"
+            message := sprintf("Seccomp profile '%v' with file '%v' is not allowed for container '%v'. Found at: %v. Allowed profiles: %v", [profile, file, name, location, allowed_profiles])
         }
 
         input_wildcard_allowed_profiles {
-        	input.parameters.allowedProfiles[_] == "*"
+            input.parameters.allowedProfiles[_] == "*"
         }
 
         input_wildcard_allowed_files {
-        	input.parameters.allowedLocalhostFiles[_] == "*"
+            input.parameters.allowedLocalhostFiles[_] == "*"
         }
 
         input_wildcard_allowed_files {
-        	"localhost/*" == input.parameters.allowedProfiles[_]
+            "localhost/*" == input.parameters.allowedProfiles[_]
         }
 
         # Simple allowed Profiles
         allowed_profile(profile, file, allowed) {
-        	not startswith(lower(profile), "localhost")
-        	profile == allowed[_]
+            not startswith(lower(profile), "localhost")
+            profile == allowed[_]
         }
 
         # seccomp Localhost without wildcard
         allowed_profile(profile, file, allowed) {
-        	profile == "Localhost"
-        	not input_wildcard_allowed_files
-        	profile == allowed[_]
-        	allowed_files := {x | x := object.get(input.parameters, "allowedLocalhostFiles", [])[_]} | get_annotation_localhost_files
-        	file == allowed_files[_]
+            profile == "Localhost"
+            not input_wildcard_allowed_files
+            profile == allowed[_]
+            allowed_files := {x | x := object.get(input.parameters, "allowedLocalhostFiles", [])[_]} | get_annotation_localhost_files
+            file == allowed_files[_]
         }
 
         # seccomp Localhost with wildcard
         allowed_profile(profile, file, allowed) {
-        	profile == "Localhost"
-        	input_wildcard_allowed_files
-        	profile == allowed[_]
+            profile == "Localhost"
+            input_wildcard_allowed_files
+            profile == allowed[_]
         }
 
         # annotation localhost with wildcard
         allowed_profile(profile, file, allowed) {
-        	"localhost/*" == allowed[_]
-        	startswith(profile, "localhost/")
+            "localhost/*" == allowed[_]
+            startswith(profile, "localhost/")
         }
 
         # annotation localhost without wildcard
         allowed_profile(profile, file, allowed) {
-        	startswith(profile, "localhost/")
-        	profile == allowed[_]
+            startswith(profile, "localhost/")
+            profile == allowed[_]
         }
 
         # Localhost files from annotation scheme
         get_annotation_localhost_files[file] {
-        	profile := input.parameters.allowedProfiles[_]
-        	startswith(profile, "localhost/")
-        	file := replace(profile, "localhost/", "")
+            profile := input.parameters.allowedProfiles[_]
+            startswith(profile, "localhost/")
+            file := replace(profile, "localhost/", "")
         }
 
         # The profiles explicitly in the list
         get_allowed_profiles[allowed] {
-        	allowed := input.parameters.allowedProfiles[_]
+            allowed := input.parameters.allowedProfiles[_]
         }
 
         # The simply translated profiles
         get_allowed_profiles[allowed] {
-        	profile := input.parameters.allowedProfiles[_]
-        	not startswith(lower(profile), "localhost")
-        	allowed := naming_translation[profile][_]
+            profile := input.parameters.allowedProfiles[_]
+            not startswith(lower(profile), "localhost")
+            allowed := naming_translation[profile][_]
         }
 
         # Seccomp Localhost to annotation translation
         get_allowed_profiles[allowed] {
-        	profile := input.parameters.allowedProfiles[_]
-        	profile == "Localhost"
-        	file := object.get(input.parameters, "allowedLocalhostFiles", [])[_]
-        	allowed := sprintf("%v/%v", [naming_translation[profile][_], file])
+            profile := input.parameters.allowedProfiles[_]
+            profile == "Localhost"
+            file := object.get(input.parameters, "allowedLocalhostFiles", [])[_]
+            allowed := sprintf("%v/%v", [naming_translation[profile][_], file])
         }
 
         # Annotation localhost to Seccomp translation
         get_allowed_profiles[allowed] {
-        	profile := input.parameters.allowedProfiles[_]
-        	startswith(profile, "localhost")
-        	allowed := naming_translation.localhost[_]
+            profile := input.parameters.allowedProfiles[_]
+            startswith(profile, "localhost")
+            allowed := naming_translation.localhost[_]
         }
 
         # Container profile as defined in pod annotation
         get_profile(container) = {"profile": profile, "file": "", "location": location} {
-        	not has_securitycontext_container(container)
-        	not has_annotation(get_container_annotation_key(container.name))
-        	not has_securitycontext_pod
-        	profile := input.review.object.metadata.annotations[pod_annotation_key]
-        	location := sprintf("annotation %v", [pod_annotation_key])
+            not has_securitycontext_container(container)
+            not has_annotation(get_container_annotation_key(container.name))
+            not has_securitycontext_pod
+            profile := input.review.object.metadata.annotations[pod_annotation_key]
+            location := sprintf("annotation %v", [pod_annotation_key])
         }
 
         # Container profile as defined in container annotation
         get_profile(container) = {"profile": profile, "file": "", "location": location} {
-        	not has_securitycontext_container(container)
-        	not has_securitycontext_pod
-        	container_annotation := get_container_annotation_key(container.name)
-        	has_annotation(container_annotation)
-        	profile := input.review.object.metadata.annotations[container_annotation]
-        	location := sprintf("annotation %v", [container_annotation])
+            not has_securitycontext_container(container)
+            not has_securitycontext_pod
+            container_annotation := get_container_annotation_key(container.name)
+            has_annotation(container_annotation)
+            profile := input.review.object.metadata.annotations[container_annotation]
+            location := sprintf("annotation %v", [container_annotation])
         }
 
         # Container profile as defined in pods securityContext
         get_profile(container) = {"profile": profile, "file": file, "location": location} {
-        	not has_securitycontext_container(container)
-        	profile := input.review.object.spec.securityContext.seccompProfile.type
-        	file := object.get(input.review.object.spec.securityContext.seccompProfile, "localhostProfile", "")
-        	location := "pod securityContext"
+            not has_securitycontext_container(container)
+            profile := input.review.object.spec.securityContext.seccompProfile.type
+            file := object.get(input.review.object.spec.securityContext.seccompProfile, "localhostProfile", "")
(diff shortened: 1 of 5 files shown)


swarren83 (approved):
This change provides backwards compatibility for seccomp annotations while extending the policy to support seccompProfile.

library/pod-security-policy/seccomp/template.yaml
```diff
@@ -42,45 +42,89 @@ spec:
...
+        
+        violation[{"msg": msg}] {
+            not input_wildcard_allowed
+            container := input_containers[name]
             not is_exempt(container)
-            not input_container_allowed(metadata, container)
-            msg := sprintf("Seccomp profile is not allowed, pod: %v, container: %v, Allowed profiles: %v", [metadata.name, container.name, input.parameters.allowedProfiles])
+            result := get_profile(container)
+            not allowed_profile(result.profile)
+            msg := sprintf("Seccomp profile '%v' is not allowed for container '%v'. Found at: %v. Allowed profiles: %v", [result.profile, name, result.location, input.parameters.allowedProfiles])
```
maxsmythe, line 58: should we be defensive in case input.parameters.allowedProfiles is undefined? Use object.Get() and provide an empty default?

library/pod-security-policy/seccomp/template.yaml
```diff
@@ -42,45 +42,89 @@ spec:
...
+            profile == input.parameters.allowedProfiles[_]
         }
-
-        input_container_allowed(metadata, container) {
-          profile := get_container_profile(metadata, container)
-          profile == input.parameters.allowedProfiles[_]
+        
+        # Container profile as defined in pod annotation
+        get_profile(container) = {"profile": profile, "location": location} {
+            not has_annotation(get_container_annotation_key(container.name))
```
maxsmythe, line 71: What happens if both the annotation and the security context are defined?

maxsmythe (commented):
Thanks for the PR!

I think there might be an edge case where both annotations and config are provided.

library/pod-security-policy/seccomp/template.yaml
alain-baxter: input.parameters.allowedProfiles is a required parameter because this policy will fail closed. If it find a profile configured and its not in the list (which it won't if empty) then a violation is triggered and pod blocked, and if it does not find a profile configured it will also trigger a violation and pod blocked.

Do we have a mechanism for requiring a parameter?

library/pod-security-policy/seccomp/template.yaml
alain-baxter: I did some testing of these scenarios. Thankfully the `kube-apiserver` has protections for the case where the securityContext and annotation don't match by sending this error:
```
The Pod "nginx-seccomp-not-match" is invalid: spec.containers[0].securityContext.seccompProfile.type: Forbidden: seccomp type in annotation and field must match
```
Then I confirmed that the preference for container specific over pod default works even across the two options (i.e. container annotation is used over the pod default in securityContext). So I have added protections to the get_profile functions to ensure only one returns in all cases. I have added a preference to the securityContext version over the annotation only because annotations are deprecated but as stated before they must match before getting to the admission controller.

I've added unit test coverage for these.

alain-baxter:
Thanks for the review @maxsmythe I have pushed updates to address the case where annotations and securityContext are used together.

library/pod-security-policy/seccomp/template.yaml
maxsmythe: Should be possible to set minimum list length requirements in the JSONSchema of the template

maxsmythe (approved):
LGTM! Thanks for doing this!

Requiring a minimum list length in the constraint parameters would be a bonus, but is a slight scope creep.

src/pod-security-policy/seccomp/src.rego
```diff
@@ -2,40 +2,88 @@ package k8spspseccomp
...
+	not has_annotation(get_container_annotation_key(container.name))
+	profile := input.review.object.spec.securityContext.seccompProfile.type
+	location := "pod securityContext"
+}
+
+# Container profile as defined in containers securityContext
+get_profile(container) = {"profile": profile, "location": location} {
+	has_securitycontext_container(container)
+	profile := container.securityContext.seccompProfile.type
+	location := "container securityContext"
```
swarren83, line 56: `Integration test on Gatekeeper release-3.6` is failing here when creating [library/pod-security-policy/seccomp/samples/psp-seccomp/example_allowed.yaml](https://github.com/open-policy-agent/gatekeeper-library/blob/master/library/pod-security-policy/seccomp/samples/psp-seccomp/example_allowed.yaml).  This should not be happening because the yaml does not set `securityContext.seccompProfile`.  How do we debug the integration test to see what the object being presented to k8s looks like?

alain-baxter: I have investigated this issue, looks like in a live cluster, the `kube-apiserver` will translate the annotation to securityContext. So the example file:
```
apiVersion: v1
kind: Pod
metadata:
  name: nginx-seccomp-allowed
  annotations:
    container.seccomp.security.alpha.kubernetes.io/nginx: runtime/default
  labels:
    app: nginx-seccomp
spec:
  containers:
  - name: nginx
    image: nginx
```
results in the input with the following object (trimmed):
```
"object": {
      "apiVersion": "v1",
      "kind": "Pod",
      "metadata": {
        "annotations": {
          "container.seccomp.security.alpha.kubernetes.io/nginx": "runtime/default"
        },
        "creationTimestamp": "2022-02-04T15:12:26Z",
        "labels": {
          "app": "nginx-seccomp"
        },
        "name": "nginx-seccomp-allowed",
        "namespace": "policy-review-exporting",
        "uid": "d40d3dc3-fe9d-433f-beae-50d2b275c4f2"
      },
      "spec": {
        "containers": [
          {
            "image": "nginx",
            "imagePullPolicy": "Always",
            "name": "nginx",
            "resources": {},
            "securityContext": {
              "seccompProfile": {
                "type": "RuntimeDefault"
              }
            },
            "terminationMessagePath": "/dev/termination-log",
            "terminationMessagePolicy": "File",
            "volumeMounts": [
              {
                "mountPath": "/var/run/secrets/kubernetes.io/serviceaccount",
                "name": "kube-api-access-66kfc",
                "readOnly": true
              }
            ]
          }
        ],
        "dnsPolicy": "ClusterFirst",
        "enableServiceLinks": true,
        "preemptionPolicy": "PreemptLowerPriority",
        "priority": 0,
        "restartPolicy": "Always",
        "schedulerName": "default-scheduler",
        "securityContext": {},
        "serviceAccount": "default",
        "serviceAccountName": "default",
        "terminationGracePeriodSeconds": 30,
        "tolerations": [
          {
            "effect": "NoExecute",
            "key": "node.kubernetes.io/not-ready",
            "operator": "Exists",
            "tolerationSeconds": 300
          },
          {
            "effect": "NoExecute",
            "key": "node.kubernetes.io/unreachable",
            "operator": "Exists",
            "tolerationSeconds": 300
          }
        ],
        "volumes": [
          {
            "name": "kube-api-access-66kfc",
            "projected": {
              "defaultMode": 420,
              "sources": [
                {
                  "serviceAccountToken": {
                    "expirationSeconds": 3607,
                    "path": "token"
                  }
                },
                {
                  "configMap": {
                    "items": [
                      {
                        "key": "ca.crt",
                        "path": "ca.crt"
                      }
                    ],
                    "name": "kube-root-ca.crt"
                  }
                },
                {
                  "downwardAPI": {
                    "items": [
                      {
                        "fieldRef": {
                          "apiVersion": "v1",
                          "fieldPath": "metadata.namespace"
                        },
                        "path": "namespace"
                      }
                    ]
                  }
                }
              ]
            }
          }
        ]
      },
      "status": {
        "phase": "Pending",
        "qosClass": "BestEffort"
      }
    },
```

Notice that the container has the appropriate securityContext in the admission review.

I think for backwards compatibility I may have to prefer the annotation over the securityContext, what do you think @maxsmythe and @swarren83 ?

swarren83: Yes, I think the annotation needs to be evaluated first and then seccompProfile.

alain-baxter: I have run the integration tests manually successfully with this modification, I will push it to the branch.
```
$ make test-integration
bats -t test/bats/test.bats
1..8
ok 1 gatekeeper-controller-manager is running
ok 2 gatekeeper-audit is running
ok 3 namespace label webhook is serving
ok 4 constrainttemplates crd is established
ok 5 waiting for validating webhook
ok 6 applying sync config
ok 7 waiting for namespaces to be synced using metrics endpoint
ok 8 testing constraint templates
```

maxsmythe:
Do we know what kubernetes' plan is once the annotations are fully deprecated? Will it continue to perform this translation?

alain-baxter:
I'm not aware of the plans for when the annotations are fully deprecated.

@maxsmythe I've pushed to resolve merge conflicts with the new `make generate` functionality, I've also run `make generate` to ensure that the template is properly synced with the src rego now.

maxsmythe:
@ritazh @sozercan @shomron Do you know what k8s plans to do when deprecating the security context annotations? Would be bad if we defaulted to analyzing the annotations but they were ignored, as that would effectively nullify the check.

alain-baxter:
I've looked into it, and kubernetes actually syncs the securityContext and annotation both ways until v1.23. PR for this change is here: https://github.com/kubernetes/kubernetes/pull/91408/files 

So basically if it has securityContext but no annotation, it will build the annotation. If it has the annotation but no securityContext it will build the securityContext. And if it has both it makes sure they are the same (as I noted before). Then the kubelet only sources the profile value from the securityContext.

This translation will be removed with this issue: https://github.com/kubernetes/kubernetes/issues/95171

With the removal the annotation will do nothing, and the kubelet will continue to only use the securityContext profile. There is an issue to warn the user about this annotation deprecation using this mechanism https://github.com/kubernetes/enhancements/blob/master/keps/sig-api-machinery/1693-warnings/README.md , but I can't find confirmation that this work has been completed.

The difficulty for us is the constraints in the wild have the annotation values like `runtime/default`, `docker/default`, `unconfined`, or `localhost/*` as their configured `allowedProfiles`. Maybe even specific `localhost/profile.yaml`. So backwards compatibility is not simple, since the new values should be `RuntimeDefault`, `Unconfined`, or `Localhost` (in the Localhost case there is a separate field for the profile file that I have not made support for).

If we follow Kubernetes we would likely want to make a proper translation in Rego between the two schemes. But do we have a matrix for kubernetes versions supported by Gatekeeper? We may need to leave the translation in for much longer to allow the library to be used by older clusters. Then we should decide how to manage the two schemes, ideally we would support both so if someone configured their annotation with `docker/default` then we would allow this if either `docker/default` or `RuntimeDefault` are in the `allowedProfiles` list. This makes the Rego more complicated but still doable.

maxsmythe:
Unfortunately G8rs support matrix probably isn't relevant here, as the constraint templates are a separate entity (if the Rego is recognized, any template should work on any version of G8r)

I think this rewriting of states is reasonable. Is there any reason to only use the rewritten states for the securityContext field and not for the annotation?

alain-baxter:
Yeah I'll support both directions of translation that way constraints can be written with either naming scheme for the `allowedProfiles` list and it'll work for both securityContext and annotation. I will return to preferring the securityContext fields when searching for the profile name once translation is in place to mirror the kubelet's behaviour.

To better support the `Localhost` profile in the new scheme I'll be adding an optional list parameter to verify `securityContext.seccompProfile.localhostProfile` correctly. This will make sure both schemes have the same configuration options.

I'll let you know once this work is pushed to this PR.

alain-baxter:
@maxsmythe I've implemented the naming translation now. This allows for annotation and/or securityContext based naming to be used in the `allowedProfiles` list, and it will automatically translate between the to so that we can match either. Since the translation works both ways it also supports mixed naming schemes in allowedProfiles but its still recommended to use one or the other.

I have also implemented verification for the `localhostProfile` field in `securityContext`, this uses a new parameter for the new naming scheme, but does successfully match the json files from the annotation scheme `localhost/profile.json`

New unit tests implemented for the new functionality, also confirmed gator test and integration test.

library/pod-security-policy/seccomp/template.yaml
```diff
@@ -35,52 +35,220 @@ spec:
...
+        	profile == "Localhost"
+        	file := object.get(input.parameters, "allowedLocalhostFiles", [])[_]
+        	allowed := sprintf("%v/%v", [naming_translation[profile][_], file])
+        }
+
+        # Annotation localhost to Seccomp translation
+        get_allowed_profiles[allowed] {
+        	profile := input.parameters.allowedProfiles[_]
+        	startswith(profile, "localhost")
+        	allowed := naming_translation.localhost[_]
```
maxsmythe, line 184: scratch: what happens to the file name?

library/pod-security-policy/seccomp/template.yaml
```diff
@@ -35,52 +35,220 @@ spec:
...
-            value := metadata.annotations[key]
-            startswith(key, "container.seccomp.security.alpha.kubernetes.io/")
-            [prefix, name] := split(key, "/")
-            name == container.name
-            profile = value
+        # Localhost files from annotation scheme
+        get_annotation_localhost_files[file] {
+        	profile := input.parameters.allowedProfiles[_]
+        	startswith(profile, "localhost/")
+        	file := split(profile, "/")[1]
```
maxsmythe, line 157: to guard against file names in subdirectories (not sure if that's possible), do we want to trim the `localhost/` prefix instead?

library/pod-security-policy/seccomp/template.yaml
```diff
@@ -35,52 +35,220 @@ spec:
...
+        	not has_securitycontext_container(container)
+        	not has_annotation(get_container_annotation_key(container.name))
+        	not has_securitycontext_pod
+        	profile := input.review.object.metadata.annotations[pod_annotation_key]
+        	location := sprintf("annotation %v", [pod_annotation_key])
+        }
+
+        # Container profile as defined in container annotation
+        get_profile(container) = {"profile": profile, "file": "", "location": location} {
+        	not has_securitycontext_container(container)
```
maxsmythe, line 198: Do we also want to check for the existence of a pod security context?

In the world where the annotations and the contexts are kept in sync, there will always be a container security context if there is a container annotation.

In a world where the annotations are deprecated, the existence of a container annotation (which would be ignored), short-circuits considering a pod-level SC here.

library/pod-security-policy/seccomp/template.yaml
```diff
@@ -35,52 +35,220 @@ spec:
...
+        	container_annotation := get_container_annotation_key(container.name)
+        	has_annotation(container_annotation)
+        	profile := input.review.object.metadata.annotations[container_annotation]
+        	location := sprintf("annotation %v", [container_annotation])
+        }
+
+        # Container profile as defined in pods securityContext
+        get_profile(container) = {"profile": profile, "file": file, "location": location} {
+        	not has_securitycontext_container(container)
+        	not has_annotation(get_container_annotation_key(container.name))
```
maxsmythe, line 208: probably want to remove the container annotation check here for the same reason as why we want to add a pod SC check for the container annotation case above.

maxsmythe (commented):
Getting very close!

library/pod-security-policy/seccomp/template.yaml
alain-baxter: Yeah I think that would be better, I was thinking about this earlier as a potential problem if there can be another `/` in the path.

library/pod-security-policy/seccomp/template.yaml
alain-baxter: The file names are extracted in `get_annotation_localhost_files` when we verify the `localhostProfile` field.

library/pod-security-policy/seccomp/template.yaml
maxsmythe: @sozercan was this comment addressed?

library/pod-security-policy/seccomp/template.yaml
maxsmythe: Sorry, I forgot to delete this comment, I was keeping it around so I wouldn't forget the thought.

library/pod-security-policy/seccomp/template.yaml
maxsmythe: oops, thought this was merged into main for some reason

maxsmythe (changes requested):
un-LGTMing until the annotation comment is addressed

library/pod-security-policy/seccomp/template.yaml
alain-baxter: I took another look at the `get_profile` logic to understand what you mean. So right now the preference in each feature version (due to SC availability and version skew) is:

>1.19 Kubernetes (SC might as well be ignored since it is not available):
Container Annotation > Pod Annotation > Missing 

1.19 to 1.23 Kubernetes:
Container SC ( = Container Annotation due to skew ) > Pod SC ( = Pod Annotation due to skew ) > Missing

> 1.23 Kubernetes:
Container SC > Container Annotation > Pod SC > Pod Annotation > Missing

So you are correct that > 1.23 is not right, and adding "NOT Pod SC" to "Container Annotation" would change the order to:

Container SC > Pod SC > Container Annotation > Pod Annotation > Missing

Which is closer (And will not effect the other two orders), however I have had confirmation from the Kubernetes team that the Container/Pod Annotations are allowed to exist > 1.23, so we would not properly detect when the profile is missing. What we really want is:

Container SC > Pod SC > Missing ( > Container/Pod Annotation )

However I don't think it is possible to get all 3 logically at the same time because "Missing ( > Container/Pod Annotation )" conflicts with the > 1.19 ordering. 

In the short term adding "NOT Pod SC" will get > 1.23 closer to desired outcome safely, I think a future enhancement to Gatekeeper to add Kubernetes server version to input could help to fix missing detection.

Thanks for your diligent review.

library/pod-security-policy/seccomp/template.yaml
alain-baxter: This has been done, thanks

library/pod-security-policy/seccomp/template.yaml
maxsmythe: Thank you for putting up with the diligent review! ;)

Hard to say what we should do for a fully post-annotation world. Worst case we can add a parameter to disable annotation-based checks, or maybe create a V2 version of the CT. Best case, that check can be removed.

maxsmythe (approved):
LGTM

library/pod-security-policy/seccomp/template.yaml
alain-baxter: There are a couple options, also I just looked again and it's version 1.25 where annotations are officially dropped (some of the stories I combed through still had the original planned version of 1.23). I think as we move forward with some of our ideas for the library I'm happy to revisit this policy within the new framework and how to make sure it works 100% for all cases post 1.25

sozercan (approved):
minor nit, otherwise LGTM

library/pod-security-policy/seccomp/template.yaml
```diff
@@ -35,52 +35,220 @@ spec:
...
-            container := input_containers[_]
-            not is_exempt(container)
-            not input_container_allowed(metadata, container)
-            msg := sprintf("Seccomp profile is not allowed, pod: %v, container: %v, Allowed profiles: %v", [metadata.name, container.name, input.parameters.allowedProfiles])
+        container_annotation_key_prefix = "container.seccomp.security.alpha.kubernetes.io/"
+
+        pod_annotation_key = "seccomp.security.alpha.kubernetes.io/pod"
+
+        naming_translation = {
+        	# securityContext -> annotation
```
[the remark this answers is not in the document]
alain-baxter: The src rego has been formatted with `opa fmt` which makes it tabs. I personally prefer spaces for indentation but wanted to keep with the format provided by the tooling. I guess the `make generate` uses spaces by default for indentation in the YAML file.

Do you prefer that I use spaces in the src rego instead of using the format output by `opa fmt` @sozercan?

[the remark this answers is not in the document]
sozercan: yea just so we align with rest of the templates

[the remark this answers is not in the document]
alain-baxter: Pushed tabs -> spaces

sozercan (approved):

PR #168, opened 2022-02-02
Require and execute suite.yaml files

### LINKED ISSUE #164: Require existence of suite.yaml and run gator test as part of presubmit
We should require that new submissions have a suite.yaml living next to the template.

We should also run gator test as part of presubmit.

@willbeason

julianKatz:
This PR adds github actions to:

1. Require a `suite.yaml` file alongside every `template.yaml` file
2. Execute the `suite.yaml` files as a Github Action

Fixes #164


.github/workflows/workflow.yaml
```diff
@@ -53,3 +53,18 @@ jobs:
           name: logs
           path: |
             logs-*.json
+  require_suites:
```
willbeason, line 56: Any reason to have these as separate actions instead of one?

willbeason (approved):

.github/workflows/workflow.yaml
julianKatz: This way I can know what the problem is without having to look at the log output directly.  I'd just see the failed job and know what to go debug.

They also run in parallel, though the speedup likely isn't meaningful.

maxsmythe (approved):
LGTM

Thanks for doing this!

.github/workflows/workflow.yaml
willbeason: Sounds good to me!

ritazh (approved):
lgtm

pending rebase and address merge conflict

PR #171, opened 2022-02-02
Add gator tests to workflow

becky-hd:
Signed-off-by: Becky Huang <beckyhd@google.com>


maxsmythe (commented):
It looks like this overlaps with #168

PR #173, opened 2022-02-04
Add noupdateserviceaccount policy to library

mtaufen:
This policy helps prevent workloads with "update-self" permissions from
escalating further in the cluster by selecting a new service account to
run as. It is especially useful for workloads running in sensitive
namespaces like kube-system, where nearby service accounts are likely to
have cluster admin rights.

@maxsmythe

### DIFF
--- a/library/general/noupdateserviceaccount/samples/noupdateserviceaccount/constraint.yaml
+++ b/library/general/noupdateserviceaccount/samples/noupdateserviceaccount/constraint.yaml
@@ -1,0 +1,33 @@
+# IMPORTANT: Before deploying this policy, make sure you allow-list any groups
+# or users that need to deploy workloads to kube-system, such as cluster-
+# lifecycle controllers, addon managers, etc. Such controllers may need to
+# update service account names during automated rollouts (e.g. of refactored
+# configurations). You can allow-list them with the allowedGroups and
+# allowedUsers properties of the NoUpdateServiceAccount Constraint.
+apiVersion: constraints.gatekeeper.sh/v1beta1
+kind: NoUpdateServiceAccount
+metadata:
+  name: no-update-kube-system-service-account
+spec:
+  match:
+    namespaces: ["kube-system"]
+    kinds:
+    - apiGroups: [""]
+      kinds:
+      # You can optionally add "Pod" here, but it is unnecessary because
+      # Pod service account immutability is enforced by the Kubernetes API.
+      - "ReplicationController"
+    - apiGroups: ["apps"]
+      kinds:
+      - "ReplicaSet"
+      - "Deployment"
+      - "StatefulSet"
+      - "DaemonSet"
+    - apiGroups: ["batch"]
+      kinds:
+      # You can optionally add "Job" here, but it is unnecessary because
+      # Job service account immutability is enforced by the Kubernetes API.
+      - "CronJob"
+  parameters:
+    allowedGroups: []
+    allowedUsers: []
--- a/library/general/noupdateserviceaccount/samples/noupdateserviceaccount/example_allowed.yaml
+++ b/library/general/noupdateserviceaccount/samples/noupdateserviceaccount/example_allowed.yaml
@@ -1,0 +1,31 @@
+# Note: The gator tests currently require exactly one object per example file.
+# Since this is an update-triggered policy, at least two objects are technically
+# required to demonstrate it. Due to the gator requirement, we only have one
+# object below. The policy should allow changing everything but the
+# serviceAccountName field.
+kind: Deployment
+apiVersion: apps/v1
+metadata:
+  name: policy-test
+  namespace: kube-system
+  labels:
+    app: policy-test
+spec:
+  replicas: 1
+  selector:
+    matchLabels:
+      app: policy-test-deploy
+  template:
+    metadata:
+      labels:
+        app: policy-test-deploy
+    spec:
+      # Changing anything except this field should be allowed by the policy.
+      serviceAccountName: policy-test-sa-1
+      containers:
+      - name: policy-test
+        image: ubuntu
+        command:
+        - /bin/bash
+        - -c
+        - sleep 99999
--- a/library/general/noupdateserviceaccount/suite.yaml
+++ b/library/general/noupdateserviceaccount/suite.yaml
@@ -1,0 +1,20 @@
+kind: Suite
+apiVersion: test.gatekeeper.sh/v1alpha1
+metadata:
+  name: noupdateserviceaccount
+tests:
+  - name: noupdateserviceaccount
+    template: template.yaml
+    constraint: samples/noupdateserviceaccount/constraint.yaml
+    cases:
+      - name: example-allowed
+        object: samples/noupdateserviceaccount/example_allowed.yaml
+        assertions:
+          - violations: no
+      # Since the tests are currently stateless, they can't exercise level-triggered
+      # events like changing a field. For now, manually test example_disallowed.yaml
+      # instead.
+      # - name: example-disallowed
+      #   object: samples/noupdateserviceaccount/example_disallowed.yaml
+      #   assertions:
+      #     - violations: yes
--- a/library/general/noupdateserviceaccount/template.yaml
+++ b/library/general/noupdateserviceaccount/template.yaml
@@ -1,0 +1,99 @@
+apiVersion: templates.gatekeeper.sh/v1beta1
+kind: ConstraintTemplate
+metadata:
+  name: noupdateserviceaccount
+  annotations:
+    description: "Blocks updating the service account on resources that abstract over Pods. This policy is ignored in audit mode."
+spec:
+  crd:
+    spec:
+      names:
+        kind: NoUpdateServiceAccount
+      validation:
+        openAPIV3Schema:
+          properties:
+            allowedGroups:
+              description: Groups that should be allowed to bypass the policy.
+              type: array
+              items: string
+            allowedUsers:
+              description: Users that should be allowed to bypass the policy.
+              type: array
+              items: string
+  targets:
+  - target: admission.k8s.gatekeeper.sh
+    rego: |
+      package noupdateserviceaccount
+
+      privileged(userInfo, allowedUsers, allowedGroups) {
+        # Allow if the user is in allowedUsers.
+        # Use object.get so omitted parameters can't cause policy bypass by
+        # evaluating to undefined.
+        username := object.get(userInfo, "username", "")
+        allowedUsers[_] == username
+      } {
+        # Allow if the user's groups intersect allowedGroups.
+        # Use object.get so omitted parameters can't cause policy bypass by
+        # evaluating to undefined.
+        userGroups := object.get(userInfo, "groups", [])
+        groups := {g | g := userGroups[_]}
+        allowed := {g | g := allowedGroups[_]}
+        intersection := groups & allowed
+        count(intersection) > 0
+      }
+
+      get_service_account(obj) = spec {
+        obj.kind == "Pod"
+        spec := obj.spec.serviceAccountName
+      } {
+        obj.kind == "ReplicationController"
+        spec := obj.spec.template.spec.serviceAccountName
+      } {
+        obj.kind == "ReplicaSet"
+        spec := obj.spec.template.spec.serviceAccountName
+      } {
+        obj.kind == "Deployment"
+        spec := obj.spec.template.spec.serviceAccountName
+      } {
+        obj.kind == "StatefulSet"
+        spec := obj.spec.template.spec.serviceAccountName
+      } {
+        obj.kind == "DaemonSet"
+        spec := obj.spec.template.spec.serviceAccountName
+      } {
+        obj.kind == "Job"
+        spec := obj.spec.template.spec.serviceAccountName
+      } {
+        obj.kind == "CronJob"
+        spec := obj.spec.jobTemplate.spec.template.spec.serviceAccountName
+      }
+
+      violation[{"msg": msg}] {
+        # This policy only applies to updates of existing resources.
+        input.review.operation == "UPDATE"
+
+        # Use object.get so omitted parameters can't cause policy bypass by
+        # evaluating to undefined.
+        params := object.get(input, "parameters", {})
+        allowedUsers := object.get(params, "allowedUsers", [])
+        allowedGroups := object.get(params, "allowedGroups", [])
+
+        # Extract the service account.
+        oldKSA := get_service_account(input.review.oldObject)
+        newKSA := get_service_account(input.review.object)
+
+        # Deny unprivileged users and groups from changing serviceAccountName.
+        not privileged(input.review.userInfo, allowedUsers, allowedGroups)
+        oldKSA != newKSA
+        msg := "user does not have permission to modify serviceAccountName"
+      } {
+        # Defensively require object to have a serviceAccountName.
+        input.review.operation == "UPDATE"
+        not get_service_account(input.review.object)
+        msg := "missing serviceAccountName field in object under review"
+      } {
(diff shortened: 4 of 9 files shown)


library/general/noupdateserviceaccount/samples/noupdateserviceaccount/example_allowed.yaml
```diff
@@ -0,0 +1,65 @@
+kind: ServiceAccount
```
mtaufen, line 1: I'm not really sure how these tests work. Do we have an emulated kubernetes API such that we can test mutations of existing resources, or is it a stateless test?

maxsmythe: Unfortunately the tests are essentially stateless (only CREATE is tested). For our local tests, we are looking at ways to provide a full K8s AdmissionReview object so that things like "only-on-update" can be appropriately tested.

Generally edge-based enforcement is discouraged if there is an alternative (only works at request time, does not work for audit), but because this is specifically testing changes to the service account it seems reasonable.

mtaufen: Got it - what's the best thing to do here? Just comment out the "disallowed" example from `suite.yaml`?

For level-triggered audits like this, you _might_ be able to extract past revisions of Deployments, etc. They have to be stored _somewhere_ since there's a rollback option: https://kubernetes.io/docs/concepts/workloads/controllers/deployment/#checking-rollout-history-of-a-deployment

maxsmythe: Yeah, commenting out for now seems like the best course.

re: level-triggered audits... interesting. I wonder where this data lives? Currently we are scraping the objects as they exist on the cluster.

maxsmythe (commented):
Thanks for the PR!

Code LGTM, but we'll also need the Rego source in the src/ directory, like so:

https://github.com/open-policy-agent/gatekeeper-library/tree/master/src/general/allowedrepos

It'll also be possible to cover the edge-based unit tests using rego tests.

ritazh:
@mtaufen Thanks for the PR! 
In addition to Rego source in the src/ directory, a recent automation change requires the creation of 
> A constraint template named `src/<policy-name>/constraint.tmpl` with a `description` annotation and the parameter structure.

e.g. https://github.com/open-policy-agent/gatekeeper-library/blob/master/src/general/allowedrepos/constraint.tmpl

PTAL https://github.com/open-policy-agent/gatekeeper-library#new-policy for more details.

@chewong should `make generate` fail if `constraint.tmpl` is missing for a policy in `src/`?

mtaufen:
@ritazh @chewong FYI. I ran `make generate` but it actually deleted the contents of a bunch of template files:

```
Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git restore <file>..." to discard changes in working directory)
        modified:   library/general/allowedrepos/template.yaml
        modified:   library/general/automount-serviceaccount-token/template.yaml
        modified:   library/general/block-endpoint-edit-default-role/template.yaml
        modified:   library/general/block-nodeport-services/template.yaml
        modified:   library/general/containerlimits/template.yaml
        modified:   library/general/containerresourceratios/template.yaml
        modified:   library/general/disallowanonymous/template.yaml
        modified:   library/general/disallowedtags/template.yaml
        modified:   library/general/externalip/template.yaml
        modified:   library/general/httpsonly/template.yaml
        modified:   library/general/imagedigests/template.yaml
        modified:   library/general/noupdateserviceaccount/template.yaml
        modified:   library/general/replicalimits/template.yaml
        modified:   library/general/requiredannotations/template.yaml
        modified:   library/general/requiredlabels/template.yaml
        modified:   library/general/requiredprobes/template.yaml
        modified:   library/general/uniqueingresshost/template.yaml
        modified:   library/general/uniqueserviceselector/template.yaml
        modified:   library/pod-security-policy/allow-privilege-escalation/template.yaml
        modified:   library/pod-security-policy/apparmor/template.yaml
        modified:   library/pod-security-policy/capabilities/template.yaml
        modified:   library/pod-security-policy/flexvolume-drivers/template.yaml
        modified:   library/pod-security-policy/forbidden-sysctls/template.yaml
        modified:   library/pod-security-policy/fsgroup/template.yaml
        modified:   library/pod-security-policy/host-filesystem/template.yaml
        modified:   library/pod-security-policy/host-namespaces/template.yaml
        modified:   library/pod-security-policy/host-network-ports/template.yaml
        modified:   library/pod-security-policy/privileged-containers/template.yaml
        modified:   library/pod-security-policy/proc-mount/template.yaml
        modified:   library/pod-security-policy/read-only-root-filesystem/template.yaml
        modified:   library/pod-security-policy/seccomp/template.yaml
        modified:   library/pod-security-policy/selinux/template.yaml
        modified:   library/pod-security-policy/users/template.yaml
        modified:   library/pod-security-policy/volumes/template.yaml
```

For example:

```
diff --git a/library/general/allowedrepos/template.yaml b/library/general/allowedrepos/template.yaml
index 23c0c7d..e69de29 100644
--- a/library/general/allowedrepos/template.yaml
+++ b/library/general/allowedrepos/template.yaml
@@ -1,40 +0,0 @@
-apiVersion: templates.gatekeeper.sh/v1beta1
-kind: ConstraintTemplate
-metadata:
-  name: k8sallowedrepos
-  annotations:
-    description: >-
-      Requires container images to begin with a string from the specified list.
-spec:
-  crd:
-    spec:
-      names:
-        kind: K8sAllowedRepos
-      validation:
-        # Schema for the `parameters` field
-        openAPIV3Schema:
-          type: object
-          properties:
-            repos:
-              description: The list of prefixes a container image is allowed to have.
-              type: array
-              items:
-                type: string
-  targets:
-    - target: admission.k8s.gatekeeper.sh
-      rego: |
-        package k8sallowedrepos
-
-        violation[{"msg": msg}] {
-          container := input.review.object.spec.containers[_]
-          satisfied := [good | repo = input.parameters.repos[_] ; good = startswith(container.image, repo)]
-          not any(satisfied)
-          msg := sprintf("container <%v> has an invalid image repo <%v>, allowed repos are %v", [container.name, container.image, input.parameters.repos])
-        }
-
-        violation[{"msg": msg}] {
-          container := input.review.object.spec.initContainers[_]
-          satisfied := [good | repo = input.parameters.repos[_] ; good = startswith(container.image, repo)]
-          not any(satisfied)
-          msg := sprintf("initContainer <%v> has an invalid image repo <%v>, allowed repos are %v", [container.name, container.image, input.parameters.repos])
-        }
```

chewong:
Did you install [`gomplate`](https://docs.gomplate.ca/installing/) to your PATH?

mtaufen:
> Did you install [gomplate](https://docs.gomplate.ca/installing/) to your PATH?

Thanks, that was it!

I added tests and pushed some updates PTAL @ritazh @maxsmythe @chewong 

Also I ended up using `strings.Indent 6` instead of `strings.Indent 8` (from the [example](https://github.com/open-policy-agent/gatekeeper-library/blob/master/src/general/allowedrepos/constraint.tmpl) you gave) in my `constraint.tmpl` since that matched the original yaml indentation I had. LMK if you prefer `strings.Indent 8`, I just didn't want to anger the yaml indentation gods by changing more than necessary.

library/general/noupdateserviceaccount/samples/noupdateserviceaccount/example_allowed.yaml
mtaufen: Ack, done.

src/general/noupdateserviceaccount/src.rego
```diff
@@ -0,0 +1,74 @@
...
+  # Defensively require object to have a serviceAccountName.
+  input.review.operation == "UPDATE"
+  not get_service_account(input.review.object)
+  msg := "missing serviceAccountName field in object under review"
+} {
+  # Defensively require oldObject to have a serviceAccountName.
+  input.review.operation == "UPDATE"
+  not get_service_account(input.review.oldObject)
+  msg := "missing serviceAccountName field in oldObject under review"
+}
```
maxsmythe, line 74: nit: newline

src/general/noupdateserviceaccount/src_test.rego
```diff
@@ -0,0 +1,320 @@
...
+    msg := "missing serviceAccountName field in oldObject under review"
+}
+
+missing_new_violation[{"msg": msg}] {
+    true
+    msg := "missing serviceAccountName field in object under review"
+}
+
+policy_violation[{"msg": msg}] {
+    true
```
maxsmythe, line 318: nit: `true` superfluous here and above

maxsmythe (approved):
LGTM! 2 small nits

src/general/noupdateserviceaccount/src.rego
mtaufen: Done/

src/general/noupdateserviceaccount/src_test.rego
mtaufen: Done.

library/general/noupdateserviceaccount/template.yaml
```diff
@@ -0,0 +1,99 @@
...
+  targets:
+  - target: admission.k8s.gatekeeper.sh
+    rego: |
+      package noupdateserviceaccount
+
+      privileged(userInfo, allowedUsers, allowedGroups) {
+        # Allow if the user is in allowedUsers.
+        # Use object.get so omitted parameters can't cause policy bypass by
+        # evaluating to undefined.
+        username := object.get(userInfo, "username", "")
```
ritazh, line 32: Since this rego looks for `username` from `userInfo`, it will definitely fail when in audit, but since it also only applies to `UPDATE` operations, which is also not available in audit, it's essentially a pass thru policy. We should document this policy does NOT work for audit.

library/general/noupdateserviceaccount/suite.yaml
```diff
@@ -0,0 +1,20 @@
...
+  - name: noupdateserviceaccount
+    template: template.yaml
+    constraint: samples/noupdateserviceaccount/constraint.yaml
+    cases:
+      - name: example-allowed
+        object: samples/noupdateserviceaccount/example_allowed.yaml
+        assertions:
+          - violations: no
+      # Since the tests are currently stateless, they can't exercise level-triggered
+      # events like changing a field. For now, manually test example_disallowed.yaml
```
ritazh, line 15: @willbeason Would this type of policy that requires `input.review.oldObject`  need to be tested with `gator test`?

maxsmythe: I think yes

willbeason: Yeah - we're considering allowing test authors to optionally specify a full AdmissionReview instead of an object so that they can specify all of the other fields

library/general/noupdateserviceaccount/template.yaml
willbeason: That's a really good point - it gives me a question unrelated to the PR so I'll ask in gatekeeper-maintainers.

mtaufen: Where is the best place to document that? Just in the docstrings?

library/general/noupdateserviceaccount/suite.yaml
mtaufen: Do I need to add something to the PR for that?

willbeason: Nope - this is a side design discussion for something completely different. No action necessary on your part!

willbeason: Filed https://github.com/open-policy-agent/gatekeeper/issues/1843

willbeason (approved):

library/general/noupdateserviceaccount/template.yaml
sozercan: we can add in annotation and readme in the `noupdateserviceaccount` folder perhaps

mtaufen: Thanks, added.

ritazh (approved):
lgtm

mtaufen:
Fixed that one, but now it wants the `example_*` files to each contain exactly one object. Should I just stick placeholder Deployments in each or something, since we can't really provide an example with only one object for this specific policy?

```
    --- FAIL: example-allowed	(0.066s)
        object file must contain exactly one object: "usr/local/google/home/mtaufen/proj/gatekeeper-library/library/general/noupdateserviceaccount/samples/noupdateserviceaccount/example_allowed.yaml" defines 4 objects

```

maxsmythe:
Yeah, placeholder sounds fine.

library/general/noupdateserviceaccount/samples/noupdateserviceaccount/constraint.yaml
```diff
@@ -0,0 +1,33 @@
...
+      - "Deployment"
+      - "StatefulSet"
+      - "DaemonSet"
+    - apiGroups: ["batch"]
+      kinds:
+      # You can optionally add "Job" here, but it is unnecessary because
+      # Job service account immutability is enforced by the Kubernetes API.
+      - "CronJob"
+  parameters:
+    allowedGroups: []
```
ritazh, line 32: 👍

mtaufen:
Any ideas why https://github.com/open-policy-agent/gatekeeper-library/runs/5146018047?check_suite_focus=true failed?

I saw this in the logs but looks unrelated to this PR:
```
# error: unable to recognize "library/general/automount-serviceaccount-token/samples/automount-serviceaccount-token/constraint.yaml": no matches for kind "K8sPSPAutomountServiceAccountTokenPod" in version "constraints.gatekeeper.sh/v1beta1"
```

maxsmythe:
That looks like for some reason a CRT didn't get created for the automount service account token CT. That could be due to timeouts or a bug in the CT (the joys of testing against an eventually consistent system).

library/general/noupdateserviceaccount/samples/noupdateserviceaccount/example_allowed.yaml
```diff
@@ -0,0 +1,31 @@
+# Note: The gator tests currently require exactly one object per example file.
```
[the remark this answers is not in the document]
willbeason: Yep - tracking in https://github.com/open-policy-agent/gatekeeper/issues/1843

ritazh:
```
# error: unable to recognize "library/general/automount-serviceaccount-token/samples/automount-
```
we can ignore retry errors like these

This is not a flake. seems the `noupdateserviceaccount` policy is failing the integration test with:
```
# deployment.apps "policy-test" deleted
# expected: denied the request
# actual: deployment.apps/policy-test created
```
which is caused by https://github.com/open-policy-agent/gatekeeper-library/blob/master/test/bats/test.bats#L102 since the integration test will look for `example-disallowed.yaml` and expect it to be blocked. But the rego will let it thru because it's not an update since the `example-allowed.yaml` policy-test resources have been [deleted](https://github.com/open-policy-agent/gatekeeper-library/blob/master/test/bats/test.bats#L86). I don't think the current integration test works for any referential constraints.

To unblock this PR, we could update test.bats to conditionally skip/delay deletion of allowed resources for certain policies OR remove the disallowed yaml for this policy. Do folks have any preference?

maxsmythe:
IMO let's remove disallowed, since the resource itself is not technically disallowed, only in the context of updating the allowed resource. Because we don't currently have a way to broadcast that, it makes the example misleading.

maxsmythe:
re: error line... oops for over-fixating on that, I wonder if we should add a "retrying due to error" context so people know it's a handled error.

mtaufen:
Sweet, deleted the file. Let's try again :)

PR #174, opened 2022-02-08
Library addition: containerrequests policy

bradthebuilder:
Depending on which school of thought one subscribes to, you can either be [for the inclusion of resource limits](https://cloud.google.com/blog/products/containers-kubernetes/kubernetes-best-practices-resource-requests-and-limits) on your k8s pods or [against them](https://github.com/robusta-dev/alert-explanations/wiki/CPUThrottlingHigh-%28Prometheus-Alert%29#why-you-dont-need-cpu-limits). This PR is meant to support the later philosophy.

I've basically copied and slightly tweaked the `containerlimits` policy to turn it into a `containerrequests` policy, which does what its name suggests: it requires pods to have resource _requests_ instead of limits.

`make generate` successfully generates the `template.yaml file`, but running unit tests via `test.sh` didn't work on my box on the already-present `containerlimits` policy; 

```
opa test src/general/allowedrepos/*.rego
PASS: 14/14
opa test src/general/automount-serviceaccount-token/*.rego
PASS: 3/3
opa test src/general/block-endpoint-edit-default-role/*.rego
PASS: 5/5
opa test src/general/block-nodeport-services/*.rego
PASS: 2/2
opa test src/general/containerlimits/*.rego
2 errors occurred:
src/general/containerlimits/lib_exempt_container.rego:3: rego_unsafe_var_error: var _ is unsafe
src/general/containerlimits/lib_exempt_container.rego:17: rego_unsafe_var_error: expression is unsafe
```
Because I've derived `containerrequests` from `containerlimits`, I'd think the same error might show itself somewhere in this new addition. I'm chalking up the error to be, perhaps, a difference in the `opa` or `gomplate` versions I'm using vs the versions used in CI. I'm hoping that the PR's CI will be able to provide some insight into the matter.

Happy to coordinate and polish/tweak this PR to make it worth maintaining. Thank you for your consideration.


maxsmythe:
Thanks for the PR! Not sure why you're getting unit test failures. It appears to work on the GitHub side. I'll look at this tomorrow.

bradthebuilder:
Thank you very much! That is indeed reassuring.

I signed the CNCF individual contributor agreement yesterday and emailed off the signed agreement, so I think/hope the final check is just waiting on that human-in-the-loop.

maxsmythe (approved):
LGTM!

The CLA should be removed. Lemme see if I can rerun the tests

maxsmythe:
DCO test failing

bradthebuilder:
Sorry about that. Amended the signoff

ritazh (approved):
lgtm

PR #175, opened 2022-02-08
chore: add additional checks when generating templates

chewong:
Addresses https://github.com/open-policy-agent/gatekeeper-library/pull/173#issuecomment-1031100760.

- Move commands in `make generate` to `./scripts/generate.sh`
- Add additional checks when generating templates:
  - make sure every folder in `library/*` has a corresponding folder in `src/*`
  - make sure `src.rego`, `src_test.rego` and `constraint.tmpl` exist in every folder in `src/*`
- Backfill block-wildcard-ingress policy to `src/` and add unit tests


scripts/generate.sh
```diff
@@ -0,0 +1,56 @@
+#!/bin/bash
+
+set -o errexit
+set -o nounset
+set -o pipefail
+
+if ! command -v gomplate &>/dev/null; then
+  echo "gomplate not found in PATH"
```
ritazh, line 8: Instead of asking users to install this, should we build an image like https://github.com/open-policy-agent/gatekeeper/blob/master/Makefile#L404
Then we can use it as part of `make generate`. WDYT?

chewong: Updated the PR.

maxsmythe (approved):
LGTM

PR #177, opened 2022-02-10
feat: add PodDisruptionBudget policy 

### LINKED ISSUE #52: PodDisruptionBudget check
An issue I've hit a few times that would be really nice to have a library for,
PodDisruptionBudgets with minAvailable: 1 and a Deployment with replicas = 1.

This really messes with node drains. Its a valid config for some clusters but in a multi-tenant environment, not so much.

chewong:
Fixes #52.
Addresses https://github.com/open-policy-agent/community/discussions/69.

Adding a policy for checking `minAvailable` and `maxUnavailable` values in PodDisruptionBudget to make sure they are within the acceptable range. This will prevent PodDisruptionBudget from blocking voluntary disruptions such as node draining (e.g. a pdb with `minAvailable: 2` and a deployment with 2 replicas will prevent API server from evicting pods during node draining)

Follow-up: `minAvailable` and `maxUnavailable` can be expressed in percentage but right now this policy can only handle integers.


library/general/poddisruptionbudget/template.yaml
```diff
@@ -0,0 +1,86 @@
...
+            "PodDisruptionBudget <%v> has maxUnavailable of 0, only positive integers are allowed for maxUnavailable",
+            [pdb.metadata.name],
+          )
+        }
+
+        violation[{"msg": msg}] {
+          input.review.kind.kind == "PodDisruptionBudget"
+          pdb := input.review.object
+
+          obj := data.inventory.namespace[pdb.metadata.namespace][_][_][_]
```
maxsmythe, line 38: We should probably scope down to only iterate over the appropriate kinds here to avoid unnecessary iterations. Though maybe scoping by namespace is sufficient?

maxsmythe (commented):
Code looks good, though I do have some questions about how this CT would behave in the wild.

There may be a performance improvement by reducing iteration over the cache, but hard to say since we are at least scoping iteration by namespace.

library/general/poddisruptionbudget/template.yaml
```diff
@@ -0,0 +1,86 @@
...
+          obj.spec.selector.matchLabels == pdb.spec.selector.matchLabels
+
+          not valid_pdb_min_available(obj, pdb)
+          msg := sprintf(
+            "%v <%v> has %v replica(s) but PodDisruptionBudget <%v> has minAvailable of %v, only positive integers less than %v are allowed for minAvailable",
+            [obj.kind, obj.metadata.name, obj.spec.replicas, pdb.metadata.name, pdb.spec.minAvailable, obj.spec.replicas],
+          )
+        }
+
+        violation[{"msg": msg}] {
```
maxsmythe, line 62: We probably don't need to check this for non PDBs, since this can be fully validated just looking at the PDB in isolation

library/general/poddisruptionbudget/sync.yaml
```diff
@@ -0,0 +1,23 @@
...
+        kind: "Deployment"
+      - group: "apps"
+        version: "v1"
+        kind: "ReplicaSet"
+      - group: "apps"
+        version: "v1"
+        kind: "StatefulSet"
+      - group: "policy"
+        version: "v1"
+        kind: "PodDisruptionBudget"
```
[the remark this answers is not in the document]
chewong: That makes sense. I agreed that it's a reasonable tradeoff for improved performance.

maxsmythe (approved):
LGTM

library/general/poddisruptionbudget/template.yaml
```diff
@@ -0,0 +1,68 @@
...
+
+        valid_pdb_min_available(obj, pdb) {
+          # default to -1 if minAvailable is not set so valid_pdb_min_available is always true for objects with >= 0 replicas
+          min_available := object.get(pdb.spec, "minAvailable", -1)
+          obj.spec.replicas > min_available
+        }
+
+        valid_pdb_max_unavailable(pdb) {
+          # default to 1 if maxUnavailable is not set so valid_pdb_max_unavailable always returns true
+          max_unavailable := object.get(pdb.spec, "maxUnavailable", 1)
```
ritazh, line 66: can you add some docs to why default is 1 when maxUnavailable is not set

ritazh (approved):
one nit comment, otherwise LGTM

ritazh (approved):
lgtm

PR #185, opened 2022-04-06
Enforce container resources without max/min values

olegy2008:
I've created this template to only enforce the presence of resources defined in the constraint without specifying max values like in `containerrequests` and `containerlimits`.
That's my first time using `rego` so would appreciate any suggestions to make it look and work better if possible.


olegy2008:
@maxsmythe @ritazh is there any chance to get your review/opinion about this?

maxsmythe:
Sorry for the slow response :/ I like the idea of this CT!

I haven't had time to look through the tests, but the overall code LGTM.

Any chance this could be updated to also support ephemeral containers, per #191 ?

olegy2008:
Hey @maxsmythe!
I checked [ephemeralContainers](https://kubernetes.io/docs/concepts/workloads/pods/ephemeral-containers/#what-is-an-ephemeral-container) docs and it says that setting `resources` for ephemeralContainers is disallowed and ephemeral containers use spare resources already allocated to the pod.

library/general/containerresources/suite.yaml
```diff
@@ -0,0 +1,69 @@
...
+tests:
+  - name: container-limits-and-requests
+    template: template.yaml
+    constraint: samples/container-must-have-limits-and-requests/constraint.yaml
+    cases:
+      - name: example-allowed1
+        object: samples/container-must-have-limits-and-requests/example_allowed1.yaml
+        assertions:
+          - violations: no
+      - name: example-disallowed1
```
maxsmythe, line 14: nit: I wonder if we could name these tests to explain what they are testing?

library/general/containerresources/template.yaml
```diff
@@ -0,0 +1,94 @@
...
+          properties:
+            exemptImages:
+              description: >-
+                Any container that uses an image that matches an entry in this list will be excluded
+                from enforcement. Prefix-matching can be signified with `*`. For example: `my-image-*`.
+
+                It is recommended that users use the fully-qualified Docker image name (e.g. start with a domain name)
+                in order to avoid unexpectedly exempting images from an untrusted repository.
+              type: array
+              items:
```
maxsmythe, line 28: Is it possible to add a regex check for the string so that users get an error if they type anything other than "cpu" or "memory"?

maxsmythe (approved):
Thank you for checking WRT the ephemeral container!

The code looks good, I have a couple nits WRT test names and parameter validation.

library/general/containerresources/suite.yaml
olegy2008: Renamed tests, but not sure if it's more readable/clear now, do you have any naming conventions?

library/general/containerresources/template.yaml
olegy2008: Added enum to validate the values.

library/general/containerresources/samples/container-must-have-cpu-requests-memory-limits-and-requests/only-memory-limits-defined-disallowed.yaml
```diff
@@ -0,0 +1,17 @@
+apiVersion: v1
+kind: Pod
+metadata:
+  name: opa-disallowed
+  labels:
+    owner: me.agilebank.demo
+spec:
+  containers:
+    - name: opa
+      image: openpolicyagent/opa:0.9.2
```
willbeason, line 10: Does this work on a more recent version of OPA? We're on version 0.39.0 now:

https://hub.docker.com/layers/opa/openpolicyagent/opa/0.39.0/images/sha256-e6ad6740318e7050de7da895eba5046c040ef11903190d445a4fd1aabe0d3b68?context=explore

library/general/containerresources/template.yaml
maxsmythe: thanks!

library/general/containerresources/suite.yaml
maxsmythe: No conventions as yet, since the test function is fairly new, but this is much more clear. Thank you for doing this!

maxsmythe:
LGTM

ritazh (approved):
LGTM

willbeason (approved):

PR #190, opened 2022-04-13
add imageException for some pod Spec polies

fseldow:
For volume related polices, if the violated volume is not mounted or mounted on excluded container, it will not report violation.
HostNamespace if all images in this pod are excluded, the pod will be view compliant.


library/pod-security-policy/host-namespaces/samples/psp-host-namespace/constraint.yaml
```diff
@@ -7,3 +7,5 @@ spec:
     kinds:
       - apiGroups: [""]
         kinds: ["Pod"]
+  parameters:
+    exemptImages: ["exception.azurecr.io/exception*"]
```
sozercan, line 11: i don't think we need to define these in constraints (especially if it can resolve to a real registry), users can add them if it's needed

fseldow: remove related changes now

PR #191, opened 2022-04-13
Add ephemeral containers to policies

### LINKED ISSUE #188: PSP Ephemeral Containers
Newer versions of Kubernetes have introduced the concept of [ephemeral containers](https://kubernetes.io/docs/concepts/workloads/pods/ephemeral-containers/) as a beta feature. This introduces yet another container spec `ephemeralContainers` on top of the existing `initContainers` and `containers`. We need to update all constraints and assigns to make sure that the new spec is covered.

phillebaba:
This change adds ephemeral containers to all policies that currently check the containers and init containers field. It is kind of a quick solution just to bring all the policies to a level where ephemeral containers do not introduce security concerns. 

I have on purpose skipped the required probes policy as it currently only looks at the containers field. I have also only updated one of the mutation samples.

One thing that comes to mind when looking at the different policies is that there are a bunch of different ways to solve this problem. A couple of rules have a method which returns the container data from each field, and others duplicate code at different levels. I feel like it would be good to find a standard. Right now it would be easy to make mistakes and not include one of the containers field. At the same time I can see a situation where one would want to require read only root file systems for the init containers and containers fields, but not the ephemeral containers fields. Might be better to solve these challenges in the future.

Fixes #188

### DIFF
--- a/library/general/containerlimits/template.yaml
+++ b/library/general/containerlimits/template.yaml
@@ -172,6 +172,8 @@ spec:
           general_violation[{"msg": msg, "field": "initContainers"}]
         }
 
+        # Ephemeral containers not checked as it is not possible to set field.
+
         general_violation[{"msg": msg, "field": field}] {
           container := input.review.object.spec[field][_]
           not is_exempt(container)
--- a/library/general/allowedrepos/samples/repo-must-be-openpolicyagent/disallowed_both.yaml
+++ b/library/general/allowedrepos/samples/repo-must-be-openpolicyagent/disallowed_both.yaml
@@ -17,3 +17,10 @@ spec:
         limits:
           cpu: "100m"
           memory: "30Mi"
+  ephemeralContainers:
+    - name: nginx
+      image: nginx
+      resources:
+        limits:
+          cpu: "100m"
+          memory: "30Mi"
--- a/library/general/allowedrepos/suite.yaml
+++ b/library/general/allowedrepos/suite.yaml
@@ -26,8 +26,10 @@ tests:
   - name: both-disallowed
     object: samples/repo-must-be-openpolicyagent/disallowed_both.yaml
     assertions:
-    - violations: 2
+    - violations: 3
     - message: initContainer
       violations: 1
     - message: container
+      violations: 1
+    - message: ephemeralContainer
       violations: 1
--- a/library/general/allowedrepos/template.yaml
+++ b/library/general/allowedrepos/template.yaml
@@ -38,3 +38,10 @@ spec:
           not any(satisfied)
           msg := sprintf("initContainer <%v> has an invalid image repo <%v>, allowed repos are %v", [container.name, container.image, input.parameters.repos])
         }
+
+        violation[{"msg": msg}] {
+          container := input.review.object.spec.ephemeralContainers[_]
+          satisfied := [good | repo = input.parameters.repos[_] ; good = startswith(container.image, repo)]
+          not any(satisfied)
+          msg := sprintf("ephemeralContainer <%v> has an invalid image repo <%v>, allowed repos are %v", [container.name, container.image, input.parameters.repos])
+        }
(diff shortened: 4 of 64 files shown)


phillebaba:
Currently having a look at the best way to implement tests for both init containers and ephemeral containers.

ritazh (approved):
LGTM
Thanks for the PR @phillebaba! 🎉

maxsmythe:
Indeed, TY for the PR! LMK when you think you have the tests sorted out.

Rego tests can make it easy to stamp tests out. Adding them to `suite.yaml` might have a more uniform experience if you're trying to test them all in bulk.

library/general/containerlimits/template.yaml
```diff
@@ -172,6 +172,10 @@ spec:
           general_violation[{"msg": msg, "field": "initContainers"}]
         }
 
+        violation[{"msg": msg}] {
+          general_violation[{"msg": msg, "field": "ephemeralContainers"}]
```
fseldow, line 176: may have one question
It is said `Pod resource allocations are immutable, so setting resources is disallowed.` in
https://kubernetes.io/docs/concepts/workloads/pods/ephemeral-containers/#what-is-an-ephemeral-container

So, do we need to add ephemeralContainer check in container limit policy?

fseldow, line 176: may have one question
It is said `Pod resource allocations are immutable, so setting resources is disallowed.` in
https://kubernetes.io/docs/concepts/workloads/pods/ephemeral-containers/#what-is-an-ephemeral-container

So, do we need to add ephemeralContainer check in container limit policy?

fseldow: so doe the policy contaienrrequests and containerresourceratios

fseldow: so doe the policy contaienrrequests and containerresourceratios

phillebaba: @fseldow good catch, you are right. It is not needed so will remove. Going to have a second look to see if there are more of these.

JAORMX (approved):

maxsmythe:
@phillebaba How's the testing looking?

maxsmythe (approved):
LGTM, thank you for doing this!

fseldow (approved):

phillebaba:
@maxsmythe sorry I got a bit carried away with the testing, but started facing issues as certain rules have multiple violations and other do not. Maybe something to look more into in the future. I added the easiest test cases that I could.

I just need to figure out why the integration tests are failing, do you have any insight into this?

mac-chaffee:
@phillebaba Looks like the kind cluster used in testing is using v1.21 where the `ephemeralContainers` field is just ignored, which is causing "disallowed" pods to still get created since the unknown field is stripped out before gatekeeper sees it.

I tested upgrading the kind cluster to v1.23 where `ephemeralContainers` are enabled, but that still doesn't work because `spec.ephemeralContainers: Forbidden: cannot be set on create`.

I think you just can't integration-test this feature until we refactor `test/bats/test.bats`. may want to leave that work for another PR.

ritazh:
To unblock this PR, perhaps remove the deployments with the ephemeral containers for now?

phillebaba:
The issue with removing the deployments using ephemeral containers is that it is part of the actual test. Could we not just update the kind version?

mac-chaffee:
@phillebaba even if you increase the kind version, the integration tests still fail because you can't "kubectl apply" a Pod with an ephemeralContainer in it: `spec.ephemeralContainers: Forbidden: cannot be set on create`.

maxsmythe:
To exempt these ephemeral objects from on-cluster testing, can we rename them to remove the `example_` prefix?

https://github.com/open-policy-agent/gatekeeper-library/blob/259ad1bdd8945d6df451d3fc2867109d76b854e2/test/bats/test.bats#L79-L107

Then they will only be subject to OPA and `gator verify` tests

ritazh:
@phillebaba are you able to make the suggested changes above and in #197 so we can get this PR merged soon? Please let us know how we can help.

Issue #178, opened 2022-02-14
Discuss: Policies support various parameter values for different containers


fseldow:
# Context
It is about one requirement that wants to set policy parameter value on container level. Original issue link is https://github.com/Azure/azure-policy/issues/890

Currently, I can provide a workaround for this issue that suggest cx to assign additional constraints on the except container (use label selector to target container). Then it might be one constraint one container.

Will be suitable to add one new map-structure (container-value) parameter just as the issue suggested? Or do you have any other idea about this?

# Below is Duplication of issue description:

## Details of the scenario you tried and the problem that is occurring
We have an application container that needs SYS_PTRACE capability to be able to send SIGHUP signal to another container process. One way to allow this is by defining allowedCapabilities = ["SYS_PTRACE"] parameter. But if we do this, it basically means any other application container will also be able to use SYS_PTRACE capability, which is not intended.
Other way is to exclude the app container that needs the capability by defining excludedContainers = ["app1"] as an example and not add anything to allowedCapabilities. This way no other app container can make use of SYS_PTRACE here. But that would allow app1 container to use any other capability (as its excluded from policy evaluation)!

 (__Comment:__ we have name based container exclusion in azure policy for now whereas gk-lib is using image based. I am introducing image base exception for now, but it takes no differences in this issue)

## Suggested solution to the issue
So I would like to suggest to have a parameter like allowedCapabilities as a map(/object) of array instead of plain array, where key will be container_name and value would be an array of capabilities allowed for that specific container.
So we can specify it something like below, where app1 and app2 are container names
__allowedCapabilities = { "app1" : ["SYS_PTRACE"] , "app2" : ["SYS_RAWIO"] }__

This way we can make sure that specific containers have the required capabilities they need.

Though I took ContainerAllowedCapabilities as an example, similar logic shall apply for few other policies like AllowedHostPaths.

__If policy is Guest Configuration - details about target node__

fseldow:
My personal opinion is not to add such map-structure parameter. Because it is sort of like the policy is one duplicated configuration of cluster. In my understand, policy is a standard that customer should modify the cluster configuration to meet it. Not to change the policy to fit the cluster configuration.

maxsmythe:
Allowing special parameters for specific containers via map is interesting. It does have a "priority" problem... if a container could be matched by more than one key in the map, which one takes precedence? Or apply all possible allow lists (this would default to the most restrictive set of allowed capabilities).

[...]

PR #193, opened 2022-04-26
Separation of capabilities for initContainers

tspearconquest:
This allows a separate set of capabilities to be defined for initContainers which may be more or less restrictive than the capability set for containers that are not initContainers, and thus facilitates situations where you need higher capabilities in an initContainer for the express purpose of doing pre-configuration in order to run a container for the main service with lesser capabilities.

Signed-off-by: Thomas Spear <tspear@conquestcyber.com>

### DIFF
--- a/library/pod-security-policy/capabilities/samples/capabilities-demo/constraint.yaml
+++ b/library/pod-security-policy/capabilities/samples/capabilities-demo/constraint.yaml
@@ -10,5 +10,7 @@ spec:
     namespaces:
       - "default"
   parameters:
+    initContainerAllowedCapabilities: []
+    initContainerRequiredDropCapabilities: []
     allowedCapabilities: ["something"]
     requiredDropCapabilities: ["must_drop"]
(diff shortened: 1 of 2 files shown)


JAORMX (approved):

maxsmythe:
I'm wondering if this could be solved more generically by something like what's mentioned in this issue?

https://github.com/open-policy-agent/gatekeeper-library/issues/178

tspearconquest:
Yes thats actually a better way to handle the issue, but I see there is no PR against it yet.

This way is less granular but does provides the necessary functionality immediately, while a PR against #178 can further tighten the policy to grant capabilities by specific container metadata, but may take some time to implement. :)

maxsmythe:
My main concern is that we want to maintain backwards compatibility, which means if we implemented this and then #178 later, then we'd either have to figure out how to support both paradigms simultaneously (which might lead to some awkward syntax) or rev the constraint template version (which would make it a higher-touch upgrade for consumers).

I was wondering if we wanted to morph this PR to accept that type of parameter schema.

tspearconquest:
I can't commit more time to that unfortunately, so I will withdraw this pull request. Thanks anyway.

maxsmythe:
No worries, thanks for the PR!

PR #194, opened 2022-05-11
Add all policies to kustomization.yaml and add test

mac-chaffee:
I noticed some of the policies were missing from kustomization.yaml. I added those missing policies and a test that will ensure the CI pipeline fails if someone forgets to update kustomization.yaml after adding/removing policies.

In order to detect missing entries in the kustomization.yaml, I had to convert our files into the format that kustomize spits out (meaning it adds the apiVersion/kind and removed the leading space in lists)


maxsmythe (approved):
LGTM

Thank you for doing this!

ritazh (approved):
Thank you for adding this @mac-chaffee! 🎉

PR #196, opened 2022-05-15
ci: bump GK and k8s versions

sozercan:


maxsmythe:
@sozercan KIND not starting it looks like

ritazh:
I think you need to upgrade all ingress resources to `networking.k8s.io/v1`.

maxsmythe:
This LGTM once test failures are addressed, which it sounds like #197 will do

PR #201, opened 2022-05-18
Remove ephemeral examples from CI

ritazh:
Signed-off-by: Rita Zhang <rita.z.zhang@gmail.com>


library/general/allowedrepos/samples/repo-must-be-openpolicyagent/example_disallowed_initcontainer.yaml
```diff
@@ -0,0 +1,23 @@
+apiVersion: v1
```
sozercan, line 1: do we need this still? does `both` scenario cover this too?

library/general/allowedrepos/suite.yaml
```diff
@@ -12,24 +12,24 @@ tests:
...
-    object: samples/repo-must-be-openpolicyagent/disallowed_both.yaml
+    object: samples/repo-must-be-openpolicyagent/example_disallowed_both.yaml #disallowed_all.yaml
     assertions:
-    - violations: 3
+    - violations: 2 #3
     - message: initContainer
       violations: 1
     - message: container
       violations: 1
-    - message: ephemeralContainer
```
sozercan, line 34: do gator suite files break CI too?

library/general/allowedrepos/samples/repo-must-be-openpolicyagent/example_disallowed_initcontainer.yaml
ritazh: We do need it. This test case has a positive case and a negative case where the initcontainer is not pulling allowed repo, whereas `both` has 2 violations for both init and container and `all` has 3 violations for init, container, and ephemeral.

library/general/allowedrepos/suite.yaml
maxsmythe: gator tests should be well-behaved

library/general/disallowedtags/suite.yaml
```diff
@@ -23,10 +23,10 @@ tests:
     object: samples/container-image-must-not-have-latest-tag/example_disallowed_tag.yaml
     assertions:
     - violations: yes
-  - name: single-disallowed-tag-ephemeral
-    object: samples/container-image-must-not-have-latest-tag/example_disallowed_tag_ephemeral.yaml
-    assertions:
-    - violations: yes
+  # - name: single-disallowed-tag-ephemeral
```
maxsmythe, line 26: Do we need to disable the suite tests? they should be unaffected by the issues live k8s tests have

ritazh:
@sozercan @maxsmythe addressed your comments. PTAL

sozercan (approved):
Thank you! LGTM

PR #211, opened 2022-07-21
Added policy to block LoadBalancer services

sstoyanovucsd:
In some environments we'd like the cluster customers to use the pre-provisioned Ingress Controllers to expose their workloads to the outside world.  This policy allows the cluster administrator to block the creation of LoadBalancer services, which can lead to provisioning costly cloud resources (LBs).  This policy along with the one for NodePorts, on which this one is based, also prevent customers from exposing their workload in an unexpected and potentially unmonitored way, which can present a security issue.

### DIFF
--- a/library/general/block-loadbalancer-services/kustomization.yaml
+++ b/library/general/block-loadbalancer-services/kustomization.yaml
@@ -1,0 +1,2 @@
+resources:
+  - template.yaml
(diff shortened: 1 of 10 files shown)


maxsmythe (changes requested):
Thanks for the PR! Can you include the Rego source and templated template to mirror the rest of the templates?

IMO no need for Rego tests given the suite test should cover things sufficiently (not sure if others will agree).

Here is a sample PR that shows what the templated template stuff looks like: #214

sstoyanovucsd:
Thanks for the pointers @maxsmythe.  Added the things you requested and also rego tests.

maxsmythe (approved):
LGTM, thanks for the PR!

ritazh (approved):
LGTM

Thanks for the PR!

PR #214, opened 2022-07-23
Add checks for storage class usage

kfox1111:
This change checks statefulsets and pvc's to see if a storage class is specified and valid. This enables problems to be detected at the time the objects are loaded into the cluster rather then much later. This requires a lot less cleanup to happen to fix the misconfiguration.

### DIFF
--- a/library/general/storageclass/README.md
+++ b/library/general/storageclass/README.md
@@ -11,6 +11,6 @@ pending, and requires deleting the StatefulSet and any PVCs it has created along
 with redeploying the workload in order to fix. Blocking it up front makes it
 much easier to fix before there is a mess to clean up.
 
-There is a known limitation with this constraint. It only functions properly
+**WARNING** This constraint only functions properly
 on gatekeeper version 3.9 or above.
 
--- a/library/general/storageclass/README.md
+++ b/library/general/storageclass/README.md
@@ -7,7 +7,7 @@ storage class at all is specified.
 This policy helps prevent workloads from getting stuck indefinitely waiting
 for a storage class to provision the persistent storage that will never 
 happen. This often causes users to get confused as to why their pods are stuck
-pending, and requires deleting the StatefulSet and any PVCs its created along
+pending, and requires deleting the StatefulSet and any PVCs it has created along
 with redeploying the workload in order to fix. Blocking it up front makes it
 much easier to fix before there is a mess to clean up.
 
--- a/library/general/storageclass/README.md
+++ b/library/general/storageclass/README.md
@@ -1,0 +1,16 @@
+# StorageClass
+
+The `StorageClass` constraint blocks the creation of PVCs or StatefulSets 
+where the specified storage class doesn't exist on the cluster, or that no
+storage class at all is specified.
+
+This policy helps prevent workloads from getting stuck indefinitely waiting
+for a storage class to provision the persistent storage that will never 
+happen. This often causes users to get confused as to why their pods are stuck
+pending, and requires deleting the StatefulSet and any PVCs its created along
+with redeploying the workload in order to fix. Blocking it up front makes it
+much easier to fix before there is a mess to clean up.
+
+There is a known limitation with this constraint. It only functions properly
+on gatekeeper version 3.9 or above.
+
--- a/src/general/storageclass/constraint.tmpl
+++ b/src/general/storageclass/constraint.tmpl
@@ -1,0 +1,25 @@
+apiVersion: templates.gatekeeper.sh/v1
+kind: ConstraintTemplate
+metadata:
+  name: k8sstorageclass
+  annotations:
+    description: >-
+      Requires storage classes to be specified when used. Currently only gatekeeper 3.9+ is ssupoprted.
+spec:
+  crd:
+    spec:
+      names:
+        kind: K8sStorageClass
+      validation:
+        openAPIV3Schema:
+          type: object
+          description: >-
+            Requires storage classes to be specified when used.
+          properties:
+            includeStorageClassesInMessage:
+              type: boolean
+              default: true
+  targets:
+    - target: admission.k8s.gatekeeper.sh
+      rego: |
+{{ file.Read "src/general/storageclass/src.rego" | strings.Indent 8 | strings.TrimSuffix "\n" }}
--- a/src/general/storageclass/constraint.tmpl
+++ b/src/general/storageclass/constraint.tmpl
@@ -4,7 +4,7 @@ metadata:
   name: k8sstorageclass
   annotations:
     description: >-
-      Requires storage classes to be specified when used. Currently only gatekeeper 3.9+ is ssupoprted.
+      Requires storage classes to be specified when used. Only gatekeeper 3.9+ is supported.
 spec:
   crd:
     spec:
--- a/src/general/storageclass/src.rego
+++ b/src/general/storageclass/src.rego
@@ -1,0 +1,88 @@
+package k8sstorageclass
+
+is_pvc(obj) {
+  obj.apiVersion == "v1"
+  obj.kind == "PersistentVolumeClaim"
+}
+
+is_statefulset(obj) {
+  obj.apiVersion == "apps/v1"
+  obj.kind == "StatefulSet"
+}
+
+violation[{"msg": msg}] {
+  count(data.inventory.cluster["storage.k8s.io/v1"]["StorageClass"]) == 0
+  msg := sprintf("StorageClasses not synced. Gatekeeper may be misconfigured. Please have a cluster-admin consult the documentation.", [])
+}
+
+storageclass_found(name) {
+  data.inventory.cluster["storage.k8s.io/v1"]["StorageClass"][name]
+}
+
+violation[{"msg": pvc_storageclass_badname_msg}] {
+  input.parameters.includeStorageClassesInMessage == true
+  is_pvc(input.review.object)
+  not storageclass_found(input.review.object.spec.storageClassName)
+}
+pvc_storageclass_badname_msg := sprintf("pvc did not specify a valid storage class name <%v>. Must be one of [%v]", args) {
+  input.parameters.includeStorageClassesInMessage
+  args := [
+    input.review.object.spec.storageClassName,
+    concat(", ", [n | data.inventory.cluster["storage.k8s.io/v1"]["StorageClass"][n]])
+  ]
+} else := sprintf(
+  "pvc did not specify a valid storage class name <%v>.",
+  [input.review.object.spec.storageClassName]
+)
+
+violation[{"msg": pvc_storageclass_noname_msg}] {
+  is_pvc(input.review.object)
+  not input.review.object.spec.storageClassName
+}
+pvc_storageclass_noname_msg := sprintf("pvc did not specify a storage class name. Must be one of [%v]", args) {
+  input.parameters.includeStorageClassesInMessage
+  args := [
+    concat(", ", [n | data.inventory.cluster["storage.k8s.io/v1"]["StorageClass"][n]])
+  ]
+} else := sprintf(
+  "pvc did not specify a storage class name.",
+  []
+)
+
+violation[{"msg": statefulset_vct_badname_msg(vct)}] {
+  is_statefulset(input.review.object)
+  vct := input.review.object.spec.volumeClaimTemplates[_]
+  not storageclass_found(vct.spec.storageClassName)
+}
+statefulset_vct_badname_msg(vct) := msg {
+  input.parameters.includeStorageClassesInMessage
+  msg := sprintf(
+      "statefulset did not specify a valid storage class name <%v>. Must be one of [%v]", [
+      vct.spec.storageClassName,
+      concat(", ", [n | data.inventory.cluster["storage.k8s.io/v1"]["StorageClass"][n]])
+  ])
+}
+statefulset_vct_badname_msg(vct) := msg {
+  not input.parameters.includeStorageClassesInMessage
+  msg := sprintf(
+    "statefulset did not specify a valid storage class name <%v>.", [
+      vct.spec.storageClassName
+  ])
+}
+
+violation[{"msg": statefulset_vct_noname_msg}] {
+  is_statefulset(input.review.object)
+  vct := input.review.object.spec.volumeClaimTemplates[_]
+  not vct.spec.storageClassName
+}
+statefulset_vct_noname_msg := sprintf("statefulset did not specify a storage class name. Must be one of [%v]", args) {
+  input.parameters.includeStorageClassesInMessage
+  args := [
+    concat(", ", [n | data.inventory.cluster["storage.k8s.io/v1"]["StorageClass"][n]])
+  ]
+} else := sprintf(
+  "statefulset did not specify a storage class name.",
+  []
+)
+
+#FIXME pod generic ephemeral might be good to validate some day too.
--- a/Makefile
+++ b/Makefile
@@ -1,10 +1,11 @@
+docker := docker #You can build with podman by doing: make docker=podman
 KIND_VERSION ?= 0.14.0
 # note: k8s version pinned since KIND image availability lags k8s releases
 KUBERNETES_VERSION ?= 1.24.0
 KUSTOMIZE_VERSION ?= 4.5.5
 GATEKEEPER_VERSION ?= release-3.8
 BATS_VERSION ?= 1.3.0
-GATOR_VERSION ?= 3.8.1
+GATOR_VERSION ?= 3.9.0
 GOMPLATE_VERSION ?= 3.10.0
 
 integration-bootstrap:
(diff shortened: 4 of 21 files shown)


src/general/storageclass/constraint.tmpl
```diff
@@ -0,0 +1,22 @@
...
+  crd:
+    spec:
+      names:
+        kind: K8sStorageClass
+      validation:
+        openAPIV3Schema:
+          type: object
+          description: >-
+            Requires storage classes to be specified when used.
+          properties:
```
maxsmythe, line 18: Probably best to remove this if empty

src/general/storageclass/src.rego
```diff
@@ -0,0 +1,48 @@
...
+}
+
+is_statefulset(obj) {
+  obj.apiVersion == "apps/v1"
+  obj.kind == "StatefulSet"
+}
+
+violation[{"msg": msg}] {
+  count(data.inventory.cluster["storage.k8s.io/v1"]["StorageClass"]) == 0
+  msg := sprintf("storageclasses not synced. Gatekeeper is misconfigured.", [])
```
maxsmythe, line 15: This could also be due in lag waiting for objects to sync, so worth mentioning that to avoid sending users down a bad path.

Are there any cases when StorageClass would be expected to not have any members?

maxsmythe (commented):
Thanks for the PR! 2 nits, but basically LGTM

src/general/storageclass/src.rego
```diff
@@ -0,0 +1,48 @@
...
+}
+
+storageclass_found(name) {
+  data.inventory.cluster["storage.k8s.io/v1"]["StorageClass"][name]
+}
+
+violation[{"msg": msg}] {
+  is_pvc(input.review.object)
+  not storageclass_found(input.review.object.spec.storageClassName)
+  msg := sprintf("pvc did not specify a valid storage class name <%v>. Must be one of [%v]", [input.review.object.spec.storageClassName, concat(", ", [n | data.inventory.cluster["storage.k8s.io/v1"]["StorageClass"][n]])])
```
ritazh, line 25: Could there be privacy and security concerns sharing this info since this is printing out information of all storageclass, a cluster-level resource with a user that may not have cluster-level access rbac?

src/general/storageclass/src.rego
kfox1111: /me shrugs

Its possible, but StorageClasses are highly likely to be static. so the vast majority of the time, no StorageClasses are seen, gateeper needs to be reconfigured.  But... maybe we lighten it to "may be misconfigured"?

src/general/storageclass/src.rego
kfox1111: I debated that... on my cluster, I don't allow them to be listed, just because the details could be listed too. but in this case, its just the names. so, I don't know.... I guess we could maybe add a parameter to allow this to be configured?

kfox1111: Made it conditional

src/general/storageclass/constraint.tmpl
kfox1111: Added a use.

kfox1111:
Squashed to fix the DCO issue.

maxsmythe (approved):
LGTM, thank you for the PR!

maxsmythe:
Though it looks like there is a Rego parsing error, un-LGTM-ing until fix:

https://github.com/open-policy-agent/gatekeeper-library/runs/7587956474?check_suite_focus=true#step:3:92

maxsmythe (changes requested):
un-LGTMing for error fix

kfox1111:
Ok. I believe this updated version will fix the testing problems.

maxsmythe:
Looks like an integration test error now:

```
# actual: Error from server (Forbidden): error when creating "library/general/storageclass/samples/storageclass/example_allowed_pvc.yaml": admission webhook "validation.gatekeeper.sh" denied the request: [storageclass] pvc did not specify a valid storage class name <somestorageclass>. Must be one of []
# cleaning...
```

Also, what change did you make to fix the old error, which was:

```
  adding template: invalid ConstraintTemplate: 1 error occurred: template:33: rego_parse_error: else keyword cannot be used on rule declared with := operator
	} else := sprintf(
	  ^
```

The rego source hasn't changed, but I can't see the rest of the diff is force pushed. It's a bit weird that that error would go away without changing the rego.

kfox1111:
> Looks like an integration test error now:
> 
> ```
> # actual: Error from server (Forbidden): error when creating "library/general/storageclass/samples/storageclass/example_allowed_pvc.yaml": admission webhook "validation.gatekeeper.sh" denied the request: [storageclass] pvc did not specify a valid storage class name <somestorageclass>. Must be one of []
> # cleaning...
> ```

My bad I think. when I split out library/general/storageclass/samples/storageclass/test_storageclass.yaml I forgot to remove some things from it. I just pushed a new version.

> Also, what change did you make to fix the old error, which was:
> 
> ```
>   adding template: invalid ConstraintTemplate: 1 error occurred: template:33: rego_parse_error: else keyword cannot be used on rule declared with := operator
> 	} else := sprintf(
> 	  ^
> ```
> 
> The rego source hasn't changed, but I can't see the rest of the diff is force pushed. It's a bit weird that that error would go away without changing the rego.

The fix was the patch to Makefile:
GATOR_VERSION ?= 3.9.0

The else form was suggested by [Anders Eknert](https://app.slack.com/team/UM50VAS8Z) on slack. But apparently its a newer opa feature and needs an opa engine new enough to support it. It looks like gator included the right version of opa at 3.9.0+.

maxsmythe:
Makes sense, thanks for the explanation!

maxsmythe:
Still getting an error:

```
# actual: Error from server (Forbidden): error when creating "library/general/storageclass/samples/storageclass/example_allowed_pvc.yaml": admission webhook "validation.gatekeeper.sh" denied the request: [storageclass] pvc did not specify a valid storage class name <somestorageclass>. Must be one of []
```

Looks like referential data hasn't been loaded?

kfox1111:
Ok, now I'm at a loss...
The test data is loaded here:
https://github.com/open-policy-agent/gatekeeper-library/blob/5f7fec75d048ae38df4182afbdbd3652093cc7e3/library/general/storageclass/suite.yaml#L10-L13

```
[kfox@zathras gatekeeper-library]$ make verify-gator-dockerized docker=podman
podman build --build-arg GATOR_VERSION=3.9.0 -f build/gator/Dockerfile -t gator-container .
STEP 1/9: FROM alpine
STEP 2/9: ARG GATOR_VERSION
--> Using cache eacef95bacd56ce9d2bf9faf7f82867a5fadb4694d6377f4ed030f717b0687cb
--> eacef95bacd
STEP 3/9: RUN apk --no-cache add curl tar libc6-compat
--> Using cache a0f93c76cb71b5b8bdc2ca307d1e6f48a6f5be4b357894506718b4f704fe104b
--> a0f93c76cb7
STEP 4/9: WORKDIR /bin
--> Using cache 377f59aaf6b827c5b4e5493d20ba17420b592d33d9f700478bcad3db4d826381
--> 377f59aaf6b
STEP 5/9: RUN curl -OL https://github.com/open-policy-agent/gatekeeper/releases/download/v${GATOR_VERSION}/gator-v${GATOR_VERSION}-linux-amd64.tar.gz
--> Using cache 706dc2a3f1bd949f8776647d3b2cd1264e7e05a45df5f8efdb16ce6c2f0a3183
--> 706dc2a3f1b
STEP 6/9: RUN tar xvf gator-v${GATOR_VERSION}-linux-amd64.tar.gz && rm gator-v${GATOR_VERSION}-linux-amd64.tar.gz
--> Using cache 294a7d82c306afbe3b58016679e7ccd219055a5e6b26cc10acfd1f8e859ec84a
--> 294a7d82c30
STEP 7/9: RUN mkdir /gatekeeper-library
--> Using cache 7d2694cd01312b36cd53166bc27800ecbd39808fe9b1e07c6fc1b5450797babd
--> 7d2694cd013
STEP 8/9: WORKDIR /gatekeeper-library
--> Using cache d40d29e5d19be84227a1c2ac1172ff1ffe5c76dfdec9e7f5ed8ab1c8711fa564
--> d40d29e5d19
STEP 9/9: ENTRYPOINT ["/bin/gator"]
--> Using cache 952c9d6ebb29b91a088d08cc09f89e15b5f76eb069565c47e90c3c2468af8959
COMMIT gator-container
--> 952c9d6ebb2
Successfully tagged localhost/gator-container:latest
952c9d6ebb29b91a088d08cc09f89e15b5f76eb069565c47e90c3c2468af8959
podman run -i -v /home/kfox/git/gatekeeper-library:/gatekeeper-library gator-container verify ./...
ok	gatekeeper-library/library/general/allowedrepos/suite.yaml	0.017s
ok	gatekeeper-library/library/general/automount-serviceaccount-token/suite.yaml	0.006s
ok	gatekeeper-library/library/general/block-endpoint-edit-default-role/suite.yaml	0.006s
ok	gatekeeper-library/library/general/block-nodeport-services/suite.yaml	0.003s
ok	gatekeeper-library/library/general/block-wildcard-ingress/suite.yaml	0.009s
ok	gatekeeper-library/library/general/containerlimits/suite.yaml	0.020s
ok	gatekeeper-library/library/general/containerrequests/suite.yaml	0.021s
ok	gatekeeper-library/library/general/containerresourceratios/suite.yaml	0.053s
ok	gatekeeper-library/library/general/containerresources/suite.yaml	0.053s
ok	gatekeeper-library/library/general/disallowanonymous/suite.yaml	0.006s
ok	gatekeeper-library/library/general/disallowedtags/suite.yaml	0.022s
ok	gatekeeper-library/library/general/externalip/suite.yaml	0.006s
ok	gatekeeper-library/library/general/httpsonly/suite.yaml	0.005s
ok	gatekeeper-library/library/general/imagedigests/suite.yaml	0.012s
ok	gatekeeper-library/library/general/noupdateserviceaccount/suite.yaml	0.007s
ok	gatekeeper-library/library/general/poddisruptionbudget/suite.yaml	0.017s
ok	gatekeeper-library/library/general/replicalimits/suite.yaml	0.007s
ok	gatekeeper-library/library/general/requiredannotations/suite.yaml	0.007s
ok	gatekeeper-library/library/general/requiredlabels/suite.yaml	0.008s
ok	gatekeeper-library/library/general/requiredprobes/suite.yaml	0.010s
ok	gatekeeper-library/library/general/storageclass/suite.yaml	0.025s
ok	gatekeeper-library/library/general/uniqueingresshost/suite.yaml	0.009s
ok	gatekeeper-library/library/general/uniqueserviceselector/suite.yaml	0.008s
ok	gatekeeper-library/library/pod-security-policy/allow-privilege-escalation/suite.yaml	0.010s
ok	gatekeeper-library/library/pod-security-policy/apparmor/suite.yaml	0.012s
ok	gatekeeper-library/library/pod-security-policy/capabilities/suite.yaml	0.018s
ok	gatekeeper-library/library/pod-security-policy/flexvolume-drivers/suite.yaml	0.006s
ok	gatekeeper-library/library/pod-security-policy/forbidden-sysctls/suite.yaml	0.006s
ok	gatekeeper-library/library/pod-security-policy/fsgroup/suite.yaml	0.008s
ok	gatekeeper-library/library/pod-security-policy/host-filesystem/suite.yaml	0.014s
ok	gatekeeper-library/library/pod-security-policy/host-namespaces/suite.yaml	0.005s
ok	gatekeeper-library/library/pod-security-policy/host-network-ports/suite.yaml	0.011s
ok	gatekeeper-library/library/pod-security-policy/privileged-containers/suite.yaml	0.010s
ok	gatekeeper-library/library/pod-security-policy/proc-mount/suite.yaml	0.013s
ok	gatekeeper-library/library/pod-security-policy/read-only-root-filesystem/suite.yaml	0.010s
ok	gatekeeper-library/library/pod-security-policy/seccomp/suite.yaml	0.035s
ok	gatekeeper-library/library/pod-security-policy/selinux/suite.yaml	0.013s
ok	gatekeeper-library/library/pod-security-policy/users/suite.yaml	0.025s
ok	gatekeeper-library/library/pod-security-policy/volumes/suite.yaml	0.005s
PASS
```

kfox1111:
Oh, I think I see. there are two different sets of tests under library... so there are actually 3 different ways things are being tested. at the rego level, with gator, and with gatekeeper. Its the gatekeeper ones that are still broken.

kfox1111:
Ok. I see three problems. The test code in question is here:
https://github.com/open-policy-agent/gatekeeper-library/blob/master/test/bats/test.bats#L104-L110

To fix problem one, I think I just need to rename the file to the filename format that the script is expecting.

The second issue I think is more difficult... The section that loads the inventory:
https://github.com/open-policy-agent/gatekeeper-library/blob/master/test/bats/test.bats#L104-L110
comes after the section loading the allowed tests:
https://github.com/open-policy-agent/gatekeeper-library/blob/master/test/bats/test.bats#L93-L102

So it seems like inventory may only work with disallowed tests, not allowed ones? Should the section that loads inventory be moved up?

Possible third issue: The tests check that there isn't a storage class defined in the pvc... If there exists a storage class in the test cluster already, that is marked as Default, there is an admission hook in the apiserver that slides in the storageClassName if it doesn't exist in the pvc automatically, before the gatekeeper gets the call. so the test will pass because it does have a valid storage class name even though the test is checking that the pvc doesn't have a name and it should fail. Not sure if we will hit that case in this environment.

kfox1111:
Took a stab at fixing the first two issues. Also noticed the test/bats/sync.yaml needed the storage class type added. Fixed that too.
Won't know if the potential issue three is an issue until allowed to be tested.

kfox1111:
This does look like the third issue. :/

kfox1111:
Totally not sure this is a good idea, but unannotating all storageclasses as default in the test script so the tests should work.

kfox1111:
One missing '{' :(

Fixed.

kfox1111:
The failure is ingress related. Unrelated to this changeset by the look at it. Should we retry?

maxsmythe:
retrying

kfox1111:
Dug through a bit this morning... The issue now, is that uniqueingresshost's inventories utilize the same hostname name in the deny inventories as the allowed rules. So they conflict now. :/

Will try and fix....

kfox1111:
Updated the broken test to use a different name.

kfox1111:
Ok. Last problem.... Seems like the library may not support less then gatekeeper 3.9. :/

What should we do about that?

ritazh:
>  Seems like the library may not support less then gatekeeper 3.9. :/

@kfox1111 which feature from 3.9 are you using?

kfox1111:
> @kfox1111 which feature from 3.9 are you using?

The "else := "  feature seems to be 3.9+. For example:
https://github.com/open-policy-agent/gatekeeper-library/blob/7f83740597dbc152174e81b7db453d4477a18cbe/src/general/storageclass/src.rego#L27-L36

That form was suggested by [Anders Eknert](https://app.slack.com/team/UM50VAS8Z) on slack

ritazh:
> The "else := " feature seems to be 3.9+.

Should we remove "else" for now until we no longer support 3.8?

kfox1111:
Would need to rewrite a bunch of code, and/or bring it back to what I originally proposed, where it always just printed out the storage class names...

How far out is 3.10? 

Another option would be to skip the template on 3.8 and mark the template as 3.9+ only.

ritazh:
> How far out is 3.10?

We have a 2 months release cadence. We have ~1.5 months left for 3.8 support.

kfox1111:
ok. How about we skip the 3rd set of tests until then?

ritazh:
> How about we skip the 3rd set of tests until then

To clarify, which tests are you referring to just to make sure we are talking about the same thing?

kfox1111:
The ones driven by bats. The opa / gator ones would still be enabled.

ritazh:
> The ones driven by bats. The opa / gator ones would still be enabled.

What do you think @sozercan @maxsmythe?

kfox1111:
Updated with the skip test code, to see if it even works.

sozercan:
Can we add a `README.md` explaining the scenario and minimum version of GK required in the storageclass folder?

src/general/storageclass/constraint.tmpl
```diff
@@ -0,0 +1,25 @@
+apiVersion: templates.gatekeeper.sh/v1
+kind: ConstraintTemplate
+metadata:
+  name: k8sstorageclass
+  annotations:
+    description: >-
+      Requires storage classes to be specified when used.
```
sozercan, line 7: can we add min version of GK required here too?

kfox1111: As a comment, or is there a different way to do that?

kfox1111: Oh. In the description? kk.

library/general/storageclass/README.md
```diff
@@ -0,0 +1,16 @@
+# StorageClass
+
+The `StorageClass` constraint blocks the creation of PVCs or StatefulSets 
+where the specified storage class doesn't exist on the cluster, or that no
+storage class at all is specified.
+
+This policy helps prevent workloads from getting stuck indefinitely waiting
+for a storage class to provision the persistent storage that will never 
+happen. This often causes users to get confused as to why their pods are stuck
+pending, and requires deleting the StatefulSet and any PVCs its created along
```
ritazh, line 10:
```suggestion
pending, and requires deleting the StatefulSet and any PVCs it has created along
```

library/general/storageclass/README.md
```diff
@@ -0,0 +1,16 @@
...
+with redeploying the workload in order to fix. Blocking it up front makes it
+much easier to fix before there is a mess to clean up.
+
+There is a known limitation with this constraint. It only functions properly
```
ritazh, line 14:
```suggestion
**WARNING** This constraint only functions properly
```

ritazh:
```
diff --git a/library/general/storageclass/template.yaml b/library/general/storageclass/template.yaml
index aec014b..c96b891 100644
--- a/library/general/storageclass/template.yaml
+++ b/library/general/storageclass/template.yaml
@@ -4,7 +4,7 @@ metadata:
   name: k8sstorageclass
   annotations:
     description: >-
-      Requires storage classes to be specified when used.
+      Requires storage classes to be specified when used. Currently only gatekeeper 3.9+ is ssupoprted.
 spec:
   crd:
     spec:
Please run 'make generate' to generate the templates
```

src/general/storageclass/constraint.tmpl
```diff
@@ -0,0 +1,25 @@
+apiVersion: templates.gatekeeper.sh/v1
+kind: ConstraintTemplate
+metadata:
+  name: k8sstorageclass
+  annotations:
+    description: >-
+      Requires storage classes to be specified when used. Currently only gatekeeper 3.9+ is ssupoprted.
```
ritazh, line 7:
```suggestion
      Requires storage classes to be specified when used. Only gatekeeper 3.9+ is supported.
```

kfox1111:
and... messed up my signoffs. :/

kfox1111:
ok, I think I fixed my broken signoffs.

ritazh (approved):
LGTM

ritazh:
@maxsmythe LGTY?

maxsmythe (approved):
LGTM

PR #219, opened 2022-08-18
feat: remove k8shttpsonly tls block requirement

apeabody:
* Add a new `tlsOptional` parameter, that when `true` makes the `tls {}` optional.
* Clarify metadata.name of example_allowed


maxsmythe:
Thanks for the PR!

Question about the `tls` block: if the block is defined but `kubernetes.io/ingress.allow-http` is not set, is that also sufficient for enforcing TLS?

apeabody:
> Thanks for the PR!
> 
> Question about the `tls` block: if the block is defined but `kubernetes.io/ingress.allow-http` is not set, is that also sufficient for enforcing TLS?

Hi Max - One major factor is per the docs: [`Note: There is a gap between TLS features supported by various Ingress controllers. Please refer to documentation on nginx, GCE, or any other platform specific Ingress controller to understand how TLS works in your environment.`](https://kubernetes.io/docs/concepts/services-networking/ingress/#tls)

My understanding from the docs is adding `tls: {}` will enable HTTPS/443, but generally will NOT disable HTTP/80.  Hence GCE's [`kubernetes.io/ingress.allow-http` ](https://cloud.google.com/kubernetes-engine/docs/concepts/ingress-xlb#disabling_http).  However looking at the [NGINX Ingress](https://kubernetes.github.io/ingress-nginx/user-guide/nginx-configuration/annotations/), it doesn't support the `kubernetes.io/ingress.allow-http` annotation, and instead of an integral method to disable HTTP/80, by default implements SSL-redirects when `tls: {}` is present.

sozercan:
I remember we got some question in the past about this policy, is `kubernetes.io/ingress.allow-http` annotation GCE/GKE only?

maxsmythe:
Hard to say, since I don't think these things are necessarily standardized. Looking at the commit history, it seems like checking the annotation was part of the initial commit. I do remember some kind of conversation around annotations, but not sure if it was about this template.

Looking around, it looks like there is at least one other project that uses this annotation: https://projectcontour.io/docs/v1.18.0/config/annotations/

And there is this VMWare doc that mentions the annotation:
https://docs.vmware.com/en/VMware-NSX-Container-Plugin/4.0/ncp-kubernetes/GUID-E03D6EE5-9C6C-457F-AD81-25CF2056F4D8.html

So might not be purely GCP.

WRT the change as currently written: IMO we should have a parameter that allows users to disable the TLS check in order to maintain backwards-compatibility for existing users.

apeabody:
Thanks for the feedback @maxsmythe and @sozercan.  I'll rework with an optional parameter for backwards-compatibility.  Cheers!

apeabody:
Hi @maxsmythe - When you have some time could you please take a look at the new version.  Thanks!

maxsmythe (approved):
Code LGTM.

Stylistically, the code would be simpler if it had separate violation{} rules for TLS and the annotation check (since then we wouldn't need the extra `tls_is_optional` function and would only need to evaluate `not input.parameters.tlsOptional` in the violation that checks the `tls` block.

But the code as-is is functional

ritazh (approved):
LGTM

PR #220, opened 2022-08-19
Fix storageclass no datasync test

astraldawn:
`results` before:

```
{{"msg": "pvc did not specify a valid storage class name <fast>. Must be one of []"}}
```

`results` after:

```
{{"msg": "StorageClasses not synced. Gatekeeper may be misconfigured. Please have a cluster-admin consult the documentation."}, {"msg": "pvc did not specify a valid storage class name <fast>. Must be one of []"}}
```

### DIFF
--- a/src/general/storageclass/src.rego
+++ b/src/general/storageclass/src.rego
@@ -11,7 +11,7 @@ is_statefulset(obj) {
 }
 
 violation[{"msg": msg}] {
-  count(data.inventory.cluster["storage.k8s.io/v1"]["StorageClass"]) == 0
+  not data.inventory.cluster["storage.k8s.io/v1"]["StorageClass"]
   msg := sprintf("StorageClasses not synced. Gatekeeper may be misconfigured. Please have a cluster-admin consult the documentation.", [])
 }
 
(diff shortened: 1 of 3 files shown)


maxsmythe:
Thanks for the PR!

Not from this PR, but it looks like line 23 may be a bug?

```
violation[{"msg": pvc_storageclass_badname_msg}] {
  input.parameters.includeStorageClassesInMessage == true
  is_pvc(input.review.object)
  not storageclass_found(input.review.object.spec.storageClassName)
}
```

reads like a violation will only be raised if `input.parameters.includeStorageClassesInMessage` is true, there are no other violations (other than the misconfiguration one), and that switch is already honored elsewhere.

Because this affects enforcement, would you mind also fixing that bug?

maxsmythe (approved):
LGTM, thank you for the follow-up fix!

src/general/storageclass/src.rego
```diff
@@ -20,7 +20,6 @@ storageclass_found(name) {
 }
 
 violation[{"msg": pvc_storageclass_badname_msg}] {
-  input.parameters.includeStorageClassesInMessage == true
```
ritazh, line 23: Please run 'make generate' to generate the constraint template file to include recent updates in src.rego.

PR #222, opened 2022-08-24
feat: adds Website for library

nilekhc:


library/general/automount-serviceaccount-token/template.yaml
```diff
@@ -3,6 +3,7 @@ kind: ConstraintTemplate
 metadata:
   name: k8spspautomountserviceaccounttokenpod
   annotations:
+    title: "Automount Service Account Token for Pod"
```
maxsmythe, line 6: Maybe we can have a more semantically (as opposed to usage-oriented) name for this field?

e.g. `prettyName` or similar.

Also, WRT annotations... should we be careful about using fully-qualified names to avoid collisions? And if we want to make these features standard, we may want to integrate these fields into the actual ConstraintTemplate schema.

scripts/website/generate-md.sh
```diff
@@ -0,0 +1,108 @@
...
+}
+
+getExamples() {
+    allDetailsExample=""
+    for i in $(ls -d ${1}samples/*/)
+    do
+        constraint=$(cat ${i}/constraint.yaml)
+        constraintExample="<details>\n<summary>constraint</summary>\n\n\`\`\`yaml\n$constraint\n\`\`\`\n\n</details>"
+
+        # find all files that contain the word "allowed" in the filename
```
maxsmythe, line 29: We should have one SOT for what purpose a given file serves. In this case the `suite.yaml` seems the better candidate to me since it allows for more flexible associations. For example, this `suite.yaml` makes it clear which referential data is associated with which tests:

https://github.com/open-policy-agent/gatekeeper-library/blob/6c3b8a75db8d7b3f0e10ea13d872eef9f3c69d48/library/general/uniqueingresshost/suite.yaml#L1-L25

And does so in a way that would allow for re-use vs. the naming-convention based paradigm.

Another example of edge cases we'll eventually need to support are policies that rely on request metadata, such as CREATE vs UPDATE or requesting user: https://github.com/open-policy-agent/gatekeeper/issues/1819

Right now there are ~2 sources of truth because some of the e2e test stuff still relies on naming conventions, but IMO new stuff should tend toward suite.yaml so we can phase out the use of semantically significant file names.

scripts/website/generate-md.sh
```diff
@@ -0,0 +1,108 @@
+#!/bin/bash
```
maxsmythe, line 1: I'm a bit leery of writing this in bash, since the complexity of this script will likely grow significantly over time (e.g. introspecting the open API schema to generate easily-grokable parameter documentation), in ways that bash is not well-suited for.

maxsmythe (commented):
Website looks awesome!

I have a couple of thoughts on how we want to store metadata and the pipeline used to generate documentation.

library/general/automount-serviceaccount-token/template.yaml
nilekhc: Could you please elaborate with a couple eg for me to get started?

scripts/website/generate-md.sh
nilekhc: Could you propose an alternative? I am open to other options. Bash was super quick to get started and consistent with things we are doing in GKL, eg we have a bash script to generate CT from src.

library/general/automount-serviceaccount-token/template.yaml
maxsmythe: so for fully-qualified, an example would be `metadata.gatekeeper.sh/title`

For names, I'm imagining this string could live both as a title to a web page and as a one-line entry in a table of constraints that describes what it does at a high-level.

`mediumName` (vs `shortName`) or `longName`

`oneLiner`

`elevatorPitch` <- I don't like this one so much

`tldr`

WRT putting it in as a field... curious about your thoughts on that.

scripts/website/generate-md.sh
maxsmythe: My instinct would be to use Golang for uniformity and code re-use (e.g. we can parse suite.yaml easily using the associated struct).

I'm not against bash as a POC, but what I'd want to avoid is continuing building off it as complexity grows simply b/c it's already there. Generally speaking, that's a common instinct vs. rewriting.

library/general/httpsonly/template.yaml
```diff
@@ -3,6 +3,7 @@ kind: ConstraintTemplate
 metadata:
   name: k8shttpsonly
   annotations:
+    title: "HTTP only"
```
ritazh, line 6:
```suggestion
    title: "HTTPS only"
```

nilekhc: fixed. I guess I relied too much on GH Copilot... 😅

library/general/automount-serviceaccount-token/template.yaml
nilekhc: Fully qualified makes sense. I have updated it to `metadata.gatekeeper.sh/title`

scripts/website/generate-md.sh
nilekhc: I added go code and created `generate-website-docs` make target to generate website docs.

scripts/website/generate-md.sh
nilekhc: Thanks. `suites.yaml` is more reliable source. I updated script to work off of this.

scripts/website/generate.go
```diff
@@ -0,0 +1,206 @@
...
+			Object     string `yaml:"object"`
+			Assertions []struct {
+				Violations string `yaml:"violations"`
+			} `yaml:"assertions"`
+		} `yaml:"cases"`
+	} `yaml:"tests"`
+}
+
+// ConstraintTemplate ...
+// ToDo (nilekh): Get this struct from the Gatekeeper repo
```
maxsmythe, line 35: framework repo has the CT structs

scripts/website/generate.go
```diff
@@ -0,0 +1,206 @@
...
+					yaml.Unmarshal(constraintTemplateContent, &constraintTemplate)
+
+					allExamples := ""
+					for _, test := range suite.Tests {
+						constraintContent, err := os.ReadFile(filepath.Join(basePath, dir.Name(), test.Constraint))
+						if err != nil {
+							fmt.Println("error while reading constraint.yaml")
+							panic(err)
+						}
+						constraintExample := fmt.Sprintf("<details>\n<summary>constraint</summary>\n\n```yaml\n%s\n```\n\n</details>\n", constraintContent)
```
maxsmythe, line 128: If you wanted a templating-style solution, you could use go templates:

https://pkg.go.dev/text/template

maxsmythe (commented):
LGTM after bash is removed and we resolve code copying stuffs.

Left some suggestions for the golang code and it could be a bit more modular in general, but that can be sussed out over time.

scripts/website/generate.go
nilekhc: I had looked at it, but they lack yaml annotations. Am I missing anything here?

nilekhc: Switched to unstructured reading.

maxsmythe (approved):
LGTM!

nilekhc:
@maxsmythe I do not have write access. Could you merge the PR?

maxsmythe:
@ritazh @sozercan Good to merge?

sozercan (approved):
Thank you! LGTM

PR #223, opened 2022-08-24
fix: Capitalize Gatekeeper

apeabody:
* Capitalize Gatekeeper as it is a proper noun.


maxsmythe:
b/243714848 is not publicly viewable, which means other reviewers don't have context. We should avoid doing this.

Basically, it says we should capitalize "Gatekeeper" because it's a proper noun.

maxsmythe (approved):
LGTM, thanks for the PR!

ritazh (approved):
LGTM

PR #226, opened 2022-09-01
Add tags validation

FaresHerhar:
# Description
I added the tags validation part.
Unlike the valid tags control that is already implemented, this one forbids the image without a tag, or the one that uses the 'latest' tag.

### DIFF
--- a/library/general/validtags/template.yaml
+++ b/library/general/validtags/template.yaml
@@ -1,0 +1,58 @@
+apiVersion: templates.gatekeeper.sh/v1
+kind: ConstraintTemplate
+metadata:
+  name: k8svalidtags
+  annotations:
+    description: >-
+      Requires container images to have a tag, and forbids the latest tag.
+spec:
+  crd:
+    spec:
+      names:
+        kind: K8sValidTags
+  targets:
+    - target: admission.k8s.gatekeeper.sh
+      rego: |
+        package k8svalidtags
+
+        violation[{"msg": msg}] {
+          container := input.review.object.spec.containers[_]
+          satisfied := [bad | bad = re_match(":latest", container.image)]
+          any(satisfied)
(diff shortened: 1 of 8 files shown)


maxsmythe (changes requested):
Thanks for the PR!

The DCO check would need to succeed before we can move forward.

Also, you mentioned another tag control. I assume you mean [K8sDisallowedTags](https://github.com/open-policy-agent/gatekeeper-library/blob/master/library/general/disallowedtags/template.yaml)

This example constraint:

```

kind: K8sDisallowedTags
metadata:
  name: container-image-must-not-have-latest-tag
spec:
  match:
    kinds:
      - apiGroups: [""]
        kinds: ["Pod"]
    namespaces:
      - "default"
  parameters:
    tags: ["latest"]
    exemptImages: ["openpolicyagent/opa-exp:latest", "openpolicyagent/opa-exp2:latest"]
```

Would:

* Throw a violation if no tag is specified
* throw a violation if any image used the `latest` tag
* would avoid throwing a violation for any of the listed exempt images.

This looks like it overlaps 100% with the functionality of this template, unless there's something I'm missing?

One thing we could do to make this template unique would be to provide a list of valid tags that each image must have.

Also, using the `exemptImages` code would allow people to carve out exemptions for injected sidecars:

https://github.com/open-policy-agent/gatekeeper-library/blob/ce3edff2f454994f883058d73fb705616c318462/library/general/disallowedtags/template.yaml#L67-L87

library/general/validtags/template.yaml
```diff
@@ -0,0 +1,58 @@
+apiVersion: templates.gatekeeper.sh/v1
+kind: ConstraintTemplate
+metadata:
+  name: k8svalidtags
+  annotations:
```
sozercan, line 5: can we add the new `metadata.gatekeeper.sh/title` annotation for the library website

sozercan:
+1 on Max's comment re: disallowedtags template

PR #235, opened 2022-09-13
feat: add HorizontalPodAutoscaler policy

ctrought:
Add a policy that disallows the following scenarios when deploying `HorizontalPodAutoscalers

1. Deployment of HorizontalPodAutoscalers with `.spec.minReplicas` or `.spec.maxReplicas` outside the ranges defined in the constraint
2. Deployment of HorizontalPodAutoscalers where the difference between `.spec.minReplicas` and `.spec.maxReplicas` is less than the configured `minimumReplicaSpread`
3. Deployment of HorizontalPodAutoscalers that do not reference a valid `scaleTargetRef` (e.g. Deployment, StatefulSet)

I acknowledged that to enforce number 3 it requires the resources to be cached which could have a large impact on memory depending on how many there are in the cluster. This would be user decision whether they'd accept the impact on memory to have it enforced. All 3 violation types (replica range, replica spread, enforce scale target ref) are optional and configured in the constraint.

Signed-off-by: Craig Trought k8s@trought.ca


maxsmythe (approved):
LGTM, thank you for the PR!

src/general/horizontalpodautoscaler/src.rego
```diff
@@ -0,0 +1,43 @@
...
+  
+  msg := sprintf("The %v <%v> is configured with minReplicas %v and maxReplicas %v which is a spread of %v replica(s). The spread must be at least %v replica(s)", [hpa.kind, hpa.metadata.name, hpa.spec.minReplicas, hpa.spec.maxReplicas, hpa.spec.maxReplicas - hpa.spec.minReplicas, input.parameters.minimumReplicaSpread])
+}
+
+violation[{"msg": msg}] {
+  input.review.kind.kind == "HorizontalPodAutoscaler"
+  hpa := input.review.object
+  input.parameters.enforceScaleTargetRef
+  
+  not data.inventory.namespace[hpa.metadata.namespace][hpa.spec.scaleTargetRef.apiVersion][hpa.spec.scaleTargetRef.kind][hpa.spec.scaleTargetRef.name]
```
ctrought, line 25: If the scaleTargetRef kind has different casing than what is considered to be standard, the policy will consider it a violation due to case sensitivity of the keys in OPA. For example, an HPA with scaleTargetRef.kind=statefulset when it exists in GK cache under the key **S**tateful**S**et. They are both valid from the perspective of the HPA, I don't know whether there is any solution to this.

maxsmythe: probably you'd have to iterate over all GVKs in `data.inventory`, setting the Kind, etc. to variables, then doing a case-insensitive comparison via `strings.lower()`.

The downside of this is that it may be less performant, depending on the contents of data.inventory.

artifacthub/library/general/horizontalpodautoscaler/1.0.0/template.yaml
```diff
@@ -0,0 +1,87 @@
+apiVersion: templates.gatekeeper.sh/v1
+kind: ConstraintTemplate
+metadata:
+  name: k8shorizontalpodautoscaler
+  annotations:
+    metadata.gatekeeper.sh/title: "Horizontal Pod Autoscaler"
+    metadata.gatekeeper.sh/version: 1.0.0
```
sozercan, line 7: can you add the requiresSyncData annotation? see examples in https://github.com/open-policy-agent/gatekeeper-library/pull/251

ctrought: absolutely, done 👍

sozercan (approved):
Sorry about the delay. LGTM

ctrought:
Hmm, could there be a minor bug in the CI tests? The inventory being applied to the test is not correct (3 -> 3, 4 -> 4, not deployment 4 ->  inventory 3) or it does not remove the old inventory before beginning the next test? The tests all pass locally via `gator verify`. I can update inventories to be completely unique to avoid conflicts with inventories used in previous tests but I imagine this should work as is. Lemme know if I'm missing something though.

> actual: Error from server (Forbidden): error when creating "library/general/poddisruptionbudget/samples/poddisruptionbudget/example_allowed_deployment4.yaml": admission webhook "validation.gatekeeper.sh" denied the request: [pod-distruption-budget] Deployment <nginx-deployment-allowed-4> has 1 replica(s) but PodDisruptionBudget <inventory-nginx-pdb-allowed-3> has minAvailable of 2, PodDisruptionBudget count should always be lower than replica(s), and not used when replica(s) is set to 1



```
  - name: example-allowed-subset-selector
    object: samples/poddisruptionbudget/example_allowed_deployment3.yaml
    inventory:
    - samples/poddisruptionbudget/example_inventory_allowed3.yaml
    assertions:
    - violations: no
  - name: example-allowed-nomatch-selector
    object: samples/poddisruptionbudget/example_allowed_deployment4.yaml
    inventory:
    - samples/poddisruptionbudget/example_inventory_allowed4.yaml
    assertions:
    - violations: no
``` 

```
$ gator verify suite.yaml -v
    === RUN   pod-disruption-budget
    === RUN   example-allowed-pdb
    --- PASS: example-allowed-pdb	(0.004s)
    === RUN   example-disallowed-pdb
    --- PASS: example-disallowed-pdb	(0.003s)
    === RUN   example-allowed-min-available
    --- PASS: example-allowed-min-available	(0.004s)
    === RUN   example-allowed-max-unavailable
    --- PASS: example-allowed-max-unavailable	(0.003s)
    === RUN   example-allowed-subset-selector
    --- PASS: example-allowed-subset-selector	(0.003s)
    === RUN   example-allowed-nomatch-selector
    --- PASS: example-allowed-nomatch-selector	(0.004s)
    === RUN   example-disallowed-min-available
    --- PASS: example-disallowed-min-available	(0.002s)
--- PASS: pod-disruption-budget	(0.025s)
ok	gatekeeper-library-upstream/library/general/poddisruptionbudget/suite.yaml	0.025s
PASS
```

JaydipGabani:
The way test execution is intended is, first apply all the inventory, then apply all the `allowed` resources & delete the same, then apply all the `disallowed` resources & delete the same, based on [here](https://github.com/open-policy-agent/gatekeeper-library/blob/master/test/bats/test.bats#L101). Test code as of now is not deleting any inventory resources, and that's why we were seeing the error in the first place. Additionally, there isn't a way to map inventory to a resource for each situation in a test case. Best to make sure that inventories are completely unique to avoid conflicts.

maxsmythe:
This is a good example of feature asymmetry between the suite.yaml tests and the bash-based tests.

IMO we should retire the bash-based tests. If we really want to run tests on a live cluster, something that runs off of suite.yaml would be better. However I personally think the live-cluster tests are more brittle than valuable.

maxsmythe:
Checking to see if tests are failing

maxsmythe:
Sorry for the delay and thanks for the ping, looks like everything is green, merging.

PR #238, opened 2022-09-16
feat: implements versioning and artifacthub directories

nilekhc:
Signed-off-by: Nilekh Chaudhari <1626598+nilekhc@users.noreply.github.com>

Policies on staging AH - https://staging.artifacthub.io/packages/search?kind=14&sort=relevance&page=1


nilekhc:
cc @maxsmythe @ritazh @sozercan

nilekhc:
@maxsmythe `Examples` are improved based on our feedback.

https://staging.artifacthub.io/packages/gatekeeper/gatekeeper-policies/k8scontainerratios?modal=examples&example=memory-ratio-only&file=constraint

Makefile
```diff
@@ -63,3 +64,7 @@ require-suites:
 .PHONY: generate-website-docs
 generate-website-docs:
 	cd $(WEBSITE_SCRIPT_DIR); go run generate.go
+
+.PHONY: generate-artifacthub-artifacts
+generate-artifacthub-artifacts:
```
sozercan, line 69: can you add this to readme in the development section?

artifacthub/library/general/allowedrepos/1.0.0/artifacthub-pkg.yml
```diff
@@ -0,0 +1,22 @@
+version: 1.0.0
+name: k8sallowedrepos
+displayName: Allowed Repositories
+createdAt: "2022-09-16T16:22:28Z"
+description: Requires container images to begin with a string from the specified list.
+digest: c8ab2f7f-56b2-4779-8e81-8badb782dca7
```
sozercan, line 6: what is this digest?

library/general/allowedrepos/template.yaml
```diff
@@ -4,6 +4,7 @@ metadata:
   name: k8sallowedrepos
   annotations:
     metadata.gatekeeper.sh/title: "Allowed Repositories"
+    version: 1.0.0
```
sozercan, line 7: @maxsmythe do you think we need `metadata.gatekeeper.sh/` prefix here?

artifacthub/library/general/allowedrepos/1.0.0/artifacthub-pkg.yml
```diff
@@ -0,0 +1,22 @@
+version: 1.0.0
+name: k8sallowedrepos
+displayName: Allowed Repositories
+createdAt: "2022-09-16T16:22:28Z"
```
sozercan, line 4: looks like this gets updated on every generate, i am not sure if we want that. can we have a create and modified field that only gets updated if the rego is updated?

scripts/artifacthub/hub.go
```diff
@@ -0,0 +1,264 @@
+package main
+
+import (
+	"fmt"
+	"io"
+	"io/ioutil"
```
sozercan, line 6: ioutil is deprecated, we should move to io or os packages

scripts/artifacthub/hub.go
```diff
@@ -0,0 +1,264 @@
...
+			basePath, _ := filepath.Abs(filepath.Join(libraryPath, entry.Name()))
+			directories, err := os.ReadDir(basePath)
+			if err != nil {
+				fmt.Println("error while listing directories under ", entry.Name())
+				panic(err)
+			}
+
+			for _, dir := range directories {
+				if dir.Type().IsDir() {
+					fmt.Println("Generating ah content for ", filepath.Join(basePath, dir.Name()))
```
sozercan, line 111:
```suggestion
					fmt.Println("Generating artifacthub content for ", filepath.Join(basePath, dir.Name()))
```

artifacthub/library/general/allowedrepos/1.0.0/artifacthub-pkg.yml
nilekhc: The digest is the way AH decides whether to refresh a package or not. It will only refresh the package if the digest is updated since the last refresh.

library/general/allowedrepos/template.yaml
nilekhc: I think fully qualified names make sense here.

nilekhc: updated to a fully qualified name.

artifacthub/library/general/allowedrepos/1.0.0/artifacthub-pkg.yml
nilekhc: updated digest to CT hash. This way changes in CT will trigger refresh on AH

artifacthub/library/general/allowedrepos/1.0.0/artifacthub-pkg.yml
nilekhc: Updated the logic to only add this when first creating the metadata file.

src/general/allowedrepos/constraint.tmpl
```diff
@@ -4,6 +4,7 @@ metadata:
   name: k8sallowedrepos
   annotations:
     metadata.gatekeeper.sh/title: "Allowed Repositories"
+    metadata.gatekeeper.sh/version: 1.0.0
```
ritazh, line 7: Can we add a validation to make sure this is bumped whenever the rego is updated?
Can we also add this to the developers doc to make sure devs bumps this when the rego is updated?

artifacthub/library/general/uniqueserviceselector/%!s(<nil>)/artifacthub-pkg.yml
```diff
@@ -0,0 +1,25 @@
+version: '%!s(<nil>)'
```
sozercan, line 1: looks like an issue with version on this one

library/general/uniqueserviceselector/template.yaml
```diff
@@ -4,6 +4,7 @@ metadata:
   name: k8suniqueserviceselector
   annotations:
     metadata.gatekeeper.sh/title: "Unique Service Selector"
+    version: 1.0.0
```
sozercan, line 7:
```suggestion
    metadata.gatekeeper.sh/version: 1.0.0
```

scripts/artifacthub/go.mod
```diff
@@ -0,0 +1,11 @@
+module artifacthub
```
sozercan, line 1: should we have a single go.mod

ritazh:
I see this does not include mutation artifacts. is there a proposal for how we want to present those in the website and artifacthub?

src/general/allowedrepos/constraint.tmpl
nilekhc: I have added an [issue](https://github.com/open-policy-agent/gatekeeper-library/issues/241) to add a CI check. I'll open a PR another PR to address it.

scripts/artifacthub/go.mod
nilekhc: I purposely kept two (and any future) modules separate so that we don't have to unnecessarily add dependencies in one due to another. Let me know if I am missing anything by not combining them.

nilekhc:
> mutation artifacts

We do not have any proposal for mutation artifacts yet. Let me open a discussion with the community.

nilekhc:
> I see this does not include mutation artifacts. is there a proposal for how we want to present those in the website and artifacthub?

We do not have any proposal for mutation artifacts yet. Let me open a discussion with the community.

artifacthub/library/general/uniqueserviceselector/%!s(<nil>)/artifacthub-pkg.yml
nilekhc: fixed

src/general/allowedrepos/constraint.tmpl
nilekhc: Also added a note about versioning in the README.

sozercan (approved):
LGTM

scripts/artifacthub/hub.go
```diff
@@ -0,0 +1,307 @@
...
+
+func getConstraintTemplateHash(constraintTemplate map[string]interface{}) string {
+	constraintTemplateBytes, err := yaml.Marshal(constraintTemplate)
+	if err != nil {
+		fmt.Println("error while marshaling constraint template")
+		panic(err)
+	}
+
+	hash := sha256.New()
+	hash.Write(constraintTemplateBytes)
```
maxsmythe, line 221: Is this hash value prone to changing if the serialization order of the YAML fields changes?

If so, should we operate off of bytes-on-disk to avoid that risk?

nilekhc: @maxsmythe I did a quick test by altering the YAML field's position and it doesn't change the hash value.

scripts/artifacthub/hub.go
```diff
@@ -0,0 +1,307 @@
...
+		}
+	}
+
+	// copy content
+	source := filepath.Join(rootDir, basePath)
+	destination := filepath.Join(rootDir, ahEntryPoint, basePath, version)
+	ahBasePath := filepath.Join(ahEntryPoint, basePath, version)
+
+	// copy directory
+	copyDirectory(source, destination)
```
ritazh, line 155: No error checking or handling here?

scripts/artifacthub/hub.go
```diff
@@ -0,0 +1,307 @@
...
+		return err
+	}
+
+	for _, fd := range directoryFileInfo {
+		sourceFilePath := path.Join(src, fd.Name())
+		destinationFilePath := path.Join(dst, fd.Name())
+
+		if fd.IsDir() {
+			if err = copyDirectory(sourceFilePath, destinationFilePath); err != nil {
+				fmt.Println(err)
```
ritazh, line 271: Is logging the error sufficient here? how do we know if they failed as part of the CI?

scripts/artifacthub/hub.go
```diff
@@ -0,0 +1,307 @@
...
+			}
+		} else {
+			if err = copyFile(sourceFilePath, destinationFilePath); err != nil {
+				fmt.Println(err)
```
ritazh, line 275: same

scripts/artifacthub/hub.go
maxsmythe: How did you alter the YAML field's position?

Here is a quick example that shows character order altering the sha256 sum:

```
$ echo '{"a": 1, "b": 2}' | sha256sum
fdcde7d03783fce2e97b7a775ef45379b3ecab4cab4ddb1156e65057b811f44c  -

$ echo '{"b": 2, "a": 1}' | sha256sum
76dd7bd8db01cd533f684ccaacdbc144d08146d6aaa023212f90df348acb3877  -

```

nilekhc: I simulated hub.go's hash generation which also produces the same hash values:
https://go.dev/play/p/W4mDQThhj0X

Let me know if I am missing anything.

maxsmythe: YAML does not consider the order of keys semantically significant:
https://yaml.org/spec/1.2.2/ (look for "3.2.2.1. Mapping Key Order")

It looks like currently Golang's YAML library sorts keys before serializing:
https://github.com/go-yaml/yaml/blob/7649d4548cb53a614db133b2a8ac1f31859dda8c/encode.go#L190

So that's stable for now (however keys in Golang maps are also unordered, so that could be another cause of non-determinism).

There are also non-semantically-significant changes, such as text wrapping, that can cause the hash to drift. Here is an example:

https://go.dev/play/p/0u-_opXnKRP

TL;DR:

* If all we are looking to do with the hash is to detect changes, this is probably sufficient.
* If you are relying on hashes as an identity comparison, then generating the hash carefully is necessary in order to avoid latent bugs
* One way of doing that is to version the CT as originally serialized. This would be stable, but the hash would need to be preserved with the source and passed around. It cannot be re-derived (this makes it essentially equivalent to a version string)
* Another way to do this is to create a semantic hash, where CTs with semantically equivalent content always have the same hash (e.g. irrelevant changes are filtered out). This has the advantage of being stable and re-derivable, but is complex to code and needs some defining (e.g. is "description" part of the identity even if it doesn't effect behavior?)

TL;DR2

The choice made here is going to be a tradeoff between brittleness, difficulty, and the significance/usability of the hash.

nilekhc: Thanks for the details @maxsmythe. They are helpful. 

The hash field here is only used by AH to determine if it should refresh the content on the website or not. So even if there are any false positives we should be good here.

maxsmythe: Ack, SGTM

maxsmythe (approved):
LGTM

PR #246, opened 2022-10-15
Constraint and Template change for requiredProbes to add supoort for Deployment and StatefulSet

samba1112:
This commit makes to restrict for Deployment and StatefulSet along with Pod

### DIFF
--- a/library/general/requiredprobes/samples/must-have-probes/constraint.yaml
+++ b/library/general/requiredprobes/samples/must-have-probes/constraint.yaml
@@ -7,6 +7,8 @@ spec:
     kinds:
       - apiGroups: [""]
         kinds: ["Pod"]
+      - apiGroups: ["apps"]
+        kinds: ["Deployment","StatefulSet"]
   parameters:
     probes: ["readinessProbe", "livenessProbe"]
     probeTypes: ["tcpSocket", "httpGet", "exec"]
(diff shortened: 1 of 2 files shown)


ctrought:
+1 I had the same thoughts

1. better to enforce the policy at the controller level to reduce load and spam (ie. it's unnecessary to report warning for multiple pods owned by the same deployment, sts, etc.).
2. allows for easy exclusion of pods created by controllers that often don't have probes such as CronJobs/Jobs
3. generally users are not creating pods directly themselves, so warnings emitted don't always bubble up in a place they will immediately see them.

maxsmythe:
Thanks for the PR!

We are looking to fix this at a framework level so that individual templates don't need to be aware of which resources can create other resources. Here is the documentation for the (alpha) feature:

https://open-policy-agent.github.io/gatekeeper/website/docs/workload-resources

In addition, there are scoping reasons why it may not make sense to write constraints against generator resources, discussed in this old Gatekeeper thread:

https://github.com/open-policy-agent/gatekeeper/issues/298

ritazh:
Closing this in favor of https://open-policy-agent.github.io/gatekeeper/website/docs/workload-resources @samba1112 @ctrought feel free to reopen if your use cases are not addressed.

PR #247, opened 2022-10-17
fix: PodDisruptionPolicy match on pdb labels that are a subset of deployment labels

### LINKED ISSUE #239: Unused parameter in one of the PodDisruptionBudget constraint messages
The [validation message for one of the PodDisruptionBudget checks](https://github.com/open-policy-agent/gatekeeper-library/blob/36c420b5d40a65cb8d901b1a90ac996566a0af5b/src/general/poddisruptionbudget/src.rego#L33-L34) doesn't make use of the final parameterm, and so ends up being rendered like `Deployment <bar-app> has 1 replica(s) but PodDisruptionBudget <bar-app> has minAvailable of 1, PodDisruptionBudget count should always be lower than replica(s), and not used when replica(s) is set to 1%!(EXTRA int=1)`.

The final `obj.spec.replicas` should probably just be removed?

ctrought:
When the current pdb policy evaluates matchLabels it requires the pdb's matchLabels to be equivalent to the workload being evaluated to be considered for the policy. In reality it is possible that only a subset of the labels defined on the deployments matchLabels may be included in the pdb's matchLabels (regardless of what is best practice). This results in pdb's not being checked if the updated deployment has extra labels. This PR modifies the behaviour of the policy to be less restrictive when matching the labels by only considering the labels defined on the PDB.

Also fixes #239 

Signed-off-by: Craig Trought <k8s@trought.ca>


maxsmythe (approved):
Thanks for the PR!

Ideally we should add a test to suite.yaml to make sure this case continues to be handled if/when the template's code changes, but the code LGTM.

Here is where the suite.yaml lives:

https://github.com/open-policy-agent/gatekeeper-library/blob/master/library/general/poddisruptionbudget/suite.yaml

library/general/poddisruptionbudget/template.yaml
```diff
@@ -36,7 +36,10 @@ spec:
         violation[{"msg": msg}] {
           obj := input.review.object
           pdb := data.inventory.namespace[obj.metadata.namespace]["policy/v1"].PodDisruptionBudget[_]
-          obj.spec.selector.matchLabels == pdb.spec.selector.matchLabels
+
```
ritazh, line 39: Thanks for the PR @ctrought!
For bug fixes like this, do we need to bump the constraint template version?
> metadata.gatekeeper.sh/version: 1.0.0

+1 on adding a test to suite.yaml

nilekhc: Yes, we should.

Adding versioning doc for ref: https://docs.google.com/document/d/1IYiypA-mRcdfSVfmoeyuaeG8XtA1u4GkcqH3kEkv2uw/edit#

nilekhc: > Thanks for the PR @ctrought! For bug fixes like this, do we need to bump the constraint template version? @maxsmythe @sozercan @nilekhc
> 
> > metadata.gatekeeper.sh/version: 1.0.0
> 
> +1 on adding a test to suite.yaml

JFYI `suite.yaml` is mandatory since our website and AH treat that as the source of truth.

ctrought: Thanks all, will do! I was thinking it might also be good to have an extra test case to avoid over-enforcement. While not true in the current policy, if the template were to change it could be possible for all current test cases to pass while also inadvertently rejecting updates to workloads that aren't linked to any PDB. All the test cases only include tests of deployments & pdb's that are linked (e.g. Deployment violation depends on PDB, but not all Deployments have PDBs so we should include a test to ensure they are not inadvertently affected). This might be a good practice to apply to any referential based policies that are conditional. Take a look and let me know if that's alright.

maxsmythe: I like the idea of testing to make sure the non-enforced cases are maintained, for sure.

WRT the looser match criteria, I don't think there is not much we can do there.

artifacthub/library/general/poddisruptionbudget/1.0.0/template.yaml
```diff
@@ -36,7 +36,10 @@ spec:
         violation[{"msg": msg}] {
           obj := input.review.object
           pdb := data.inventory.namespace[obj.metadata.namespace]["policy/v1"].PodDisruptionBudget[_]
-          obj.spec.selector.matchLabels == pdb.spec.selector.matchLabels
+
+          matchLabels := { [label, value] | some label; value := pdb.spec.selector.matchLabels[label] }
```
ritazh, line 40: @nilekhc should artifacthub 1.0.0 artifacts be untouched? these changes should only go to the 1.0.1 folder right?

nilekhc: The entire `/artifact` directory should be untouched. We should update the template only inside `/src` with the correct version.  After that run `make generate-artifacthub-artifacts` and it will auto-update `/artifact` directory.

nilekhc: It looks like the `generate-artifacthub-artifacts` must have been run before updating the template version. This should go in 1.0.1

ctrought: Woops, forgot to correct that.

sozercan (approved):
Sorry about the delay. LGTM

ritazh (approved):

PR #251, opened 2022-10-27
feat: add requiresSyncData and verify-sync test

### LINKED ISSUE #248: New Annotation for Constraint Templates' which require SyncData
I propose (and can submit the PR) for adding a new Constraint Templates annotation for the 4 Constraint Templates which require SyncData:
- `K8sPodDisruptionBudget`
- `K8sStorageClass`
- `K8sUniqueIngressHost`
- `K8sUniqueServiceSelector`

Here is an example of what that would look like for `K8sStorageClass`:

```
  annotations:
    metadata.gatekeeper.sh/requiresSyncData: |-
      [
         {
            "group":"storage.k8s.io",
            "version":"v1",
            "kind":"StorageClass"
         }
      ]
```

For Constraint Templates that don't require SyncData, it should simply be omitted.

apeabody:
* Adds a new `metadata.gatekeeper.sh/requiresSyncData` annotation to templates using `data.inventory`
* Add a `sync.yaml` for `uniqueingresshost` template
* Adds a new `require-sync` test which requires templates using data.inventory have both a `sync.yaml` and `metadata.gatekeeper.sh/requiresSyncData` annotation

Fixes #248


apeabody:
I left the template versions as-is given the actual rego/etc hasn't changed, but let me know if we should bump the minor or patch.  Thanks!

library/general/uniqueingresshost/template.yaml
```diff
@@ -5,6 +5,24 @@ metadata:
   annotations:
     metadata.gatekeeper.sh/title: "Unique Ingress Host"
     metadata.gatekeeper.sh/version: 1.0.0
+    metadata.gatekeeper.sh/requiresSyncData: |
+      "[
+        {
+          "group":"extensions",
```
maxsmythe, line 11: Do all K8s versions have all of these GVKs? Interesting to think about cases where there may be missing GVKs on older clusters (or newer clusters if the "extensions" group has been removed)

scripts/require-sync.sh
```diff
@@ -0,0 +1,34 @@
+#!/bin/bash
+
+# Look in the library and find every `template.yaml` file
+TEMPLATE_PATHS=$(find . -name "*template.yaml")
+
+FAILURE=false
+for path in $TEMPLATE_PATHS; do
+
+  # Check for presence of 'data.inventory'
+  if ! grep -q 'data.inventory' "$path"; then
```
maxsmythe, line 10: This is not a 100% reliable way of detecting requiring inventory (e.g. `input.parameters.data.inventory` would trigger this as well).

More reliable would be to attempt to compile the template with `inventory` disabled in the constraint framework, but TBH I'm not sure this is a blocker.

It might be good to give `gator test` the ability to turn of inventory (@julianKatz 's old PR), as a way of easily and reliably detecting something like this via bash.

maxsmythe (commented):
Basically LGTM. Ideally we'd use more reliable detection for `data.inventory` usage, but I think requiring it would depend on how much work that entails.

The question around "what about clusters for whom these resources don't exist" is salient if we want to think about the expected uses of these annotations.

maxsmythe:
Also, WRT template versions, leaving them as-is seems right to me (no Rego changes)

library/general/uniqueingresshost/template.yaml
ekitson: Ah interesting. Maybe this schema needs a slight tweak to account for this, eg:
requiresSyncData: [list of {requirement}s, ANDed]
{requirement}: [list of {GVK}s, ORed]

That would give enough expression to account for things like "v1beta1 fully deprecated in favor of v1" or "Deployments migrated from extensions to apps"

scripts/require-sync.sh
apeabody: Thanks @maxsmythe, yes it would be great if `gator test` had that capability.  In the meantime I updated the grep to `' data.inventory.'` (space at the start) which will at least avoid the possibility of a false match on `input.parameters.data.inventory`.

library/general/uniqueingresshost/template.yaml
maxsmythe: I wouldn't presume kind to be stable, but otherwise +1

apeabody: Thanks @maxsmythe @ekitson !  Based on the feedback the PR has been updated to use this [ [{}OR{}] AND [{}OR{}] ] format:

```
    metadata.gatekeeper.sh/requiresSyncData: |
      "[
        [
          {
            "groups":["extensions"],
            "versions": ["v1beta1"],
            "kinds": ["Ingress"]
          },
          {
            "group": ["networking.k8s.io"],
            "version": ["v1beta1", "v1"],
            "kind": ["Ingress"]
          }
        ]
      ]"
```

maxsmythe (approved):
LGTM!

scripts/require-sync.sh
```diff
@@ -0,0 +1,34 @@
+#!/bin/bash
```
nilekhc, line 1: @maxsmythe Do we want to shift towards `Golang` for such new tasks?

maxsmythe: +1 good point @apeabody that should also let us alleviate the issue with detecting referential constraints.

library/general/uniqueingresshost/template.yaml
apeabody: Both the schema and detection method have been updated.

scripts/require-sync.sh
apeabody: Replaced with a golang version which also has a better check for referential constraints.

website/docs/replicalimits.md
```diff
@@ -135,7 +135,7 @@ spec:
 Usage
 
 ```shell
-kubectl apply -f https://raw.githubusercontent.com/open-policy-agent/gatekeeper-library/master/library/general/replicalimits/samples/replicalimits/example_allowed.yaml
+kubectl apply -f https://raw.githubusercontent.com/open-policy-agent/gatekeeper-library/master/library/general/replicalimits/samples/replicalimits/constraint.yaml
```
sozercan, line 138: I think we need the revert this change

website/docs/replicalimits.md
```diff
@@ -168,7 +168,7 @@ spec:
 Usage
 
 ```shell
-kubectl apply -f https://raw.githubusercontent.com/open-policy-agent/gatekeeper-library/master/library/general/replicalimits/samples/replicalimits/example_disallowed.yaml
+kubectl apply -f https://raw.githubusercontent.com/open-policy-agent/gatekeeper-library/master/library/general/replicalimits/samples/replicalimits/constraint.yaml
```
sozercan, line 171: same here

artifacthub/library/general/poddisruptionbudget/1.0.0/template.yaml
```diff
@@ -5,6 +5,16 @@ metadata:
   annotations:
     metadata.gatekeeper.sh/title: "Pod Disruption Budget"
     metadata.gatekeeper.sh/version: 1.0.0
```
sozercan, line 7: @nilekhc would we want to bump the version in this case?

scripts/require-sync/go.mod
```diff
@@ -0,0 +1,111 @@
+module require-sync
+
+go 1.19
+
+// Prevent otel dependencies from getting out of sync.
+replace (
+	go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp => go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp v0.20.0
```
sozercan, line 7: do we need the otel dependencies?

scripts/require-sync/main.go
```diff
@@ -0,0 +1,193 @@
...
+
+	driver, err := local.New(local.Tracing(false), externs)
+	if err != nil {
+		return nil, fmt.Errorf("creating driver: %w", err)
+	}
+
+	if referential {
+		driver, err = local.New(local.Tracing(false), local.Externs("inventory"))
+	}
+	if err != nil {
```
sozercan, line 183: does this need to be moved inside the referential check?

.github/workflows/workflow.yaml
```diff
@@ -78,6 +78,14 @@ jobs:
...
         run: |
           make require-suites
+  require_sync:
+    runs-on: ubuntu-latest
+    name: "Require a sync.yaml file and metadata.gatekeeper.sh/requiresSyncData annotation for every template.yaml using data.inventory"
+    steps:
+      - uses: actions/checkout@v2
+      - name: Run script
+        run: |
+          make require-sync
```
sozercan, line 88: if this fails, can we output an actionable message to the user

website/docs/replicalimits.md
apeabody: Thanks @sozercan - It appears there were manually updated in https://github.com/open-policy-agent/gatekeeper-library/commit/1def56b845b5c7a1acf37b9f611c7e7a8835dd68, and are being changed back "make generate generate-website-docs".  I'll need to look into it further.

apeabody: Hi @sozercan - I traced this to an issue with the website generate script #254 that was impacting all the website policy examples.  I just opened PR #255 to resolve.

apeabody:
Pending #255

.github/workflows/workflow.yaml
apeabody: Yes, `require-sync` will return a message with the template and what is missing the user needs to add:

`template at path '%s' is missing annotation with key '%s'`
or
`sync.yaml not found in dir '%s'`

scripts/require-sync/main.go
apeabody: Yes, that whole section has been rewritten.

scripts/require-sync/go.mod
apeabody: I believe these will be required till this issue is resolved: https://github.com/open-policy-agent/frameworks/issues/210

artifacthub/library/general/poddisruptionbudget/1.0.0/template.yaml
apeabody: I'm happy to bump if desired, however there is no change to the rego/template behavior.

scripts/require-sync/main.go
```diff
@@ -0,0 +1,193 @@
...
+			if err != nil {
+				return fmt.Errorf("detecting referential: %w", err)
+			}
+			if isRef {
+				// Get template name
+				templateName, err := getTemplateName(krm.ConstraintTemplate)
+				if err != nil {
+					return err
+				}
+				log.Println(fmt.Errorf("Found Referential Template: %s", templateName))
```
maxsmythe: is this an error? Should we use `fmt.Sprintf()`?

maxsmythe (approved):
1 nit, but LGTM

scripts/require-sync/main.go
apeabody: Hi @maxsmythe  - I believe that's from an earlier commit, it shouldn't be in the current version: https://github.com/open-policy-agent/gatekeeper-library/pull/251/files#diff-f987eb96fe4bcda2fea78c6a08551b7947c935c4a3a4d5ce8af88083957e9af2R96

scripts/require-sync/main.go
```diff
@@ -0,0 +1,193 @@
...
+	gkapis "github.com/open-policy-agent/gatekeeper/apis"
+	"github.com/open-policy-agent/gatekeeper/pkg/gator"
+	"github.com/open-policy-agent/gatekeeper/pkg/target"
+)
+
+const syncAnnotation string = "metadata.gatekeeper.sh/requiresSyncData"
+
+var (
+	pathFlag = flag.String("path", "", "Path to verify referential templates include sync data.")
+	fileFlag = flag.Bool("sync_file", false, "When `true`, require a `sync.yaml` file for each referential template.")
```
maxsmythe, line 29: nit: `sync-file` for flag consistency

maxsmythe (approved):
LGTM after 1 nit

scripts/require-sync/main.go
apeabody: SG - done

scripts/require-sync/go.mod
```diff
@@ -0,0 +1,136 @@
+module require-sync
```
sozercan, line 1: can you run `go mod tidy`? i am seeing a few updates on my side and looks like `go.work.sum` file is also generated

sozercan (approved):
minor comment, otherwise LGTM

.github/workflows/workflow.yaml
ritazh, line 88: it's good that we are checking here. should we also add validation in Gatekeeper to write an error in status if the sync resource is missing the required data in `metadata.gatekeeper.sh/requiresSyncData`?

scripts/require-sync/go.mod
apeabody: Done, should have done that after the rebase.

artifacthub/library/general/poddisruptionbudget/1.0.0/template.yaml
ritazh: Since we are adding a new annotation (that may be used for validation), we should at least bump the patch version IMO. WDYT?

nilekhc: Yup. We should bump the patch version.

.github/workflows/workflow.yaml
apeabody: Thanks @ritazh! - Absolutely, however that scope of work is beyond what I can include in this particular PR.  Should we track as a feature request issue for future contribution?  @maxsmythe @ekitson

maxsmythe: > or is missing in data.inventory

I don't think this can be reliably determined: is it missing because it's not being synced or because there is nothing to sync?

> write an error in status if the required data specified in metadata.gatekeeper.sh/requiresSyncData is not part of the sync resource

This makes sense to do, but is non-trivial. I'd file an issue in G8r.

apeabody: Opened: https://github.com/open-policy-agent/gatekeeper/issues/2393

apeabody:
Hi Everyone - Is anything else needed before this can be merged?  Thanks!

maxsmythe:
@ritazh any outstanding comments?

ritazh (approved):
LGTM

PR #253, opened 2022-11-02
(Successor of #21) Add an allow list parameter, allowedSysctls, to the Forbidden sysctl constraint template

ordovicia:
This PR is a successor of #21 and adds description for the new field, as per https://github.com/open-policy-agent/gatekeeper-library/pull/21#issuecomment-983239278.

Quoting PR description from #21:

> The `K8sPSPForbiddenSysctls` constraint template allows limiting sysctls available to pods using a deny list. This change adds another option, `allowedSysctls`, which allows a constraint to specify an allow list in addition to, or instead of, the deny list.
> 
> The matching logic is as follows:
> 
> 1. When specified, any sysctl not in the allow list is considered to be forbidden.
> 2. The allow list can be omitted.
> 3. The deny list takes precedence.


src/pod-security-policy/forbidden-sysctls/src.rego
```diff
@@ -1,11 +1,19 @@
...
     msg := sprintf("The sysctl %v is not allowed, pod: %v. Forbidden sysctls: %v", [sysctl, input.review.object.metadata.name, input.parameters.forbiddenSysctls])
 }
 
+# Block if not explicitly allowed
+violation[{"msg": msg, "details": {}}] {
+    sysctl := input.review.object.spec.securityContext.sysctls[_].name
+    not allowed_sysctl(sysctl)
+    msg := sprintf("The sysctl %v is not explictly allowed, pod: %v. Allowed sysctls: %v", [sysctl, input.review.object.metadata.name, input.parameters.allowedSysctls])
+}
+
```
nilekhc, line 16: It looks like rego is being updated here. Please bump the version as per: https://docs.google.com/document/d/1IYiypA-mRcdfSVfmoeyuaeG8XtA1u4GkcqH3kEkv2uw/edit#

nilekhc: cc @maxsmythe @ritazh @sozercan

artifacthub/library/pod-security-policy/forbidden-sysctls/1.0.0/template.yaml
```diff
@@ -51,3 +68,16 @@ spec:
         forbidden_sysctl(sysctl) {
             startswith(sysctl, trim(input.parameters.forbiddenSysctls[_], "*"))
         }
+
+        # * may be used to allow all sysctls
+        allowed_sysctl(sysctl) {
```
maxsmythe, line 73: Is there a difference between this and not using the constraint? Though if we allow globs, this is probably redundant anyway.

artifacthub/library/pod-security-policy/forbidden-sysctls/1.0.0/template.yaml
```diff
@@ -51,3 +68,16 @@ spec:
...
+            input.parameters.allowedSysctls[_] == "*"
+        }
+
+        allowed_sysctl(sysctl) {
+            input.parameters.allowedSysctls[_] == sysctl
+        }
+
+        allowed_sysctl(sysctl) {
+            startswith(sysctl, trim(input.parameters.allowedSysctls[_], "*"))
```
maxsmythe, line 82: I think this may unintentionally treat all `allowedSysctls` as having "*", we should filter to make sure that the allowedSysctl item includes `*` prior to attempting fuzzy matching. Also, since we are doing prefix matching, we should only remove the trailing `*`, not any leading `*`s.

I realize these bugs are present above for the disallowed list, but it's more of a security issue for "allowed sysctls".

IMO we should fix the disallowed list too, but not a blocker for me for this PR.

src/pod-security-policy/forbidden-sysctls/src.rego
maxsmythe: +1, because we're adding a new parameter field, we should bump the minor version

ordovicia: Updated the minor version in 5108082aaa6b8b03bad20aff77889eb997fce7d2

artifacthub/library/pod-security-policy/forbidden-sysctls/1.0.0/template.yaml
ordovicia: I think this rule should exist to accept `allowedSysctls: ["*"]`. Although this pattern can be handled by the third rule too, the third rule is for prefix matching.

artifacthub/library/pod-security-policy/forbidden-sysctls/1.0.0/template.yaml
ordovicia: Quite right. I fixed the rules, including one for `forbiddenSysctls`, in 140279af39b1376d8e028124b9a1be35e436d972

artifacthub/library/pod-security-policy/forbidden-sysctls/1.0.0/template.yaml
```diff
@@ -33,12 +42,20 @@ spec:
...
             sysctl := input.review.object.spec.securityContext.sysctls[_].name
             forbidden_sysctl(sysctl)
             msg := sprintf("The sysctl %v is not allowed, pod: %v. Forbidden sysctls: %v", [sysctl, input.review.object.metadata.name, input.parameters.forbiddenSysctls])
         }
 
+        # Block if not explicitly allowed
+        violation[{"msg": msg, "details": {}}] {
+            sysctl := input.review.object.spec.securityContext.sysctls[_].name
+            not allowed_sysctl(sysctl)
+            msg := sprintf("The sysctl %v is not explictly allowed, pod: %v. Allowed sysctls: %v", [sysctl, input.review.object.metadata.name, input.parameters.allowedSysctls])
```
maxsmythe, line 56: We should add a comment that `input.parameters.allowedSysctls` in `sprintf()` is load-bearing.

Without that line, this change would be backwards-incompatible and that seems non-obvious. Adding a comment will help prevent accidentally breaking this requirement in the future.

maxsmythe (approved):
LGTM after nit!

artifacthub/library/pod-security-policy/forbidden-sysctls/1.0.0/template.yaml
ordovicia: Sorry, what specifically do you mean by "load-bearing"?

artifacthub/library/pod-security-policy/forbidden-sysctls/1.0.0/README.md
```diff
@@ -1,5 +1,36 @@
 # Forbidden Sysctls security context policy
 
-Forbidden sysctls excludes specific sysctls. You can forbid a combination of safe and unsafe sysctls in the list. To forbid setting any sysctls, use `*` on its own. If a sysctl pattern ends with a `*` character, such as `kernel.*`, it'll match `*` with rest of the sysctl.
+The forbidden sysctls constraint allows one to limit the set of kernel parameters that can be modified by pods. This is accomplished by specifying a combination of allowed and forbidden sysctls using either of two parameters: `allowedSysctls` and `forbiddenSysctls`.
```
ritazh, line 3: I believe 1.0.0 autogenerated changes should be removed now that the version has been bumped to 1.1.0 in the source.

ordovicia: Reverted in 4fc2715c2ef1f0f1a58d302bcfb2cd97bdad1d2d

ordovicia:
Sorry, I accidentally removed a review request from @sozercan

ordovicia:
Hi, @maxsmythe @ritazh @sozercan  cloud you please take a look again and merge this?

website/docs/replicalimits.md
```diff
@@ -135,7 +135,7 @@ spec:
 Usage
 
 ```shell
-kubectl apply -f https://raw.githubusercontent.com/open-policy-agent/gatekeeper-library/master/library/general/replicalimits/samples/replicalimits/example_allowed.yaml
```
ritazh, line 138: I don't think you need to update these since these should be the example deployments, not the constraints. as you can see https://open-policy-agent.github.io/gatekeeper-library/website/replicalimits#examples

ordovicia: I regenerated the website and the change was reverted.

PR #265, opened 2022-12-10
Fix formatting in poddisruptionbudget.md

triangularcover:

Fixes #

### DIFF
--- a/website/docs/poddisruptionbudget.md
+++ b/website/docs/poddisruptionbudget.md
@@ -6,7 +6,9 @@ title: Pod Disruption Budget
 # Pod Disruption Budget
 
 ## Description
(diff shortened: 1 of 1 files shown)


maxsmythe (commented):
Thanks for the PR! I think this change would get clobbered the next time we generated website docs.

@nilekhc any breadcrumbs for how to fix the formatting of the generated docs?

nilekhc:
> Thanks for the PR! I think this change would get clobbered the next time we generated website docs.
> 
> @nilekhc any breadcrumbs for how to fix the formatting of the generated docs?

This is being read from the description annotation of CT. We need to see if we can preserve that formatting here - https://github.com/open-policy-agent/gatekeeper-library/blob/0a8f69f51a396790e4d6c7b0346b1565b107c15b/scripts/website/generate.go#L132

PR #267, opened 2022-12-13
feat: add replicasknative policy

romerobu:
Add a policy for setting a max amount of replicas for knative services.

1. maxScale must be defined.
2. initialScale if defined, cannot be greater than max replicas.
3. minScale if defined, cannot be greater than maxScale.
4. minScale if defined, cannot be greater than max replicas.
5. maxScale cannot be greater than max replicas.


JaydipGabani (changes requested):

artifacthub/library/general/replicasknative/1.0.0/template.yaml
```diff
@@ -0,0 +1,79 @@
...
+            to_number(replicas) < to_number(max_scale)
+            msg := sprintf("maxScale value %v cannot be greater than %v replicas", [max_scale, replicas])
+        }
+
+        violation[{"msg": msg}] {
+            replicas := input.parameters.replicas
+            service := input.review.object
+            min_scale := service.spec.template.metadata.annotations["autoscaling.knative.dev/minScale"]
+            to_number(replicas) < to_number(min_scale)
+            msg := sprintf("minScale value %v cannot be greater than %v replicas", [min_scale, replicas])
```
JaydipGabani, line 61:
```suggestion
            msg := sprintf("replicas value %v cannot be smaller than minScale value %v ", [replicas, min_scale])
```
If we are trying to define range for replicas, this may be a better wording for the message.

sozercan:
@romerobu thanks for the PR! Looks like DCO needs to be signed

PR #273, opened 2023-01-05
feat: add ability to restrict required-probes constraint to containers in pods selected by a service

ctrought:
Signed-off-by: Craig Trought <k8s@trought.ca>

Adds a parameter to restrict enforcement of the constraint to containers in pods that are selected by a service with at least one matching targetPort. For workloads that do not accept incoming traffic in pods it's common to not have readiness probes. This will allow the user to avoid generating violations for pods that do not accept incoming traffic via services. To leverage this services must be cached. This is a non-breaking change as the original implementation without caching is still supported by omitting the new parameter.

Adds a parameter to set an optional custom violation string that is appended to the standard one. I can revert this if it's not desired for the library, but we like to tailor certain violation messages with additional context so having the ability to do that in the constraint itself provides us the flexibility to do that without needing to alter the constraint template itself.

Removes the pod name from the violation message to avoid issues with expanded objects that likely won't contain a name. Seen in https://github.com/open-policy-agent/gatekeeper/issues/2485


maxsmythe:
The Rego LGTM, haven't taken a closer look at the rest yet.

@apeabody @ekitson

Here is an example of a constraint template where referential data is only needed for certain configurations.

library/general/requiredprobes/samples/must-have-probes/constraint.yaml
```diff
@@ -3,10 +3,12 @@ kind: K8sRequiredProbes
 metadata:
   name: must-have-probes
 spec:
+  enforcementAction: warn
```
apeabody, line 6: Hi @ctrought  I believe the CI is depending on the default `enforcementAction` mode which is deny.

```suggestion
```

ctrought: Oops my bad - fixed, thanks!

website/docs/requiredprobes.md
```diff
@@ -16,7 +16,17 @@ metadata:
   name: k8srequiredprobes
   annotations:
     metadata.gatekeeper.sh/title: "Required Probes"
-    metadata.gatekeeper.sh/version: 1.0.0
+    metadata.gatekeeper.sh/version: 1.1.0
```
JaydipGabani, line 19: Just out of curiosity, how are we deciding on version increments? why `1.1.0` and not `1.0.1`?

artifacthub/library/general/requiredprobes/1.1.0/samples/must-have-probes/example_disallowed.yaml
```diff
@@ -0,0 +1,30 @@
...
+  containers:
+  - name: nginx-1
+    image: nginx:1.7.9
+    ports:
+    - containerPort: 80
+    livenessProbe:
+      # tcpSocket:
+      #   port: 80
+      # initialDelaySeconds: 5
+      # periodSeconds: 10
```
JaydipGabani, line 15: nit: remove commented code if not needed

JaydipGabani (approved):

JaydipGabani:
Few nits, but otherwise LGTM

website/docs/requiredprobes.md
ctrought: I was going based off the versioning doc linked in the readme. It says minor updates include parameter changes that are backwards compatible. 

https://docs.google.com/document/u/0/d/1IYiypA-mRcdfSVfmoeyuaeG8XtA1u4GkcqH3kEkv2uw/mobilebasic

sozercan:
@ctrought thanks for the PR! Looks like there are conflicts, can you resolve when you get a chance?

Makefile
```diff
@@ -54,7 +54,6 @@ __build-gator:
 .PHONY: generate
 generate: __build-gomplate
 	$(docker) run \
-		-u $(shell id -u):$(shell id -g) \
```
apeabody, line 57: Was this causing an issue?

apeabody (approved):
Thanks for the contribution @ctrought!

@sozercan The CI is now passing

Makefile
ctrought: Sorry, accidental commit. Yes it causes problems when running with rootless podman.

https://github.com/open-policy-agent/gatekeeper/pull/2632#pullrequestreview-1349305142

apeabody:
Thanks for the contribution @ctrought - When you get a chance, I think there are just a few very minor open items above.

artifacthub/library/general/requiredprobes/1.1.0/samples/must-have-probes-on-service/example_disallowed_with_service.yaml
```diff
@@ -0,0 +1,35 @@
...
+  containers:
+  - name: nginx-1
+    image: nginx:1.7.9
+    ports:
+    - containerPort: 80
+    livenessProbe:
+      # tcpSocket:
+      #   port: 80
+      # initialDelaySeconds: 5
+      # periodSeconds: 10
```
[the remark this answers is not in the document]
ctrought: These are also commented in the original sample tests these were generated from. I believe this is to highlight the probe field is intentionally left empty and the policy still generates violations for these cases. Though it's not really necessary to keep, this is my interpretation of the original. I'm good either way, but it should be consistent I think.

https://github.com/open-policy-agent/gatekeeper-library/blob/master/library/general/requiredprobes/samples/must-have-probes/example_disallowed.yaml#L11

src/general/requiredprobes/constraint.tmpl
```diff
@@ -4,7 +4,17 @@ metadata:
   name: k8srequiredprobes
   annotations:
     metadata.gatekeeper.sh/title: "Required Probes"
-    metadata.gatekeeper.sh/version: 1.0.0
+    metadata.gatekeeper.sh/version: 1.1.0
+    metadata.gatekeeper.sh/requiresSyncData: |
```
apeabody, line 8: Hi @ctrought - This update and a rebuild should resolve the CI.

```suggestion
    metadata.gatekeeper.sh/requires-sync-data: |
```

maxsmythe (commented):
Rego code and idea LGTM (did not look at rest of PR structure, tests, etc., relying on other reviewers there)

library/general/requiredprobes/template.yaml
```diff
@@ -4,7 +4,17 @@ metadata:
   name: k8srequiredprobes
   annotations:
     metadata.gatekeeper.sh/title: "Required Probes"
-    metadata.gatekeeper.sh/version: 1.0.0
+    metadata.gatekeeper.sh/version: 1.1.0
+    metadata.gatekeeper.sh/requires-sync-data: |
```
ritazh, line 8: To clarify a bit around the UX, with the `onlyServices` parameter, it seems syncing service is optional, however does this annotation make syncing service a requirement?

maxsmythe: @apeabody this is a good question. How do we deal with conditionally referential templates? Should we just have a separate template, instead?

apeabody: Creating a separate template is an option, although I believe our goal should be to minimize the overall quantity of templates.

The current release Gatekeeper doesn't enforce `requires-sync-data`, and unmarshaling of the annotation was only added in [v3.13.0-beta.1](https://github.com/open-policy-agent/gatekeeper/releases/tag/v3.13.0-beta.1).  Depending on the roadmap, this might not be a blocker if the future plan was only to add a warning by default if the "required" sync wasn't configured in gatekeeper?

ritazh:
> To clarify a bit around the UX, with the `onlyServices` parameter, it seems syncing service is optional, however does this annotation make syncing service a requirement?

Per today's community call, the decision is to create a new template `K8sRequiredProbesWithServices` instead of bumping major version of the existing template as v1 is still valuable to maintain and should not be deprecated.

PR #289, opened 2023-01-31
chore: remove rouge space

apeabody:
Given no functional change, I didn't bump the version.  However if there is a concern we can increment the patch.


maxsmythe (approved):
LGTM

PR #296, opened 2023-02-27
allow exemptions for requiredprobes policy

zahodi:
This policy allows to define a list of containers that should be excluded from the requiredprobes policy.


sozercan:
@zahodi thanks for the PR! Looks like this is missing DCO. Please take a look at the dev guide for generating artifacts: https://github.com/open-policy-agent/gatekeeper-library#development

PR #298, opened 2023-03-03
add disallowed repos policy

sozercan:
- adds a disallowed repos example for scenarios such as https://kubernetes.io/blog/2023/02/06/k8s-gcr-io-freeze-announcement/
- add `generate-all` make target to simplify generation
- docker should run as user to avoid generation of library directory as root (for non-rootless)

Fixes #


JaydipGabani:
If I am not wrong, I think you also need to update [sidebar.js](https://github.com/open-policy-agent/gatekeeper-library/blob/master/website/sidebars.js#L45) and add a path for the policy to show up in website

sozercan:
@JaydipGabani good call. we should automate that https://github.com/open-policy-agent/gatekeeper-library/issues/299

maxsmythe (approved):
This is intended as the logical inverse of https://github.com/open-policy-agent/gatekeeper-library/tree/master/library/general/allowedrepos ?

LGTM

sozercan:
@maxsmythe that's right

src/general/disallowedrepos/src_test.rego
```diff
@@ -0,0 +1,132 @@
...
+            "name": "nginx"
+        },
+        "spec": {
+            "initContainers": containers,
+        }
+      }
+     }
+}
+
+input_container_disallowed = [
```
ritazh, line 102: nit: IMO this will read more clearly if it's `input_container_allowed`

PR #300, opened 2023-03-04
ci: remove skip for storage class

### LINKED ISSUE #284: remove storageclass and gk 3.8 check in CI
since GK 3.8 is no longer supported, we should be good to remove this check
https://github.com/open-policy-agent/gatekeeper-library/blob/master/test/bats/test.bats#L88

https://github.com/open-policy-agent/gatekeeper-library/blob/master/library/general/storageclass/README.md?plain=1#L14-L15

sozercan:

Fixes #284


artifacthub/library/general/storageclass/1.0.1/artifacthub-pkg.yml
```diff
@@ -1,7 +1,7 @@
 version: 1.0.1
 name: k8sstorageclass
 displayName: Storage Class
-createdAt: "2022-11-08T23:45:23Z"
```
ritazh, line 4: why does this need to be updated?

sozercan: not sure, i just ran generate. reverted it now, it doesn't regen the timestamp.

maxsmythe (approved):
LGTM

PR #303, opened 2023-03-13
allow exemptions for requiredprobes policy

zahodi:
This policy allows for definition of a list of containers that should be excluded from the requiredprobes policy.


zahodi:
@ritazh I need some help on this, when I run `make generate generate-website-docs generate-artifacthub-artifacts` my code gets removed. Am I following something not in order?

JaydipGabani:
@zahodi `make generate generate-website-docs generate-artifacthub-artifacts` generates files based on `src/general` dir, and since you are not updating root template for `requiredprobes` the command you are running overwrites your changes. You need to update [constraint.tmpl](https://github.com/open-policy-agent/gatekeeper-library/blob/master/src/general/requiredprobes/constraint.tmpl) with the changes you want to make. For rego changes, you might need to update [src.rego](https://github.com/open-policy-agent/gatekeeper-library/blob/master/src/general/requiredprobes/src.rego) and for `lib` you are trying to add here is an [examlpe template](https://github.com/open-policy-agent/gatekeeper-library/blob/master/src/general/containerlimits/constraint.tmpl#L43), you can follow similar pattern. 

Once you are done manking changes in `src/general/requiredprobes`, in your local delete the `artifacthub/library/general/requiredprobs/1.0.1` dir and generate it again using the aformentioned command

library/general/requiredprobes/template.yaml
```diff
@@ -29,32 +36,47 @@ spec:
     - target: admission.k8s.gatekeeper.sh
       rego: |
         package k8srequiredprobes
-
+        import data.lib.exempt_container.is_exempt
```
JaydipGabani, line 39: Update `src/general/requiredprobes/src.rego` with the rego changes you are trying to make

library/general/requiredprobes/template.yaml
```diff
@@ -29,32 +36,47 @@ spec:
...
         probe_field_empty(ctr, probe) = true {
             probe_fields := {field | ctr[probe][field]}
             diff_fields := probe_type_set - probe_fields
             count(diff_fields) == count(probe_type_set)
         }
-
         get_violation_message(container, review, probe) = msg {
             msg := sprintf("Container <%v> in your <%v> <%v> has no <%v>", [container.name, review.kind.kind, review.object.metadata.name, probe])
         }
+      libs:
```
JaydipGabani, line 64: add this in `src/general/requiredprobes/constraint.tmpl`

JaydipGabani (changes requested):

PR #310, opened 2023-04-06
fix proc-mount rego multiple output with same input

### LINKED ISSUE #311: proc-mount rego has potential bug
If parameter `procMount` is invalid, for example `invalid`, the function `get_allowed_proc_mount` will return different result with the same input.
https://github.com/open-policy-agent/gatekeeper-library/blob/master/library/pod-security-policy/proc-mount/template.yaml
`eval_conflict_error: functions must not produce multiple outputs for same inputs`

The issue might be avoid by enum parameter value, but it is still one potential bug in the rego langueage.

fseldow:
Rego bug if the parameter input is invalid, it will report `eval_conflict_error: functions must not produce multiple outputs for same inputs`
The situation might be avoid by parameter enum. The pr here is to improve the rego to avoid such situation

Fixes #311


ritazh:
Please run 'make generate generate-website-docs generate-artifacthub-artifacts' to generate the templates and docs

maxsmythe (approved):
LGTM, thank you!

maxsmythe:
It looks like this command needs to be run: `Please run 'make generate generate-website-docs generate-artifacthub-artifacts' to generate the templates and docs`

ritazh (approved):
LGTM

PR #312, opened 2023-04-06
feat: add allowedStorageClasses to K8sStorageClass

apeabody:
Adds an optional `allowedStorageClasses` parameter to `K8sStorageClass`.  When `allowedStorageClasses`  is omitted the behavior is a backwards compatible.


artifacthub/library/general/storageclass/1.1.0/template.yaml
```diff
@@ -0,0 +1,155 @@
...
+        violation[{"msg": pvc_storageclass_noname_msg}] {
+          is_pvc(input.review.object)
+          not input.review.object.spec.storageClassName
+        }
+        pvc_storageclass_noname_msg := sprintf("pvc did not specify a storage class name. Must be one of [%v]", args) {
+          input.parameters.includeStorageClassesInMessage
+          args := [
+            concat(", ", [n | data.inventory.cluster["storage.k8s.io/v1"]["StorageClass"][n]])
+          ]
+        } else := sprintf(
```
maxsmythe, line 103: nit: does the syntax allow just defining a string constant here?

artifacthub/library/general/storageclass/1.1.0/template.yaml
```diff
@@ -0,0 +1,155 @@
...
+          "pvc did not specify a storage class name.",
+          []
+        )
+
+        violation[{"msg": statefulset_vct_badname_msg(vct)}] {
+          is_statefulset(input.review.object)
```
maxsmythe, line 109: Why statefulset vs. pods?

apeabody: Hey @maxsmythe - This was the original design/code (#214), and my (initial) goal was to minimize changes to ensure backwards compatibility.

This CT also applies to `PersistentVolumeClaim`, so it would cover pods from that angle.  My assumption is `StatefulSet` was included to catch earlier in the stack/block on admission.

artifacthub/library/general/storageclass/1.1.0/template.yaml
apeabody: Hey @maxsmythe 

This was existing code, so I was also a bit surprised by the `variable := string { } else := "string"` syntax, but it appears to work as expected and match the Rego doc: https://www.openpolicyagent.org/docs/latest/policy-language/#else-keyword

artifacthub/library/general/storageclass/1.1.0/template.yaml
maxsmythe: Ah, it looks like stateful sets specifically have volume claim templates.

maxsmythe (approved):
LGTM

apeabody:
Hi @ritazh - Only when you have time, please let me know if anything else is required on this change.  Thanks!

sozercan (approved):
LGTM

Issue #299, opened 2023-03-04
website: automate generation of sidebar


sozercan:
sidebar for the gk-library website is not generated, we should autogenerate to create the sidebar

nilekhc:
fixed with https://github.com/open-policy-agent/gatekeeper-library/pull/319

PR #319, opened 2023-04-26
feat: adds website directory structure validation

### LINKED ISSUE #314: New policies are added to the wrong path on the website
https://github.com/open-policy-agent/gatekeeper-library/tree/master/website/docs

ephemeralstoragelimit and horizontalpodautoscaler should be added to validation not root

nilekhc:
This PR adds the validation script which currently validates the directory structure at `website/docs`. This (and any future validations) aims to prevent accidental file additions to the wrong paths.

Fixes #314 #299 


website/docs/test.md
apeabody, line 1: Is this temporary to test the CI?

nilekhc: yeah. I removed it

scripts/validate/validate.go
```diff
@@ -0,0 +1,82 @@
+package main
+
+import (
+	"fmt"
+	"io/ioutil"
```
sozercan, line 5: i think this is deprecated

scripts/validate/validate.go
```diff
@@ -0,0 +1,82 @@
+package main
```
sozercan, line 1: should we combine all these into a single script?

website/docs/ephemeralstoragelimit.md
```diff
@@ -1,374 +0,0 @@
----
```
sozercan, line 1: curious, what was the issue? this was added to wrong path manually?

.github/workflows/workflow.yaml
```diff
@@ -19,6 +19,10 @@ jobs:
         run: |
           make generate generate-website-docs generate-artifacthub-artifacts
           git diff --exit-code || (echo "Please run 'make generate generate-website-docs generate-artifacthub-artifacts' to generate the templates and docs" && exit 1)
+      - name: Validation
+        run: |
+          make validate
+          git diff --exit-code || (echo "Please run 'make validate' to ensure validations are passing" && exit 1)
```
sozercan, line 25: why do we need "git diff"? isn't this going to panic?

scripts/validate/validate.go
```diff
@@ -0,0 +1,82 @@
...
+	files, err := ioutil.ReadDir(docsDirPath)
+	if err != nil {
+		return err
+	}
+
+	// Validate the directory structure
+	for _, file := range files {
+		if file.IsDir() {
+			if !contains(expectedDirs, file.Name()) {
+				err = fmt.Errorf("unexpected directory: %s, found at : %s", file.Name(), filepath.Join(docsDirPath, file.Name()))
```
sozercan, line 46:
```suggestion
				err = fmt.Errorf("unexpected directory: %s, found at: %s", file.Name(), filepath.Join(docsDirPath, file.Name()))
```

scripts/validate/validate_test.go
```diff
@@ -0,0 +1,113 @@
...
+			name:     "Item not in list",
+			items:    []string{"item1", "item2", "item3"},
+			item:     "item4",
+			expected: false,
+		},
+		{
+			name:   "list is empty",
+			items:  []string{},
+			item:   "foo",
+			expected	: false,
```
sozercan, line 101: looks like this needs formatting, can you run a linter on this?

scripts/validate/validate.go
nilekhc: I thought about it. But we want to run those independently of 'generate script'. We use generate script for website docs, artifact hub, etc. (which runs independently). So the idea is to run the validation script separately once we generate all the docs. Also, this script can extend to validate other aspects of the repo and can run as an independent validator.

Makefile
```diff
@@ -77,5 +78,9 @@ generate-website-docs:
 generate-artifacthub-artifacts:
 	cd $(ARTIFACTHUB_SCRIPT_DIR); go test -v && go run hub.go
 
+.PHONY: validate
+validate:
+	cd $(VALIDATE_SCRIPT_DIR); go test -v & go run validate.go
```
sozercan, line 83:
```suggestion
	cd $(VALIDATE_SCRIPT_DIR); go test -v && go run validate.go
```

website/docs/ephemeralstoragelimit.md
nilekhc: Yeah. That's my guess. It's hard to validate it using commit history. But someone unsure how we generate docs is likely to do it.

Makefile
sozercan: this is kinda confusing, since unit tests will pass but "go run" will panic

nilekhc: It's possible, but I prefer to run unit tests before proceeding with the validation code to ensure it's fully tested beforehand, especially since we're not building anything.

Is there a better way to test it before `run`ning it?

scripts/validate/validate_test.go
```diff
@@ -0,0 +1,113 @@
+package main
+
+import (
+	"io/ioutil"
```
sozercan, line 4: update ioutil here too

Makefile
sozercan: unit tests can be in a separate/parallel workflow then execution so we don't have wait for them.

sozercan: can we separate out a "test" make target and run that in CI?

nilekhc: makes sense. I added a separate stage for testing scripts.

ritazh:
I do not see ephemeralstoragelimit and hpa in the left navigation under validation/general today

.github/workflows/workflow.yaml
```diff
@@ -15,10 +15,16 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
+      - name: Test scripts
```
sozercan, line 18: can we make "test" into a new workflow so they can run in parallel?

Makefile
```diff
@@ -73,9 +74,21 @@ require-sync:
 generate-website-docs:
 	cd $(WEBSITE_SCRIPT_DIR); go run generate.go
 
+.PHONY: test-artifacthub-script
+test-artifacthub-script:
```
sozercan, line 78: can we just call this "make test" and have all tests run there? is there a reason for each of these to have their own make target?

website/docs/ephemeralstoragelimit.md
apeabody: @nilekhc - Agreed, I wasn't able to replicate the issue and don't see how this could have resulted in `docs` directly? https://github.com/open-policy-agent/gatekeeper-library/blob/master/scripts/website/generate.go#L152

nilekhc: Thanks for validating @apeabody 😄

.github/workflows/workflow.yaml
```diff
@@ -11,6 +11,14 @@ on:
       - "website/**"
     branches: [master]
 jobs:
+  website_script_unit_test:
+    runs-on: ubuntu-latest
+    name: "Test scripts"
+    steps:
+      - uses: actions/checkout@v2
```
apeabody, line 18: Just curious, is there a need to stay with v2?  I can't recall if v2 is using the Node 12 runtime which is deprecated later this year.

Alternatively we could do a follow-up issue/PR bumping all `actions/checkout`

```suggestion
      - uses: actions/checkout@v3
```

README.md
```diff
@@ -80,6 +80,7 @@ Each new policy should contain:
 * `make generate-website-docs` will generate the markdown files required for the website.
 * `make generate-artifacthub-artifacts` will generate or update the artifact hub packages and associated `artifacthub-pkg.yml` file under `/artifacthub` directory.
 * `make generate-all` will generate all artifacts above.
+* `make validate` will run validation checks on the library repo. Currently it validates directory structure of `website/docs` directory.
```
apeabody, line 83: It is covered by the repo CI, but do we want to include `make unit-test`? The `generate-artifacthub-artifacts` step used in include the `cd $(ARTIFACTHUB_SCRIPT_DIR); go test` item which was moved under the new `make unit-test`.

apeabody (approved):
@nilekhc LGTM! Added two optional questions/suggestions.

.github/workflows/workflow.yaml
nilekhc: updated with sha

sozercan (approved):
Thanks! LGTM

.github/workflows/workflow.yaml
```diff
@@ -64,7 +75,7 @@ jobs:
           kubectl logs -n gatekeeper-system -l control-plane=audit-controller --tail=-1 > logs-audit.json
 
       - name: Upload artifacts
-        uses: actions/upload-artifact@v2
+        uses: actions/checkout@8e5e7e5ab8b370d6c329ec480221332ada57f0ab
```
apeabody, line 78: @nilekhc Just wanted to make sure this change was intentional?

nilekhc: It is. It's always better to pin to the sha version. I have also opened https://github.com/open-policy-agent/gatekeeper-library/pull/322 to automatically update it.

apeabody: @nilekhc Sorry, to clarify I meant the changing action from `upload-artifact` to `checkout`?  No concerns on using the SHA instead of tag, that is very common.

nilekhc: Updated. Thanks.

nilekhc:
@apeabody I have addressed your comments. PTAL.

apeabody (approved):
LGTM @nilekhc!

PR #332, opened 2023-05-04
feat: add Gatekeeper 3.12 to CI

apeabody:
Add CI test coverage for Gatekeeper 3.12

### DIFF
--- a/.github/workflows/workflow.yaml
+++ b/.github/workflows/workflow.yaml
@@ -65,7 +65,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        gatekeeper: [ "release-3.10", "release-3.11" ]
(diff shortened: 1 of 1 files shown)


.github/workflows/workflow.yaml
```diff
@@ -65,7 +65,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        gatekeeper: [ "release-3.10", "release-3.11" ]
+        gatekeeper: [ "release-3.10", "release-3.11", "release-3.12" ]
```
sozercan, line 68: we should be okay to drop 3.10 since we support n-1 https://github.com/open-policy-agent/gatekeeper/blob/master/docs/Release_Management.md#supported-releases

apeabody: SG - Thanks for the link!

sozercan (approved):

PR #339, opened 2023-05-04
fix: fixes version check for Artifact Hub generation

### LINKED ISSUE #331: fix version check in Artifact Hub generation
Ref - https://github.com/open-policy-agent/gatekeeper-library/pull/305#issuecomment-1507529116

nilekhc:

Currently, we are checking the template hash from a local file. This requires a constant version upgrade when we add new template files and incrementally update it before merging. This is incorrect since for a new file, the version would start at `1.0.0`.

With this fix, we are now validating the hash of the template file from GitHub to the hash of the local template file. So when a contributor is trying to create a new template file, it will skip the version check and perform the same for an existing template file for a potential version bump.

Fixes #331 

### DIFF
--- a/scripts/artifacthub/hub.go
+++ b/scripts/artifacthub/hub.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"io"
 	"io/fs"
+	"net/http"
 	"os"
 	"path"
 	"path/filepath"
@@ -72,6 +73,19 @@ type ArtifactHubMetadata struct {
 	} `yaml:"annotations,omitempty"`
 }
 
+// HTTPClient is an interface that matches the methods needed from http.Client.
+type HTTPClient interface {
+	Get(url string) (*http.Response, error)
+}
+
+// DefaultClient is a default implementation of the HTTPClient interface.
+type DefaultClient struct{}
+
+// Get is a method of DefaultClient that makes the actual HTTP GET request.
+func (c DefaultClient) Get(url string) (*http.Response, error) {
+	return http.Get(url)
+}
+
 const (
 	// entryPoint is the directory entry point for artifact hub
 	ahEntryPoint = "artifacthub"
@@ -123,9 +137,11 @@ func main() {
 						panic(err)
 					}
 
+					githubSourceRelativePath := filepath.Join(entryPoint, entry.Name(), dir.Name(), "template.yaml")
 					createVersionDirectory(
 						rootDir,
 						filepath.Join(entryPoint, entry.Name(), dir.Name()),
+						githubSourceRelativePath,
 						constraintTemplate,
 					)
 				}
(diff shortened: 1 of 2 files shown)


scripts/artifacthub/hub.go
```diff
@@ -221,11 +224,38 @@ func addArtifactHubMetadata(sourceDirectory, destinationPath, ahBasePath string,
...
+	// compare hash with template.yaml in github
+	githubTemplateURL := sourceURL + githubSourceRelativePath
+	resp, err := http.Get(githubTemplateURL)
+	if err != nil {
+		fmt.Println("error while getting constraint template from github")
+		panic(err)
+	}
+	if resp.StatusCode == http.StatusNotFound {
+		fmt.Printf("constraint template %s not found in github. It is likely that constraint template is being updated locally and not merged to github yet.\n", githubSourceRelativePath)
+
```
apeabody, line 237:
```suggestion
```

scripts/artifacthub/hub.go
```diff
@@ -221,11 +224,38 @@ func addArtifactHubMetadata(sourceDirectory, destinationPath, ahBasePath string,
...
+	githubTemplateURL := sourceURL + githubSourceRelativePath
+	resp, err := http.Get(githubTemplateURL)
+	if err != nil {
+		fmt.Println("error while getting constraint template from github")
+		panic(err)
+	}
+	if resp.StatusCode == http.StatusNotFound {
+		fmt.Printf("constraint template %s not found in github. It is likely that constraint template is being updated locally and not merged to github yet.\n", githubSourceRelativePath)
+		
+		return nil
```
apeabody, line 238: This test probably needs to be updated? https://github.com/open-policy-agent/gatekeeper-library/blob/master/scripts/artifacthub/hub_test.go#LL184C24-L184C24

apeabody (changes requested):
Thanks @nilekhc!

I think `hub_test.go` maybe missing when an expected error does not occur (`err == nil && tc.expectedError`)? https://github.com/open-policy-agent/gatekeeper-library/blob/master/scripts/artifacthub/hub_test.go#L215

scripts/artifacthub/hub.go
nilekhc: I usually put the `return` on the new line for better code readability. LMK if you have any opinions about the same.

scripts/artifacthub/hub.go
nilekhc: I updated the entire test method.

nilekhc:
> Thanks @nilekhc!
> 
> I think `hub_test.go` maybe missing when an expected error does not occur (`err == nil && tc.expectedError`)? https://github.com/open-policy-agent/gatekeeper-library/blob/master/scripts/artifacthub/hub_test.go#L215

@apeabody I updated the entire test method. PTAL.

apeabody (approved):
Thanks @nilekhc LGTM!

PR #348, opened 2023-05-23
Exclude `UPDATE` operations in constraints for immutable fields

### LINKED ISSUE #346: Apply constraints for immutable fields only to CREATE operations
## Problem

Constraint templates for immutable fields (e.g. `K8sPSPHostFilesystem` validates pods' volumes) are enforced for `UPDATE` operations when the Gatekeeper's validating admission webhook is configured to be applied to `UPDATE` operation.
This can cause the following situation:

1. Assume that a `K8sPSPHostFilesystem` constraint with `allowedHostPaths: [{ pathPrefix: "/app" }]` is on a cluster
2. A user creates a job, which creates a pod with `/app` host path volume
3. An administrator removes `/app` path prefix from `allowedHostPaths` of the `K8sPSPHostFilesystem` constraint
4. The user deletes the job, which in turn makes the job controller tries to updates the pod's finalizers
5. The pod's finalizers cannot be updated because the pod has `/app` host path volume and it now voilates the `K8sPSPHostFilesystem` constraint
6. The `/app` host path volume also cannot be removed from the pod, because pods' volume config is immutable.
7. As a result, the pod stuck in Terminating phase

## Possible solution 1

Configuring the Gatekeeper's validating webhook to be applied only to `CREATE` operation.

This solution cannot be adopted if a cluster has constrains for mutable fields (e.g. `K8sRequiredLabels`).

## Possible soultion 2

Modifying constraints templates for immutable fields to be applied only to `CREATE` operation.

This solution can be implemented by adding a `input.review.operation` condition to constraint definitions written in Rego.

What do you think?

ordovicia:
**What this PR does / why we need it**: This PR will fix a problem described in #346, by excluding `UPDATE` operations in constraint templates for immutable fields.

Fixes #346 

**Special notes for your reviewer**: None.


ordovicia:
Hi @apeabody @maxsmythe , could you take a look? Thank you!

apeabody (commented):
Thanks for the contribution @ordovicia!

A few thoughts/questions:
- Just confirming we want/need to exclude `PATCH`?  The default validating webhook verbs are `CREATE` and `UPDATE`: https://github.com/open-policy-agent/gatekeeper/blob/v3.12.0/deploy/gatekeeper.yaml#L3747 | @maxsmythe, any caveats to `PATCH`?
- Thanks for the rego/opa test for `lib.exclude_update_patch`!  We should probably also include some tests for the templates (and maybe even constraints):
  - For the templates you can add `"operation": "UPDATE"` to the `"review": {}`, for example: https://github.com/open-policy-agent/gatekeeper-library/blob/master/src/general/automount-serviceaccount-token/src_test.rego#L10
  - For the constraints you could use a [`AdmissionReview` object](https://open-policy-agent.github.io/gatekeeper/website/docs/gator#usage-1) to include `request.operation: UPDATE` for the test suite.

maxsmythe:
PATCH is probably irrelevant. From the K8s docs:

```
Operations is the operations the admission hook cares about - CREATE, UPDATE, DELETE, CONNECT or * for all of those operations and any future admission operations that are added. If '\*' is present, the length of the slice must be one. Required.
```

https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.27/#rulewithoperations-v1-admissionregistration-k8s-io

ordovicia:
Thank you for the comments!

Fixed to exclude `UPDATE` operation only.
Also added tests for constraint templates.
But I could not find how to add `UPDATE` operation tests for constraints.

apeabody:
> Thank you for the comments!
> 
> Fixed to exclude `UPDATE` operation only. Also added tests for constraint templates. But I could not find how to add `UPDATE` operation tests for constraints.

Thanks @ordovicia!  You *should* be able to use a [`AdmissionReview`](https://open-policy-agent.github.io/gatekeeper/website/docs/gator#usage-1) object in the test suite to include the operation type.  For example with `K8sPSPAutomountServiceAccountTokenPod` you could add a new allowed similar to the existing disallowed but allowed for the `UPDATE` operation [test](https://github.com/open-policy-agent/gatekeeper-library/blob/master/library/general/automount-serviceaccount-token/suite.yaml#L15).  Instead of the using the [`Pod`](https://github.com/open-policy-agent/gatekeeper-library/blob/master/library/general/automount-serviceaccount-token/samples/automount-serviceaccount-token/example_disallowed.yaml) object directly, for `object:` you would use an `AdmissionReview` object similar to:

```
kind: AdmissionReview
apiVersion: admission.k8s.io/v1beta1
request:
  operation: "UPDATE"
  object:
    apiVersion: v1
    kind: Pod
    metadata:
      name: nginx-automountserviceaccounttoken-update-allowed
      labels:
        app: nginx-automountserviceaccounttoken
    spec:
      automountServiceAccountToken: true
      containers:
        - name: nginx
          image: nginx
```

The idea being this `Pod` should normally raise a violation, but not if the operation is "UPDATE".

ordovicia:
Hi @apeabody @maxsmythe , could you take a look again?

src/rego/lib_exclude_update/lib_exclude_update.rego
```diff
@@ -0,0 +1,7 @@
+package lib.exclude_update
+
+import future.keywords.in
```
apeabody, line 3: nit: If we aren't using `in` we should probably drop from all `lib.exclude_update`?

website/docs/validation/volumes.md
```diff
@@ -182,6 +196,49 @@ Usage
 kubectl apply -f https://raw.githubusercontent.com/open-policy-agent/gatekeeper-library/master/library/pod-security-policy/volumes/samples/psp-volume-types/example_allowed.yaml
 ```
 
+</details>
+<details>
```
apeabody, line 200: NOTE ONLY: Opened https://github.com/open-policy-agent/gatekeeper-library/issues/400 as we might want to exclude `AdmissionReview` from being used as documentation samples.

apeabody (commented):
Thanks @ordovicia!

One question/comment on `lib.exclude_update`.

Cheers!

src/rego/lib_exclude_update/lib_exclude_update.rego
ordovicia: Good catch! I removed all `future.keywords.in` import.

ordovicia:
Hi @apeabody please merge this if it's OK. Thank you!

library/general/ephemeralstoragelimit/template.yaml
```diff
@@ -150,10 +151,14 @@ spec:
...
 
         violation[{"msg": msg}] {
+          # spec.containers.resources.limits["ephemeral-storage"] field is immutable.
+          not is_update(input.review)
+
           general_violation[{"msg": msg, "field": "containers"}]
         }
 
         violation[{"msg": msg}] {
+          not is_update(input.review)
```
apeabody, line 161: Might want to include a comment similar to above.

library/general/ephemeralstoragelimit/suite.yaml
```diff
@@ -27,3 +27,7 @@ tests:
     object: samples/container-must-have-ephemeral-storage-limit/example_disallowed_ephemeral_storage_limit_1Pi-initContainer.yaml
     assertions:
     - violations: yes
+  - name: ephemeral-storage-limit-update
+    object: samples/container-must-have-ephemeral-storage-limit/update.yaml
+    assertions:
+    - violations: no
```
apeabody, line 33: nit:

```suggestion
    - violations: no

```

library/pod-security-policy/capabilities/suite.yaml
```diff
@@ -19,3 +19,7 @@ tests:
         object: samples/capabilities-demo/disallowed_ephemeral.yaml
         assertions:
           - violations: yes
+      - name: update
+        object: samples/capabilities-demo/update.yaml
+        assertions:
+          - violations: no
```
apeabody, line 25: nit:
```suggestion
          - violations: no

```

apeabody (approved):
Thanks for the contribution @ordovicia!

Another maintainer will need to review as well.

ritazh (approved):
LGTM

Thanks for the fix @ordovicia! 🎉  You might want to update the PR description to remove `patch`

Issue #345, opened 2023-05-19
Migrate require-sync CI to future gatekeeper 3.13 requires-sync-data unmarshal function


apeabody:
https://github.com/open-policy-agent/gatekeeper/pull/2734

PR #355, opened 2023-06-12
feat: update template and CI to requires-sync-data

apeabody:
Migrate applicable templates and CI from `requiresSyncData` to `requires-sync-data` per https://github.com/open-policy-agent/gatekeeper/pull/2734


maxsmythe (approved):
LGTM

apeabody:
Hi @anlandu & @sozercan - This will update Gatekeeper-Library to match https://github.com/open-policy-agent/gatekeeper/pull/2734

anlandu (approved):

.github/workflows/workflow.yaml
```diff
@@ -114,7 +114,7 @@ jobs:
           make require-suites
   require_sync:
     runs-on: ubuntu-latest
-    name: "Require a sync.yaml file and metadata.gatekeeper.sh/requiresSyncData annotation for every template.yaml using data.inventory"
+    name: "Require a sync.yaml file and metadata.gatekeeper.sh/requires-sync-data annotation for every template.yaml using data.inventory"
```
ritazh, line 117: How does the new annotation work with older GK versions since `metadata.gatekeeper.sh/requires-sync-data` is only introduced after https://github.com/open-policy-agent/gatekeeper/releases/tag/v3.13.0-beta.1? and how does the old annotation work with newer GK versions?

ritazh: I see CI is not broken even tho the GK versions we are testing with do not know about the new annotation.

apeabody: Hi @ritazh~ - The `requiresSyncData` annotation was introduced to gatekeeper-library with a "synthetic" require-suites test, but the annotation is not (yet~) used by GK/CI Tests.  This change is to have the gatekeeper-library repo ready in advance of GK 3.13.

I have https://github.com/open-policy-agent/gatekeeper-library/issues/345 open to use/test the future GK 3.13 `requiresSyncData` annotation unmarshall.

ritazh (approved):
LGTM

PR #363, opened 2023-06-26
feat: add exemptImages to k8srequiredlabels & k8srequiredannotations

apeabody:
**What this PR does / why we need it**:  add `exemptImages` parameter to `k8srequiredlabels` & `k8srequiredannotations`


maxsmythe (commented):
What is the purpose of exempting metadata-scoped policies (labels, annotations) based off of an image used?

I have a few concerns:

1) This constraint template is used for more than just pods, so using exempt images may not make sense.

2) If there is a CRD that uses `x-kubernetes-preserve-unknown-fields` a malicious user could leverage that to bypass the constraint for the CR, even though the field technically has no effect. 

3) container-level exceptions make the most sense when it's the individual container being excepted. A prime example of this is sidecar injection -- if istio's sidecar needs host networking access, only that sidecar should be exempted from the check, not the main pod image. Because labels/annotations are global for the resource, the existence of one container that triggers an exemption also letting other containers tag along seems like it could create potential for abuse.

apeabody:
> What is the purpose of exempting metadata-scoped policies (labels, annotations) based off of an image used?
> 
> I have a few concerns:
> 
> 1. This constraint template is used for more than just pods, so using exempt images may not make sense for all use cases.
> 2. If there is a CRD that uses `x-kubernetes-preserve-unknown-fields` a malicious user could leverage that to bypass the constraint for the CR, even though the field technically has no effect.
> 3. container-level exceptions make the most sense when it's the individual container being excepted. A prime example of this is sidecar injection -- if istio's sidecar needs host networking access, only that sidecar should be exempted from the check, not the main pod image. Because labels/annotations are global for the resource, the existence of one container that triggers an exemption also letting other containers tag along seems like it could create potential for abuse.

Hi @maxsmythe - Let me connect with you directly on the use-case.

apeabody:
> > What is the purpose of exempting metadata-scoped policies (labels, annotations) based off of an image used?
> > I have a few concerns:
> > 
> > 1. This constraint template is used for more than just pods, so using exempt images may not make sense for all use cases.
> > 2. If there is a CRD that uses `x-kubernetes-preserve-unknown-fields` a malicious user could leverage that to bypass the constraint for the CR, even though the field technically has no effect.
> > 3. container-level exceptions make the most sense when it's the individual container being excepted. A prime example of this is sidecar injection -- if istio's sidecar needs host networking access, only that sidecar should be exempted from the check, not the main pod image. Because labels/annotations are global for the resource, the existence of one container that triggers an exemption also letting other containers tag along seems like it could create potential for abuse.
> 
> Hi @maxsmythe - Let me connect with you directly on the use-case.

Thanks for the feedback @maxsmythe!  We'll try to accomplish using a different `match` strategy.

PR #364, opened 2023-06-27
fix: drop excludedNamespaces from sample

apeabody:
**What this PR does / why we need it**: drop `excludedNamespaces` from sample constraint for `K8sBlockLoadBalancer`


maxsmythe (approved):
LGTM

apeabody:
> The change LGTM! Do we need a version bump since it does technically change the behavior of the sample constraint? We should update this version update guidance for changes introduced in the constraint once we made a decision here.
> 
> > Patch version bump required: Whenever there is a simple backward compatible change in the policy, e.g. Simple Rego fix, updating policy metadata

Yeah, my read/thought is this is specific to the Templates themselves, but the sample constraint is published to artifacthub.  A slight concern with bumping the Temple version is it might cause end user confusion when there is no actual Template change, however if changes only to sample constraint are limited to patch version bumps that might be minimized.

@nilekhc WDYT?  Currently the artifacthub CI only considers the template itself for version enforcement.

ritazh:
> > The change LGTM! Do we need a version bump since it does technically change the behavior of the sample constraint? We should update this version update guidance for changes introduced in the constraint once we made a decision here.
> > > Patch version bump required: Whenever there is a simple backward compatible change in the policy, e.g. Simple Rego fix, updating policy metadata
> 
> Yeah, my read/thought is this is specific to the Templates themselves, but the sample constraint is published to artifacthub. A slight concern with bumping the Temple version is it might cause end user confusion when there is no actual Template change, however if changes only to sample constraint are limited to patch version bumps that might be minimized.
> 
> @nilekhc WDYT? Currently the artifacthub CI only considers the template itself for version enforcement.

@nilekhc thoughts?

nilekhc:
> > > The change LGTM! Do we need a version bump since it does technically change the behavior of the sample constraint? We should update this version update guidance for changes introduced in the constraint once we made a decision here.
> > > > Patch version bump required: Whenever there is a simple backward compatible change in the policy, e.g. Simple Rego fix, updating policy metadata
> > 
> > 
> > Yeah, my read/thought is this is specific to the Templates themselves, but the sample constraint is published to artifacthub. A slight concern with bumping the Temple version is it might cause end user confusion when there is no actual Template change, however if changes only to sample constraint are limited to patch version bumps that might be minimized.
> > @nilekhc WDYT? Currently the artifacthub CI only considers the template itself for version enforcement.
> 
> @nilekhc thoughts?

Version chance in the Artifacthub CI is only to enforce version bumping when the template changes. My main concern here is, we [update the AH package](https://github.com/open-policy-agent/gatekeeper-library/blob/0a14d2ba9d104eceb2688310ed6c99c2cc78f6d2/scripts/artifacthub/hub.go#L224-L225) only when the template changes. So even if we make this change, they won't reflect on AH. Should we update our logic for when to trigger AH refresh? If so what all files should we consider?

ritazh (approved):
LGTM

sozercan (approved):

PR #381, opened 2023-08-01
chore: update docusaurus v2.4.1 and use only yarn lockfile

apeabody:
* Update Docusaurus to v2.4.1
* Use only `yarn.lock` for lockfile (should help with dependabot)


maxsmythe (approved):
LGTM

sozercan (approved):

PR #396, opened 2023-09-05
chore: require-sync gk update

acpana:
`go get github.com/open-policy-agent/gatekeeper/v3@366ba79`

But probably want to wait on gatekeeper 3.14 before making this change

`parser` package was moved as per feedback from https://github.com/open-policy-agent/gatekeeper/pull/2852


scripts/require-sync/main.go
```diff
@@ -17,7 +17,7 @@ import (
 	"github.com/open-policy-agent/frameworks/constraint/pkg/core/templates"
 	gkapis "github.com/open-policy-agent/gatekeeper/v3/apis"
 	"github.com/open-policy-agent/gatekeeper/v3/pkg/gator/reader"
-	"github.com/open-policy-agent/gatekeeper/v3/pkg/syncutil/parser"
+	"github.com/open-policy-agent/gatekeeper/v3/pkg/cachemanager/parser"
```
apeabody, line 20: Looks like the change in path name results in a different import ordering.  You can fix automatically with `golangci-lint run main.go --fix` or `gofumpt -w main.go`.

apeabody (changes requested):
Thanks for the contribution @acpana!

For the failing CI: while you could add the updated `go.work.sum`, we don't really need to include that file in the repository as it's automatically generated.  I recommend you delete https://github.com/acpana/gatekeeper-library/blob/acpana/update-gk/go.work.sum and add `go.work.sum` to https://github.com/acpana/gatekeeper-library/blob/acpana/update-gk/.gitignore

I also included a note below on some easy ways to fix the `main.go` lint check.

Let me know if you need any assistance, cheers!

apeabody (approved):
Thanks @acpana! LGTM

PR #409, opened 2023-09-25
chore: group docusaurus npm updates

apeabody:
**What this PR does / why we need it**: We should updates all the docusaurus modules at once.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -73,6 +73,10 @@ updates:
         update-types:
         - "version-update:semver-major"
         - "version-update:semver-minor"
+    groups:
+      all:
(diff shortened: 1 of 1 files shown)


sozercan (approved):

sozercan:
in general, this is fine but some deps might break the website if upgraded.

apeabody:
> in general, this is fine but some deps might break the website if upgraded.

@sozercan - Agreed, although at least with a single PR it should be easier to test (or replace with a custom PR) as the docusaurus component versions should always be in-sync.

https://docusaurus.io/docs/installation#updating-your-docusaurus-version

PR #411, opened 2023-09-27
Seccomp mutation for both notations

alexarefev:

Currently, `Seccomp` mutation rule work only for pod annotation. `Seccomp` mutation for Pod `spec.securityContext.seccompProfile` should be supported.  

**Which issue(s) does this PR fix**:
Fixes #


website/docs/mutation-examples/seccomp.md
```diff
@@ -7,23 +7,30 @@ title: seccomp
 
 ### Usage
 ```shell
-kubectl apply -f https://raw.githubusercontent.com/open-policy-agent/gatekeeper-library/master/mutation/pod-security-policy/seccomp/samples/mutation.yaml
```
apeabody, line 10: Hmm, looks like we are missing `mutation-annotation`.

@nilekhc Looks like the website generator currently will only retain the final mutation sample per directory? https://github.com/open-policy-agent/gatekeeper-library/blob/master/scripts/website/generate.go#L226

PR #420, opened 2023-10-20
Fix replicalimits ConstraintTemplate to handle scaling to zero

### LINKED ISSUE #419: replicalimits unit tests do not include checks for Scale resources
The `general/replicalimits` ConstraintTemplate in the library only has unit tests that validate updates to a Deployment resource.  It does not check for Scale resources, which get generated when running `kubectl scale` commands.

There is also a bug in the Rego because in the event of `kubectl scale deploy <name> --replicas=0`, the Scale resource has an empty spec (so `input.review.object.spec` is empty) and a violation is generated, even if a range allows zero as a valid value.

skaven81:
The `replicalimits` ConstraintTemplate in the library had no unit tests for validating scaling to zero.

It turns out that when scaling to zero, Kubernetes generates a `Scale` resource with an empty `spec`, instead of `spec.replicas=0`.

The unit tests were also missing tests using the `Scale` resource.

This PR fixes both issues.

**Which issue(s) does this PR fix** 
Fixes #419


skaven81:
Additional unit tests confirm that the updated template works as expected.
```=== RUN   replica-limit
    === RUN   example-allowed
    --- PASS: example-allowed	(0.005s)
    === RUN   example-scale-allowed
    --- PASS: example-scale-allowed	(0.006s)
    === RUN   example-disallowed
    --- PASS: example-disallowed	(0.004s)
    === RUN   example-scale-disallowed
    --- PASS: example-scale-disallowed	(0.005s)
--- PASS: replica-limit	(0.027s)
=== RUN   replica-limit-zero
    === RUN   example-allowed
    --- PASS: example-allowed	(0.008s)
    === RUN   example-scale-allowed
    --- PASS: example-scale-allowed	(0.006s)
    === RUN   example-disallowed
    --- PASS: example-disallowed	(0.006s)
    === RUN   example-scale-disallowed
    --- PASS: example-scale-disallowed	(0.005s)
--- PASS: replica-limit-zero	(0.031s)
ok	suite.yaml	0.058s
PASS
```

apeabody (changes requested):
Thanks for the contribution @skaven81!

You will also need to make your template changes to https://github.com/open-policy-agent/gatekeeper-library/blob/master/src/general/replicalimits/constraint.tmpl and https://github.com/open-policy-agent/gatekeeper-library/blob/master/src/general/replicalimits/src.rego.  The artifacts (e.g. `template.yaml` can then be built with `make generate-all`.

skaven81:
I'm not sure what I borked here, but I'll close this and try again with a fresh branch

skaven81:
Opened #427 with corrected branch contents

PR #429, opened 2023-10-30
Ensure OPA strict mode compliance

anderseknert:
Hi Gatekeeper friends! 👋😃

In order to ensure the Gatekeeper library works with future updates of OPA, and to better catch some common mistakes in Rego policy, this PR aims to fix all errors reported from `opa check --strict src`, and to make sure future changes are compliant with the OPA [strict mode](https://www.openpolicyagent.org/docs/latest/policy-language/#strict-mode) checks.

The errors raised from strict mode, and which are fixed with this PR, include:

* `input` must not be shadowed/assigned
* Unused function args — use wildcards (`_`) in their place
* Unused local assignment
* Deprecated built-in functions
  * `re_match` -> `regex.match`
  * `all` -> these were not needed
  * `any` -> `strings.any_prefix_match` and `strings.any_suffix_match`
     as these were always used together with `startswith`/`endswith`

The by far most noisy fix was the first one, i.e. `input` getting shadowed in basically all tests. Changed to use `inp` for a name as `in` is of course another name we don't want to shadow 😅

The changes here do not change semantics of the code. I did however also make some additional fixes where e.g. the unused function args error showed that a function wasn't needed in the first place, and so on. Naturally, all tests pass as before.

Two things I'm not sure about:

1. The code introduces the `strings.any_prefix_match` and `strings.any_suffix_match` built-in functions in one policy. These were introduced in OPA v0.44.0, which is quite a few OPA and Gatekeeper versions ago. I could not find information on what the minimum version of OPA or Gatekeeper must be supported in the Rego code in this library. If needed, that change can be reverted.
2. https://github.com/open-policy-agent/gatekeeper-library mentions Versioning. Should I update the versions of each policy changed with this PR?

To make sure the code stays compliant, I've added the `opa check --strict` command to execute with the `test.sh` script, before the unit tests run.


anderseknert:
```
Generating artifact hub content for  /home/runner/work/gatekeeper-library/gatekeeper-library/library/general/allowedrepos
panic: looks like template.yaml is updated but the version is not. Please update the 'metadata.gatekeeper.sh/version' annotation in the template.yaml source
```
I guess that answered my question about whether versioning was needed 😄
Am I correct to think a patch version bump would be needed for these changes?

.github/workflows/workflow.yaml
```diff
@@ -45,7 +45,7 @@ jobs:
     strategy:
       matrix:
         os: [ "ubuntu-latest", "macos-latest" ]
-        opa: [ "v0.42.2" ]
+        opa: [ "v0.54.0" ]
```
apeabody, line 48: Thanks for the contribution @anderseknert - Is opa `v0.54.0` required for this change?  We should just be aware of allowing future Rego changes that might not be compatible with previous Gatekeeper versions.  Currently we test with n-1, for which Gatekeeper v3.12 uses opa `v0.49.2`.

apeabody:
> ```
> Generating artifact hub content for  /home/runner/work/gatekeeper-library/gatekeeper-library/library/general/allowedrepos
> panic: looks like template.yaml is updated but the version is not. Please update the 'metadata.gatekeeper.sh/version' annotation in the template.yaml source
> ```
> 
> I guess that answered my question about whether versioning was needed 😄 Am I correct to think a patch version bump would be needed for these changes?

Hi @anderseknert - Yup, assuming these are all backward compatible (no change in behavior), they can just be a patch version bumps.

.github/workflows/workflow.yaml
anderseknert: Sure thing! I've just run the tests with v0.44.0 without issues, so I can set the version to that.

Will proceed with patching the versions. Thanks!

anderseknert:
@apeabody all versions now bumped accordingly, and the tests seemingly pleased.

artifacthub/library/general/allowedrepos/1.0.0/artifacthub-pkg.yml
```diff
@@ -3,7 +3,7 @@ name: k8sallowedrepos
 displayName: Allowed Repositories
 createdAt: "2022-09-26T17:28:27Z"
 description: Requires container images to begin with a string from the specified list.
-digest: eaec68f7273f5d79ced3ed778d5c2fc26b241e5db3a0e5119650bab651f8f1ed
```
apeabody, line 6: Hi @anderseknert - I suspect these previous version changes are likely "left over" from the generate prior to the template's patch version bump.  An option might be to drop the changes from `artifacthub/library` and then re-run `make generate-all`.

anderseknert: Thanks, I'll give it a shot!

.github/workflows/workflow.yaml
apeabody: Thanks - Those folders look good now!

artifacthub/library/general/allowedrepos/1.0.0/artifacthub-pkg.yml
apeabody: Thanks - Those folders look good now!

apeabody (approved):
Thanks for the contribution @anderseknert!

A second maintainer will also need to review.

maxsmythe:
@anderseknert

Quick question about not being able to override `input`. Gatekeeper's Rego driver relies on this behavior to get constraint arguments in a performant manner. We could switch to specifying the input directly for each individual constraint, but in (admittedly old) performance testing, that was found to be slower.

The question is... will the below linked Rego code eventually break?

https://github.com/open-policy-agent/frameworks/blob/3eb381ce6cbedf3c1adedf2bfb20aa5e491c5baa/constraint/pkg/client/drivers/rego/rego.go#L27-L37

anderseknert:
@maxsmythe no, that's fine! You can still override `input` using `with`. You just can't shadow the global `input` variable itself, i.e. not `input := { ... }` but `inp := { ... }` (or any other name). Saying `with input as inp` or `with input as { ... }` is just doing what `with` is there for 🙂

maxsmythe:
@anderseknert Awesome, thanks for clarifying!

.github/workflows/workflow.yaml
```diff
@@ -45,7 +45,7 @@ jobs:
     strategy:
       matrix:
         os: [ "ubuntu-latest", "macos-latest" ]
-        opa: [ "v0.42.2" ]
+        opa: [ "v0.44.0" ]
```
sozercan, line 48: do we want to bump this to 0.57.1 (same as GK 3.14)?

apeabody: @sozercan - At this moment we are still testing against GK 3.12 (n-1), so I would recommend we not go higher than `v0.49.2` at this exact moment.  That said, we can bump the CI to GK 3.13 once 3.14 is released.  `v0.44.0` has the advantage of supporting back to GK v3.10.

More context: https://github.com/open-policy-agent/gatekeeper-library/pull/429#discussion_r1376749615

apeabody: I see the GK 3.14 release just went up, so I'll open a PR to update the CI in a moment.

apeabody: The CI update to GK 3.14 (and retaining 3.13) is now merged!

@sozercan - As this is a matrix value, what if we **add** `"0.57.1"`?  That way we  can test against both the GK 3.14 opa version, but still maintain the signal for older versions.

sozercan: sgtm

anderseknert: Do you want me to do something here?

apeabody, line 48:
```suggestion
        opa: [ "v0.44.0", "0.57.1" ]
```

apeabody: Hi @anderseknert - Sorry for the ambiguity, I just suggested the change based on the outcome.

anderseknert: No worries! I've just pushed that change.

apeabody: Thanks! Running checks with both versions looks good.

apeabody:
@sozercan - We are now testing with both OPA versions, when convenient PTAL

apeabody:
Hi @sozercan or @ritazh Please take a look when convenient - Thanks!

nilekhc (approved):

nilekhc (commented):
@anderseknert

Looks like unit tests are failing. Could you PTAL?

apeabody:
> @anderseknert
> 
> Looks like unit tests are failing. Could you PTAL?

Ahh, looks like it is the `k8sdisallowinteractivetty` template added in https://github.com/open-policy-agent/gatekeeper-library/commit/2d99845f3188821ecd261ca63e81b4c9d0fe9dea

```
src/general/disallowinteractive/src_test.rego:4: rego_compile_error: variables must not shadow input (use a different variable name)
src/general/disallowinteractive/src_test.rego:9: rego_compile_error: variables must not shadow input (use a different variable name)
src/general/disallowinteractive/src_test.rego:14: rego_compile_error: variables must not shadow input (use a different variable name)
src/general/disallowinteractive/src_test.rego:19: rego_compile_error: variables must not shadow input (use a different variable name)
src/general/disallowinteractive/src_test.rego:24: rego_compile_error: variables must not shadow input (use a different variable name)
src/general/disallowinteractive/src_test.rego:29: rego_compile_error: variables must not shadow input (use a different variable name)
src/general/disallowinteractive/src_test.rego:34: rego_compile_error: variables must not shadow input (use a different variable name)
src/general/disallowinteractive/src_test.rego:39: rego_compile_error: variables must not shadow input (use a different variable name)
src/general/disallowinteractive/src_test.rego:44: rego_compile_error: variables must not shadow input (use a different variable name)
```

anderseknert:
Yes, some new tests overloading `input`, as is now verboten. I just fixed that and pushed.

nilekhc (approved):

PR #437, opened 2023-11-06
fix: clarify template behavior in description

apeabody:
**What this PR does / why we need it**: Clarify template behavior in description and move comment to issue https://github.com/open-policy-agent/gatekeeper-library/pull/437.  If the comment was implemented the new capability would need to be gated with a new parameter to preserve backward compatible behavior anyway.


maxsmythe (approved):
LGTM

nilekhc (approved):

PR #461, opened 2024-01-11
host-network-ports: Allow hostNetwork when a container is using listed image in exemptImages

### LINKED ISSUE #460: k8spsphostnetworkingports exemptImages does not allow hostNetwork
The k8spsphostnetworkingports policy do not allow a pod that contains `exemptImage` to use `hostNetwork`.

I believe it should either

1. Documented. A similar policy is k8spsphostfilesystem which validate pod-level volumes do not support this feature and do not have `exemptImages` parameter.
2. If one of the containers is exempt, then the whole pod can set hostNetwork to any value.

whs:

Fixes #460


maxsmythe:
please see my comment on #460: https://github.com/open-policy-agent/gatekeeper-library/issues/460#issuecomment-1889961074

There are security implications for this change.

maxsmythe (changes requested):
Requesting changes to make concerns more clear.

PR #502, opened 2024-04-02
Update capabilities.md for more clarity

cccsss01:

Received feedback from a developer it was unclear what the message meant, adding the noun at the end of the 2nd sentence provides more clarity instead of carrying the context from the 1st sentence. 

Fixes #
no issue

Do I really need to do this?  

### DIFF
--- a/website/docs/validation/capabilities.md
+++ b/website/docs/validation/capabilities.md
@@ -80,7 +80,7 @@ spec:
           container := input.review.object.spec.containers[_]
           not is_exempt(container)
           missing_drop_capabilities(container)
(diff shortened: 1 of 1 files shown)


apeabody (changes requested):
Thanks for the contribution @cccsss01!

The documentation is automatically built from the actual template, so you would want to upgrade it directly (https://github.com/open-policy-agent/gatekeeper-library/blob/master/src/pod-security-policy/capabilities/src.rego) and then follow the build process (https://github.com/open-policy-agent/gatekeeper-library/tree/master?tab=readme-ov-file#development).

PR #514, opened 2024-04-24
feat(general): Add volumeresources emptyDir sizelimit

dongjiang1989:
feat(general): Add volumeresources emptyDir sizelimit.

one node in the cluster was emptyDirevicted because the log volume not set a capacity limit `sizeLimit`.

Fixes #


JaydipGabani:
@dongjiang1989 can you also modify empty `suite.yaml` with appropriate configurations? - here is an example of working `suite.yaml` - https://github.com/open-policy-agent/gatekeeper-library/blob/master/library/general/block-wildcard-ingress/suite.yaml

dongjiang1989:
> @dongjiang1989 can you also modify empty `suite.yaml` with appropriate configurations? - here is an example of working `suite.yaml` - https://github.com/open-policy-agent/gatekeeper-library/blob/master/library/general/block-wildcard-ingress/suite.yaml

Thanks. @JaydipGabani 
Fixed. Please re-check

dongjiang1989:
How to debug locally?

https://github.com/open-policy-agent/gatekeeper-library/actions/runs/8827755032/job/24272934381?pr=514

library/general/volumeresources/samples/container-emptydir-limit/constraint.yaml
```diff
@@ -0,0 +1,13 @@
+apiVersion: constraints.gatekeeper.sh/v1beta1
+kind: K8sVolumeRequests
+metadata:
+  name: container-emptydir-limit
+spec:
+  match:
+    kinds:
+      - apiGroups: [""]
+        kinds: ["Pod"]
```
JaydipGabani, line 9: This policy will not work for pods since in rego, it is looking for `input.review.object.spec.template.spec` which is not available in `Pod`. I suggest writing this policy for `Pod` and then to impose same policy for workload resources `ExpansionTemplate` of gatekeeper could be used.

JaydipGabani (changes requested):
This policy will not work for Pod. Please modify constraint or rego.

library/general/volumeresources/samples/container-emptydir-limit/constraint.yaml
dongjiang1989: Thanks @JaydipGabani 
Add `ExpansionTemplate` done.

JaydipGabani:
@dongjiang1989 appologies for going back and forth, but it would be best to keep the policy applicable to pods because currenlty `library` doesn't test policies with `expansionTemplates` and the CI will likely fail since you are using `Deployment` for allowed/disallowed examples, you would not get expected violations in return. The policy would still deny the pods spinned up by workload resouces as well, but the denied message wouldn't be in the stdout without `expansionTemplate` (the denied message could be found on `status` of the parent resource for the pod).

dongjiang1989:
> @dongjiang1989 appologies for going back and forth, but it would be best to keep the policy applicable to pods because currenlty `library` doesn't test policies with `expansionTemplates` and the CI will likely fail since you are using `Deployment` for allowed/disallowed examples, you would not get expected violations in return. The policy would still deny the pods spinned up by workload resouces as well, but the denied message wouldn't be in the stdout without `expansionTemplate` (the denied message could be found on `status` of the parent resource for the pod).

Thanks @JaydipGabani . PTAL re-check

Keep the policy applicable to pods done.

JaydipGabani:
@dongjiang1989 you will need to remove examples with `kind: Deployment` from `artifacthub/` dir.

dongjiang1989:
> @dongjiang1989 you will need to remove examples with `kind: Deployment` from `artifacthub/` dir.

Thanks for your review.  Done.

JaydipGabani (approved):
LGTM

JaydipGabani:
@maxsmythe @ritazh @sozercan PTAL.

artifacthub/library/general/volumeresources/1.0.0/template.yaml
```diff
@@ -0,0 +1,164 @@
...
+        	suffix := get_suffix(orig)
+        	raw := replace(orig, suffix, "")
+        	regex.match("^[0-9]+(\\.[0-9]+)?$", raw)
+        	new := to_number(raw) * size_multiple(suffix)
+        }
+      libs:
+        - |
+          package lib.exempt_container
+
+          is_exempt(container) {
```
ritazh, line 148: where is  is_exempt used? might be good to also add it to the allowed test case

artifacthub/library/general/volumeresources/1.0.0/template.yaml
```diff
@@ -0,0 +1,164 @@
+apiVersion: templates.gatekeeper.sh/v1
+kind: ConstraintTemplate
+metadata:
+  name: k8svolumerequests
```
ritazh, line 4: since this policy validates specifically for emptydir volume size limit, the name, description should be updated to reflect this.

ritazh:
Thanks for the PR!

artifacthub/library/general/volumeresources/1.0.0/template.yaml
dongjiang1989: @ritazh  change name to `k8semptydirvolumesizelimit`?

dongjiang1989:
@ritazh @JaydipGabani Please re-check it, thanks

dongjiang1989:
@JaydipGabani PTAL, thanks

artifacthub/library/general/volumeresources/1.0.0/template.yaml
JaydipGabani: Please add something similar to test exempted containers - https://github.com/open-policy-agent/gatekeeper-library/blob/master/library/pod-security-policy/seccomp/samples/psp-seccomp/example_allowed_exempt_image.yaml

JaydipGabani (changes requested):

PR #503, opened 2024-04-04
adding cel for allow-privilege-escalation

maxsmythe:
**What this PR does / why we need it**:

This converts K8sAllowPrivilegeEscalation to use CEL in addition to Rego.

**Which issue(s) does this PR fix** *(optional, using `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when the PR gets merged)*:
Fixes #541 

**Special notes for your reviewer**:

PR #519, opened 2024-05-02
ci: testing with cel policies

JaydipGabani:
- Modifies the testing workflow to run `gator verify` with cel and rego engines
- Tests cel and rego policies with gatekeeper
- Adds cel source for `requiredLables` and `allow-priviledge-escalation`

Fixes #


JaydipGabani:
@maxsmythe I added CEL source from your draft PR #503  here as well, that leaves you with 3 other policies to write CEL source for.

.github/workflows/workflow.yaml
```diff
@@ -65,7 +65,8 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        gatekeeper: [ "release-3.13", "release-3.14" ]
+        gatekeeper: [ "3.14.2", "3.15.1" ]
```
ritazh, line 68: why do we prefer to use specific patch versions instead of the latest?

Makefile
```diff
@@ -31,21 +32,34 @@ integration-bootstrap:
 	TERM=dumb ${GITHUB_WORKSPACE}/bin/kind create cluster --image kindest/node:v${KUBERNETES_VERSION} --wait 5m --config=test/kind_config.yaml
 
 deploy:
-	kubectl apply -f https://raw.githubusercontent.com/open-policy-agent/gatekeeper/${GATEKEEPER_VERSION}/deploy/gatekeeper.yaml
+	helm repo add gatekeeper https://open-policy-agent.github.io/gatekeeper/charts
+ifeq ($(POLICY_ENGINE), rego)
+	helm install gatekeeper/gatekeeper --name-template=gatekeeper --namespace gatekeeper-system --create-namespace --version=${GATEKEEPER_VERSION}
```
ritazh, line 37:
```suggestion
	helm install gatekeeper/gatekeeper --name-template=gatekeeper --namespace gatekeeper-system --create-namespace --version=${GATEKEEPER_VERSION} --set enableK8sNativeValidation=false
```

Explicitly setting the flag here so we dont need to update this when GK enables it by default in the future.

Makefile
```diff
@@ -31,21 +32,34 @@ integration-bootstrap:
...
-	kubectl delete -f https://raw.githubusercontent.com/open-policy-agent/gatekeeper/${GATEKEEPER_VERSION}/deploy/gatekeeper.yaml
+	helm delete gatekeeper --namespace gatekeeper-system
 
 test-integration:
 	bats -t test/bats/test.bats
 
 .PHONY: verify-gator
 verify-gator:
+ifeq ($(POLICY_ENGINE), rego)
 	gator verify ./...
```
ritazh, line 51: --set enableK8sNativeValidation=false

artifacthub/library/general/requiredlabels/1.0.2/template.yaml
```diff
@@ -0,0 +1,77 @@
...
+    - target: admission.k8s.gatekeeper.sh
+      code:
+      - engine: K8sNativeValidation
+        source:
+
+                      validations:
+                      - expression: '[object, oldObject].exists(obj, obj != null && has(obj.metadata) && variables.params.labels.all(entry, has(obj.metadata.labels) && entry.key in obj.metadata.labels))'
+                        messageExpression: '"missing required label, requires all of: " + variables.params.labels.map(entry, entry.key).join(", ")'
+                      - expression: '[object, oldObject].exists(obj, obj != null && !variables.params.labels.exists(entry, has(obj.metadata.labels) && entry.key in obj.metadata.labels && !string(obj.metadata.labels[entry.key]).matches(string(entry.allowedRegex))))'
+                        message: "regex mismatch"
```
ritazh, line 49: spacing seems off here?

library/general/requiredlabels/template.yaml
```diff
@@ -38,6 +38,15 @@ spec:
...
     - target: admission.k8s.gatekeeper.sh
+      code:
+      - engine: K8sNativeValidation
+        source:
+
+                      validations:
+                      - expression: '[object, oldObject].exists(obj, obj != null && has(obj.metadata) && variables.params.labels.all(entry, has(obj.metadata.labels) && entry.key in obj.metadata.labels))'
+                        messageExpression: '"missing required label, requires all of: " + variables.params.labels.map(entry, entry.key).join(", ")'
+                      - expression: '[object, oldObject].exists(obj, obj != null && !variables.params.labels.exists(entry, has(obj.metadata.labels) && entry.key in obj.metadata.labels && !string(obj.metadata.labels[entry.key]).matches(string(entry.allowedRegex))))'
+                        message: "regex mismatch"
```
ritazh, line 49: spacing

src/general/requiredlabels/constraint.tmpl
```diff
@@ -38,5 +38,9 @@ spec:
                       the regular expression.
   targets:
     - target: admission.k8s.gatekeeper.sh
+      code:
+      - engine: K8sNativeValidation
+        source:
+{{ file.Read "src/general/requiredlabels/src.cel" | strings.Indent 10 | strings.TrimSuffix "\n" }}
```
ritazh, line 44: should add a validation for src.cel: https://github.com/open-policy-agent/gatekeeper-library/blob/master/scripts/generate.sh#L39

.github/workflows/workflow.yaml
JaydipGabani: AFAIK, we have always tested with past two release versions. I am fine with testing it with latest, unless others have any objections.

src/general/requiredlabels/constraint.tmpl
JaydipGabani: will it not error out for policies that sill do not have `src.cel` file? May be we probably should add `src.cel` to the list once we have cel source for all policies. Even in that case the referential policies won't have `src.cel`. wdyt?

.github/workflows/workflow.yaml
ritazh: using the branch instead of the tag ensures we are always testing against the latest patch version without having to push a PR to update it.

src/general/requiredlabels/constraint.tmpl
ritazh: https://github.com/open-policy-agent/gatekeeper-library/blob/883683a0a49f82dea424f4a46cbcf9c94659c0c9/scripts/generate.sh#L39

if there's no src.cel file, it won't go into the loop.

src/general/requiredlabels/src.cel
```diff
@@ -0,0 +1,6 @@
+
+validations:
+- expression: '[object, oldObject].exists(obj, obj != null && has(obj.metadata) && variables.params.labels.all(entry, has(obj.metadata.labels) && entry.key in obj.metadata.labels))'
+  messageExpression: '"missing required label, requires all of: " + variables.params.labels.map(entry, entry.key).join(", ")'
+- expression: '[object, oldObject].exists(obj, obj != null && !variables.params.labels.exists(entry, has(obj.metadata.labels) && entry.key in obj.metadata.labels && !string(obj.metadata.labels[entry.key]).matches(string(entry.allowedRegex))))'
+  message: "regex mismatch"
```
ritazh, line 6: line break

src/general/requiredlabels/src.cel
```diff
@@ -0,0 +1,6 @@
+
```
ritazh, line 1: remove this line

src/general/requiredlabels/constraint.tmpl
JaydipGabani: So by validation of the `src.cel` file do you mean syntax validation? will it run or not kind of a thing?

.github/workflows/workflow.yaml
JaydipGabani: AFAIK helm can only install specifi version, I am not sure if it can take a branch and install the latest patch on that branch. If we do not use helm, then we would need to use `sed` or something similar to modify the file and set the flag manually.

sozercan: +1 to @JaydipGabani i think this is a better approach now since we previously installed from yaml directly

artifacthub/library/general/requiredlabels/1.0.2/template.yaml
```diff
@@ -0,0 +1,76 @@
...
+    - target: admission.k8s.gatekeeper.sh
+      code:
+      - engine: K8sNativeValidation
+        source:
+          validations:
+          - expression: '[object, oldObject].exists(obj, obj != null && has(obj.metadata) && variables.params.labels.all(entry, has(obj.metadata.labels) && entry.key in obj.metadata.labels))'
+            messageExpression: '"missing required label, requires all of: " + variables.params.labels.map(entry, entry.key).join(", ")'
+          - expression: '[object, oldObject].exists(obj, obj != null && !variables.params.labels.exists(entry, has(obj.metadata.labels) && entry.key in obj.metadata.labels && !string(obj.metadata.labels[entry.key]).matches(string(entry.allowedRegex))))'
+            message: "regex mismatch"
+      rego: |
```
sozercan, line 49: can we make this consistent with using engine for both?

src/general/requiredlabels/constraint.tmpl
```diff
@@ -38,5 +38,9 @@ spec:
...
       rego: |
```
sozercan, line 45: you'll need to update here for my prev comment about using engine here

.github/workflows/workflow.yaml
ritazh: ah I see. I'm ok with merging this as is. but it would be better to not having to bump the tag all the time. @JaydipGabani you can create an issue to do it as a follow up.

src/general/requiredlabels/constraint.tmpl
ritazh: the linked code  https://github.com/open-policy-agent/gatekeeper-library/blob/master/scripts/generate.sh#L39 checks if the file exists or not.

artifacthub/library/general/requiredlabels/1.0.2/template.yaml
ritazh, line 49:
```suggestion
       - engine: Rego
        source: 
          rego: |
```
Move `rego` and `libs` under source

ritazh: This is the code https://github.com/open-policy-agent/frameworks/blob/c2efb00269a87e94f011ec8bec28d594280cb41e/constraint/pkg/client/drivers/rego/schema/schema.go#L19-L23

.github/workflows/workflow.yaml
JaydipGabani: Sure!

src/general/requiredlabels/constraint.tmpl
JaydipGabani: > I was suggesting to create a new loop. if there's no src.cel file, it won't go into the loop.

@ritazh I am sorry for the confusion (may be I am getting something wrong here). Seems like what you are suggesting is a new for loop that verifies the existance of `src.cel` file. What happens when the file do no exists? 

IMO I do not think we need this validation loop at all (`for req in src.rego src_test.rego constraint.tmpl; do`) because if those files are missing we will error out during generation of the template - [generate.sh#l16](https://github.com/open-policy-agent/gatekeeper-library/blob/master/scripts/generate.sh#L16).

I think we should remove the validation code.

ritazh: >  if those files are missing we will error out during generation of the template

sgtm

Let's leave the existing loop since it also checks for src_test.rego otherwise it will only break when running the tests.

ritazh (approved):
LGTM

.github/workflows/workflow.yaml
sozercan: I don't think we need to bump the patch version. We'll need to bump the minor version, which is same as today

src/general/requiredlabels/src.cel
```diff
@@ -0,0 +1,5 @@
+validations:
+- expression: '[object, oldObject].exists(obj, obj != null && has(obj.metadata) && variables.params.labels.all(entry, has(obj.metadata.labels) && entry.key in obj.metadata.labels))'
```
maxsmythe, line 2: I think we want `.all()` instead of `.exists()`, otherwise oldObject being compliant will short-circuit throwing a violation for object being non-compliant.

Also, we should not check oldObject at all. That's not how the Rego is written.

src/general/requiredlabels/src.cel
```diff
@@ -0,0 +1,5 @@
...
+  messageExpression: '"missing required label, requires all of: " + variables.params.labels.map(entry, entry.key).join(", ")'
+- expression: '[object, oldObject].exists(obj, obj != null && !variables.params.labels.exists(entry, has(obj.metadata.labels) && entry.key in obj.metadata.labels && !string(obj.metadata.labels[entry.key]).matches(string(entry.allowedRegex))))'
```
maxsmythe, line 4: object/oldObject concerns here.

src/pod-security-policy/allow-privilege-escalation/src.cel
```diff
@@ -0,0 +1,26 @@
...
+  expression: |
+    (variables.containers + variables.initContainers + variables.ephemeralContainers).filter(container,
+    !(container.image in variables.exemptImages) && (
+        !has(container.securityContext) ||
+        !has(container.securityContext.allowPrivilegeEscalation) ||
+        container.securityContext.allowPrivilegeEscalation != false
+    )
+    )
+validations:
+- expression: '(has(request.operation) && request.operation == "UPDATE") || size(variables.badContainers) == 0'
```
maxsmythe, line 25: The `UPDATE` guard is off... if request.operation is undefined, then the policy should be evaluated. Otherwise audit wont work.

Rego logic path:

* if not is_update(), continue processing violation
* is_update() is falsely if: `review.operation` is undefined or `review.operation == "UPDATE"`

maxsmythe (dismissed):
Couple comments on the CEL code. Changes to testing flow LGTM!

src/general/requiredlabels/src.cel
ritazh, line 2:
```suggestion
- expression: '(has(request.operation) && request.operation == "DELETE") ||(has(object.metadata) && variables.params.labels.all(entry, has(object.metadata.labels) && entry.key in object.metadata.labels))'
              messageExpression: '"object missing required label, requires all of: " + variables.params.labels.map(entry, entry.key).join(", ")'
```
This works for audit since the `||` handles `operation == ""`

ritazh: I tested this in my cluster, `.all` would expect both to exist, which would block good resources on create.

src/pod-security-policy/allow-privilege-escalation/src.cel
ritazh: This works for audit since the || handles operation == ""

maxsmythe: Oops, you are correct.

library/general/requiredlabels/template.yaml
```diff
@@ -38,31 +38,41 @@ spec:
                       the regular expression.
   targets:
     - target: admission.k8s.gatekeeper.sh
-      rego: |
-        package k8srequiredlabels
+      code:
+      - engine: K8sNativeValidation
+        source:
+          validations:
+          - expression: '(has(request.operation) && request.operation == "DELETE") ||(has(object.metadata) && variables.params.labels.all(entry, has(object.metadata.labels) && entry.key in object.metadata.labels))'
```
maxsmythe, line 45: We do not need to handle DELETE as a special case. CEL-VAP sees the same input objects that Rego does. We should do a straight re-implementation of the Rego logic (which does not special-case delete).

In general, mirroring the Rego code's logic exactly is the correct move, otherwise there will be edge cases where the two implementations behave differently.

ritazh: hmm, recall when we first integrated vap, I had to explicitly check `oldObject` instead of `object` because on DELETE, the expression `object.metadata` was failing so it was always blocking on DELETE.

library/general/requiredlabels/template.yaml
```diff
@@ -38,31 +38,41 @@ spec:
...
+            messageExpression: '"object missing required label, requires all of: " + variables.params.labels.map(entry, entry.key).join(", ")'
+          - expression: '[object, oldObject].exists(obj, obj != null && !variables.params.labels.exists(entry, has(obj.metadata.labels) && entry.key in obj.metadata.labels && !string(obj.metadata.labels[entry.key]).matches(string(entry.allowedRegex))))'
```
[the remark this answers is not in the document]
ritazh: This whole block needs to be updated once we figure out the right expression for the one above it to handle DELETE correctly. We can ignore it for now.

Makefile
```diff
@@ -31,21 +32,34 @@ integration-bootstrap:
 	TERM=dumb ${GITHUB_WORKSPACE}/bin/kind create cluster --image kindest/node:v${KUBERNETES_VERSION} --wait 5m --config=test/kind_config.yaml
 
 deploy:
-	kubectl apply -f https://raw.githubusercontent.com/open-policy-agent/gatekeeper/${GATEKEEPER_VERSION}/deploy/gatekeeper.yaml
+	helm repo add gatekeeper https://open-policy-agent.github.io/gatekeeper/charts
+ifeq ($(POLICY_ENGINE), rego)
+	helm install gatekeeper/gatekeeper --name-template=gatekeeper --namespace gatekeeper-system --create-namespace --version=${GATEKEEPER_VERSION} --set enableK8sNativeValidation=false
```
JaydipGabani, line 37: `enableK8sNativeValidation` variable in helm chart is not available for `3.15.1`, we need `sed` or something else to enable VAP for testing in 3.15.x

library/general/requiredlabels/template.yaml
ritazh: I forgot we changed this part of  the generated vap resource so DELETE is ignored.
```yaml
operations:
   - CREATE
   - UPDATE
   - ```

Comment (ritazh) [inline library/general/requiredlabels/template.yaml:45]:
```suggestion
          - expression: '(has(object.metadata) && variables.params.labels.all(entry, has(object.metadata.labels) && entry.key in object.metadata.labels))'
```
I tested this.

src/general/requiredlabels/constraint.tmpl
```diff
@@ -38,5 +38,8 @@ spec:
                       the regular expression.
   targets:
     - target: admission.k8s.gatekeeper.sh
-      rego: |
-{{ file.Read "src/general/requiredlabels/src.rego" | strings.Indent 8 | strings.TrimSuffix "\n" }}
+      code:
+      - engine: Rego
```
ritazh, line 42: During the community call, we said let's keep the cel code for this policy and remove the changes for k8spspallowprivilegeescalationcontainer since that one has issues with the exemptImages, this policy doesn't.

Without the cel code, we cant validate this PR actually works.

you can use the code https://github.com/open-policy-agent/gatekeeper-library/pull/519#discussion_r1594342487 and update the second expression using the same logic.

src/general/requiredlabels/src.cel
maxsmythe: Suggestion of removing oldObject SGTM

src/general/requiredlabels/constraint.tmpl
JaydipGabani: Appologies for the misunderstanding. I thought the decision was to riase a different PR for policy. Added the policy here.

src/general/requiredlabels/src.cel
ritazh: still needs the line break?

src/general/requiredlabels/src.cel
```diff
@@ -0,0 +1,5 @@
+validations:
+- expression: '(has(object.metadata) && variables.params.labels.all(entry, has(object.metadata.labels) && entry.key in object.metadata.labels))'
+  messageExpression: '"missing required label, requires all of: " + variables.params.labels.map(entry, entry.key).join(", ")'
+- expression: '(has(object.metadata) && !variables.params.labels.all(entry, has(object.metadata.labels) && entry.key in object.metadata.labels && !string(object.metadata.labels[entry.key]).matches(string(entry.allowedRegex))))'
```
ritazh, line 4: Can you please test the regex use case? Maybe even add a test for it? 

> !string(object.metadata.labels[entry.key]).matches(string(entry.allowedRegex)

This feels incorrect...it should match on the allowedRegex not `!`. Please test and verify this. The example in GK might also be wrong.

artifacthub/library/pod-security-policy/allow-privilege-escalation/1.0.1/artifacthub-pkg.yml
```diff
@@ -1,7 +1,7 @@
 version: 1.0.1
 name: k8spspallowprivilegeescalationcontainer
 displayName: Allow Privilege Escalation in Container
-createdAt: "2023-05-23T09:47:31Z"
+createdAt: "2024-05-09T19:10:40Z"
```
ritazh, line 4: Please remove

src/general/requiredlabels/src.cel
JaydipGabani: @ritazh I think we need it to `not match` on the allowedRegex. If value do not match the `allowedRegex` then its a violation otherwise it is not.

JaydipGabani: I added the example testing the same.

ritazh: I missed `!variable` :)

src/general/requiredlabels/constraint.tmpl
```diff
@@ -4,7 +4,7 @@ metadata:
   name: k8srequiredlabels
   annotations:
     metadata.gatekeeper.sh/title: "Required Labels"
-    metadata.gatekeeper.sh/version: 1.0.1
+    metadata.gatekeeper.sh/version: 1.0.2
```
ritazh, line 7:
```suggestion
    metadata.gatekeeper.sh/version: 1.1.0
```
do we want to at least bump the minor version? @maxsmythe @sozercan

Makefile
```diff
@@ -31,7 +32,19 @@ integration-bootstrap:
 	TERM=dumb ${GITHUB_WORKSPACE}/bin/kind create cluster --image kindest/node:v${KUBERNETES_VERSION} --wait 5m --config=test/kind_config.yaml
 
 deploy:
-	kubectl apply -f https://raw.githubusercontent.com/open-policy-agent/gatekeeper/${GATEKEEPER_VERSION}/deploy/gatekeeper.yaml
+	wget https://raw.githubusercontent.com/open-policy-agent/gatekeeper/${GATEKEEPER_VERSION}/deploy/gatekeeper.yaml
+ifeq ($(POLICY_ENGINE), rego)
+	sed -i '/- args:/a \ \ \ \ \ \ \ \ - --validate-template-rego=true' gatekeeper.yaml
```
JaydipGabani, line 37: we should probably remove this line as well. It will cause errors for 3.16+ as the flag is removed.

.github/workflows/workflow.yaml
```diff
@@ -65,7 +65,8 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        gatekeeper: [ "release-3.13", "release-3.14" ]
+        gatekeeper: [ "release-3.14", "release-3.15" ]
+        engine: [ "cel", "rego" ]
```
ritazh, line 69: for cel, we need to add release-3.16 here and for gator since only 3.16 has the fix for the param bug. can you pls update this such that for versions before 3.16, skip CEL engine.

Makefile
JaydipGabani: removed the code

src/general/requiredlabels/src.cel
JaydipGabani: @ritazh I changed the CEL source. Let me know what you think.

maxsmythe:
Can you drop the CEL from my draft PR? I'd rather only have one SOT for that, and just rebase/merge once the build system is ready, otherwise feedback will get clobbered.

library/general/requiredlabels/template.yaml
```diff
@@ -38,31 +38,42 @@ spec:
                       the regular expression.
   targets:
     - target: admission.k8s.gatekeeper.sh
-      rego: |
-        package k8srequiredlabels
+      code:
+      - engine: K8sNativeValidation
+        source:
+          validations:
+          - expression: '(has(object.metadata) && variables.params.labels.exists_one(entry, has(object.metadata.labels) && entry.key in object.metadata.labels))'
```
maxsmythe, line 45: `exists_one()` is wrong. This check requires exactly ONE required label to be present (not zero, not more than one). `all()` is the correct check, since all required labels must be present.

library/general/requiredlabels/template.yaml
```diff
@@ -38,31 +38,42 @@ spec:
...
+            messageExpression: '"missing required label, requires all of: " + variables.params.labels.map(entry, entry.key).join(", ")'
+          - expression: '(has(object.metadata) && variables.params.labels.exists_one(entry, has(object.metadata.labels) && entry.key in object.metadata.labels && string(object.metadata.labels[entry.key]).matches(string(entry.allowedRegex))))'
```
maxsmythe, line 47: same issue here (use all(), not exists_one())

ritazh:
> Can you drop the CEL from my draft PR? I'd rather only have one SOT for that, and just rebase/merge once the build system is ready, otherwise feedback will get clobbered.

Per this https://github.com/open-policy-agent/gatekeeper-library/pull/519#discussion_r1594796796 all updates for "Allow Privilege Escalation" policy have been removed

src/general/requiredlabels/constraint.tmpl
maxsmythe: Makes sense.

maxsmythe (changes requested):

library/general/requiredlabels/template.yaml
JaydipGabani: Oh Yes! missed on that. Good catch.

JaydipGabani: I think I finally got the CEL code right. Let me know how it looks now.

ritazh:
@JaydipGabani ptal conflicts

.github/workflows/workflow.yaml
```diff
@@ -65,29 +65,35 @@ jobs:
...
     strategy:
       matrix:
-        gatekeeper: [ "release-3.13", "release-3.14" ]
-    name: "Integration test on Gatekeeper ${{ matrix.gatekeeper }}"
+        gatekeeper: [ "3.15.1", "3.16.0" ]
+        engine: [ "cel", "rego" ]
+    name: "Integration test on Gatekeeper ${{ matrix.gatekeeper }} for ${{ matrix.engine }} policies"
     steps:
       - name: Harden Runner
+        if: ${{ (matrix.gatekeeper == '3.15.1' && matrix.engine == 'rego') || matrix.gatekeeper == '3.16.0' }} # remove this condition once 3.17 is out
```
sozercan, line 73: why do we need this for harden runner?

sozercan: nm this is because of the matrix

sozercan: can we do this opposite? if gk 3.15 and cel, we exit 0

sozercan (approved):
1 minor comment, otherwise lgtm

.github/workflows/workflow.yaml
JaydipGabani: @sozercan seems like there is no way to do exit for the whole job from a step - https://github.com/orgs/community/discussions/82744. We would need if condition for all steps to skip it regardless.

PR #521, opened 2024-05-10
Forbidden sysctls cel

ritazh:

Fixes #541 

will rebase once #519 is merged


website/docs/validation/forbidden-sysctls.md
```diff
@@ -51,65 +51,86 @@ spec:
...
+      - engine: K8sNativeValidation
+        source:
+          variables:
+          - name: sysctls
+            expression: '!has(object.spec.securityContext) ? [] : !has(object.spec.securityContext.sysctls) ? [] : object.spec.securityContext.sysctls'
+          - name: forbiddenSysctls
+            expression: |
+              !has(variables.params.forbiddenSysctls) ? [] :
+              variables.sysctls.filter(sysctl,
+                  variables.params.forbiddenSysctls.exists(forbiddenSysctl, sysctl.name == forbiddenSysctl || (forbiddenSysctl.endsWith("*") && string(sysctl.name).matches("^" + string(forbiddenSysctl).replace("*", ".*") + "$"))))
```
maxsmythe, line 64: I don't think we can use regex here, sadly, since sysctls use `.` (which is a regex command char) in their allowed character set:

https://kubernetes.io/docs/tasks/administer-cluster/sysctl-cluster/#safe-and-unsafe-sysctls

Since we only allow prefix matching, I suggest the following (context links at bottom):

- create variables called `allowedSysctlPrefixes` and `disallowedSysctlPrefixes` (or similar) and a companion variable called `allowedSysctlExplicit` and `disallowedSysctlExplicit` (or similar).
- Iterate through the constraint parameters, finding entries that end with `*`, adding them to the glob list for their respective parameters (removing the trailing `*`)
- We'll also need to special-case the check for `*` if it exists, maybe set a boolean variable called `allAllowed` and/or `allDisallowed`? Missing parameter should yield `allAllowed` == true and `allDisallowed` == false for allowedSysctls/forbiddenSysctls respectively
- create still another variable `violatingSysctls` or similar iterate through `spec.securityContext.sysctls`, running the following checks (with this precedence):
  - If `allDisallowed`, add the sysctl name to `violatingSysctls`
  - If `disallowedSysctlPrefixes` satisfies a `sysctl.startsWith(prefix)` check, add the sysctl name to violatingSysctls
  - If `disallowedSysctlExplicit` contains the sysctl (`in` keyword), add the sysctl name to `violatingSysctls`
  - if allowed check fails, add the sysctl name to `violatingSysctls` allowed check performs (in order of precedence)
    - If `allAllowed`, check succeeds
    - If `allowedSysctlPrefixes` satisfies a `sysctl.startsWith(prefix)` check, check succeeds
    - If `allowedSysctlExplicit` containers the sysctl (`in` keyword), check succeeds
    - else check fails

When I say "add X to the list if condition is met", that's probably best done using the `list.filter()` macro. Performing mutations, like stripping the `*` character can be done using the `list.map()` macro and use of the extended strings library.

For a conservative string replacement (only removing the very last `*`, in case there is a `*` mid-string:

- Reverse the string
- Call `string.Replace("*", "", 1)`
- Reverse the string again

Another option could be to use `strings.Split()` , pop off the last entry from the list using `lists.Slice()`, then use `strings.join()`

Alternately, if we know `*` cannot be a valid character for a sysctl (which seems likely) we can just call string.Replace() directly.

I mention "extended CEL library".

K8s has its own CEL environments, defined here:

https://github.com/kubernetes/apiserver/blob/v0.30.0/pkg/cel/environment/base.go

Which defines some k8s-specific functions. It also incorporates extensions from cel-go, defined here:

https://github.com/google/cel-go/blob/master/ext

website/docs/validation/forbidden-sysctls.md
```diff
@@ -51,65 +51,86 @@ spec:
...
+          - name: allowedSysctls
+            expression: |
+              !has(variables.params.allowedSysctls) ? [] :
+              variables.sysctls.filter(sysctl,
+                  variables.params.allowedSysctls.exists(allowedSysctl, sysctl.name == allowedSysctl || (allowedSysctl == "*") || (string(sysctl.name).matches("^" + allowedSysctl + "$"))))
+          validations:
+          - expression: '(has(request.operation) && request.operation == "UPDATE") || size(variables.sysctls) == 0 || size(variables.forbiddenSysctls) == 0 && (size(variables.allowedSysctls) == 0 || size(variables.allowedSysctls) == size(variables.sysctls))'
```
maxsmythe, line 71: Size checking should not be necessary here with above changes (other than making sure `violatingSysctls` is empty)

website/docs/validation/forbidden-sysctls.md
maxsmythe: Also adding link to CEL macros, for completeness:

https://github.com/google/cel-spec/blob/v0.6.0/doc/langdef.md#macros

ritazh: I started with `allDisallowed` but after more testing, I dont think we need it. PTAL

library/pod-security-policy/forbidden-sysctls/template.yaml
```diff
@@ -39,62 +39,97 @@ spec:
...
+            expression: |
+              (variables.sysctls.filter(sysctl,
+                (sysctl.name in variables.forbiddenSysctlExplicit ||
+                variables.forbiddenSysctlPrefixes.exists(fsp, string(sysctl.name).startsWith(fsp))) ||
+                (!variables.allAllowed &&
+                !(sysctl.name in variables.allowedSysctlExplicit) &&
+                !variables.allowedSysctlPrefixes.exists(asp, string(sysctl.name).startsWith(asp)))))
+          validations:
+          - expression: '(has(request.operation) && request.operation == "UPDATE") || size(variables.violatingSysctls) == 0'
+            messageExpression: '"The sysctl is not allowed for pod: " + variables.anyObject.metadata.name + ", forbidden: " + variables.forbiddenSysctls.map(c, c).join(", ") + ", allowed: " + variables.allowedSysctls.map(c, c).join(", ")'
```
maxsmythe, line 74: The map() statements here seem like no-ops, are they performing any heavy lifting? Does there need to be an explicit casting to string()?

library/pod-security-policy/forbidden-sysctls/template.yaml
```diff
@@ -39,62 +39,97 @@ spec:
...
+              !has(variables.params.allowedSysctls) ? [] : 
+                variables.params.allowedSysctls.filter(sysctl, !sysctl.endsWith("*"))
+          - name: forbiddenSysctlPrefixes
+            expression: |
+              !has(variables.params.forbiddenSysctls) ? [] : variables.params.forbiddenSysctls.filter(sysctl, sysctl.endsWith("*")).map(sysctl, string(sysctl).replace("*", ""))
+          - name: forbiddenSysctlExplicit
+            expression: |
+              !has(variables.params.forbiddenSysctls) ? [] : 
+                variables.params.forbiddenSysctls.filter(sysctl, !sysctl.endsWith("*"))
+          - name: allAllowed
```
maxsmythe, line 62: I don't think the Rego code considers undefined `allowedSysctls` to mean "all allowed". I think it just denies all sysctls. Worth creating a test case to validate this edge case.

maxsmythe (changes requested):

library/pod-security-policy/forbidden-sysctls/template.yaml
ritazh: Looking at the history: https://github.com/open-policy-agent/gatekeeper-library/pull/253/files#diff-b14d9c092cc66bb50ced20769a8508a11132aa8425a27e3a65666fb580b52414R117-R119

comment in the test:
> Empty allowedSysctls means none are allowed
> unspecified allowedSysctls which does not place any restrictions by itself

The rego code also does not put any restriction on unspecified allowedSysctls.

If we want to change this behavior, it is definitely not backward compatible. IMO we should keep the same logic to reduce confusion. WDYT? @maxsmythe

library/pod-security-policy/forbidden-sysctls/template.yaml
ritazh: In this case, the string() is not needed since `variables.params.forbiddenSysctls` is `string`. But I can cast it just to be safe.

maxsmythe: Either that or we can remove the `.map()` term, since it doesn't do anything

library/pod-security-policy/forbidden-sysctls/template.yaml
maxsmythe: Definitely no need to change the behavior. My reading of the Rego code was that empty `allowedSysctls` leads to a violation (despite what the test comment says).

Maybe my reading is wrong? In either case, adding a test suite for the edge case would validate uniform behavior.

But yeah, to re-iterate, VAP-CEL should behave like the Rego, just want to ensure that is the case.

ritazh: yes I adding 2 new tests (one for empty allowedSysctls and one for unspecified allowedSysctls, which showed existing rego has the same behavior as the test comment due to:

`not allowed_sysctl(sysctl)` and `allowed_sysctl(_)` never actually checks for `not input.parameters.allowedSysctls`

maxsmythe: Ah, I think I see why. The message expression below fails if the parameter is unspecified vs. empty:

`msg := sprintf("The sysctl %v is not explicitly allowed, pod: %v. Allowed sysctls: %v", [sysctl, input.review.object.metadata.name, input.parameters.allowedSysctls])`

which will cause the violation not to be thrown. I imaging replacing the message with a constant would change the behavior of that edge case. Fun times.

Sorry for the extra work. If we add the undefined vs. empty list test, at least we can make sure the edge-case behavior doesn't change in the future. Worth doing?

artifacthub/library/pod-security-policy/forbidden-sysctls/1.2.0/samples/psp-forbidden-sysctls/constraint3.yaml
```diff
@@ -0,0 +1,15 @@
...
+  match:
+    kinds:
+      - apiGroups: [""]
+        kinds: ["Pod"]
+  parameters:
+    forbiddenSysctls:
+    # - "*" # * may be used to forbid all sysctls
+    - kernel.*
+    allowedSysctls:
+    - "net.*" # allows all sysctls. allowedSysctls is optional.
```
maxsmythe, line 15: nit, comment outdated

library/pod-security-policy/forbidden-sysctls/template.yaml
ritazh: updated for cel.  

e.g. output:
> The sysctl is not allowed for pod: nginx-forbidden-sysctls-disallowed, forbidden: kernel.*, allowed: unspecified

and 

> The sysctl is not allowed for pod: nginx-forbidden-sysctls-disallowed, forbidden: kernel.*, allowed: empty

ritazh: re: rego bug, I propose to fix the message and add condition for no violations for unspecified allowedSysctls to ensure backward compat. wdyt?

maxsmythe: Is the rego not WAI? It will simply not return a message/violation if `allowedSysctls` is unspecified, which is the expected behavior.

I am definitely down for making the test for `allowedSysctls` explicit in the Rego... having an implicit logic gate buried in a string formatting statement seems a bit hard to catch for readers.

JaydipGabani (changes requested):
LGTM otherwise

library/pod-security-policy/forbidden-sysctls/template.yaml
ritazh: addressed in https://github.com/open-policy-agent/gatekeeper-library/pull/521/commits/8f9a027ebf586e2d1c2d669f7347fa5d1264632e PTAL

JaydipGabani (approved):
lgtm

maxsmythe (approved):
LGTM

PR #523, opened 2024-05-11
fsgroup cel

ritazh:

Fixes #541 


src/pod-security-policy/fsgroup/src.cel
```diff
@@ -0,0 +1,9 @@
+variables:
+- name: fsGroup
+  expression: '!has(object.spec.securityContext) ? "" : !has(object.spec.securityContext.fsGroup) ? "" : object.spec.securityContext.fsGroup'
+- name: input_fsGroup_allowed
+  expression: |
+    !has(variables.params.rule) ? true : variables.params.rule == "RunAsAny" ? true : variables.params.rule == "MayRunAs" && variables.fsGroup == "" ? true : (variables.params.rule == "MayRunAs" || variables.params.rule == "MustRunAs") && has(variables.params.ranges) && size(variables.params.ranges) > 0 ? variables.params.ranges.all(range, range.min <= variables.fsGroup && range.max >= variables.fsGroup) : false
```
JaydipGabani, line 6: seems like there might be a bug in respective rego code. If `input.params.rule` is missing the policy will deny create violations for all  objects with rego code. However, desired behavior is where we dont want to throw violations when `params.rule` is missing. 

thoughts? @ritazh @maxsmythe

src/pod-security-policy/fsgroup/src.cel
```diff
@@ -0,0 +1,9 @@
+variables:
+- name: fsGroup
+  expression: '!has(object.spec.securityContext) ? "" : !has(object.spec.securityContext.fsGroup) ? "" : object.spec.securityContext.fsGroup'
```
JaydipGabani, line 3:
```suggestion
  expression: '!has(variable.anyObject.spec.securityContext) ? "" : !has(variable.anyObject.spec.securityContext.fsGroup) ? "" : variable.anyObject.spec.securityContext.fsGroup'
```

JaydipGabani (changes requested):

src/pod-security-policy/fsgroup/src.cel
ritazh: You are correct for completeness. Though the constraint is pretty useless without the rule specified. 

For completeness, we can add the check `has_field(input.parameters, "rule")` in rego but technically it is not backward compatible if users have missing rule today.

For this, do we want to bump major or minor?

> Major version bump required: Whenever there is a breaking change in the policy e.g. updating template Kind, updating existing parameter schema, adding the requires-sync-data annotation to sync new data, or any other breaking changes
Minor version bump required: Whenever there is a backward compatible change in the policy e.g. adding a parameter, updating Rego logic

JaydipGabani: I think Major version just to be on safe side of things.

JaydipGabani (approved):
LGTM after we settle on minor or major version bump is needed for this teamplate.

library/pod-security-policy/fsgroup/template.yaml
```diff
@@ -44,62 +44,76 @@ spec:
...
-        package k8spspfsgroup
+      code:
+      - engine: K8sNativeValidation
+        source:
+          variables:
+          - name: fsGroup
+            expression: '!has(variables.anyObject.spec.securityContext) ? "" : !has(variables.anyObject.spec.securityContext.fsGroup) ? "" : variables.anyObject.spec.securityContext.fsGroup'
+          - name: input_fsGroup_allowed
+            expression: |
+              !has(variables.params.rule) ? true : variables.params.rule == "RunAsAny" ? true : variables.params.rule == "MayRunAs" && variables.fsGroup == "" ? true : (variables.params.rule == "MayRunAs" || variables.params.rule == "MustRunAs") && has(variables.params.ranges) && size(variables.params.ranges) > 0 ? variables.params.ranges.all(range, range.min <= variables.fsGroup && range.max >= variables.fsGroup) : false
```
maxsmythe, line 55: .all() or .exists()

library/pod-security-policy/fsgroup/template.yaml
```diff
@@ -44,62 +44,76 @@ spec:
...
+          validations:
+          - expression: '(has(request.operation) && request.operation == "UPDATE") || variables.input_fsGroup_allowed'
+            messageExpression: '"The provided pod spec fsGroup is not allowed, pod: " + variables.anyObject.metadata.name + ". Allowed fsGroup: " + variables.params.rule'
```
maxsmythe, line 58: The error string will end with something like "Allowed fsGroup: MustRunAs" which I don't think is what you intended.

library/pod-security-policy/fsgroup/template.yaml
maxsmythe, line 55: It looks like the rego defaults to violating if `ranges` is an empty array. This is true for:

* MustRunAs
* MayRunAs when securityContext defines an fsGroup

It may be worth setting up a gator test case to validate these edge cases.

maxsmythe (changes requested):

src/pod-security-policy/fsgroup/src.cel
maxsmythe: I cannot parse the phrase "deny create violations" what is the behavior you are describing?

This line in the `violation[]` rule shows if parameters.rule  is unset, no violation will be thrown:

`has_field(input.parameters, "rule")`

Perhaps write a test case that shows the situation-of-concern to validate whether there actually is drift?

ritazh: @maxsmythe I added `has_field(input.parameters, "rule")` in the rego https://github.com/open-policy-agent/gatekeeper-library/pull/523/files#diff-54ba54eb002245f40e88e03c102ebf4b4783d7f3f501cf73a5996119e48e062fR8 to ensure consistency between rego and cel. without it, rego will return a violation for a constraint with unspecified `rule` which would be considered as not backward compat. do we keep the same behavior for both cel and rego or do we correct this and bump the major version?

library/pod-security-policy/fsgroup/template.yaml
ritazh: New message:
> The provided pod spec fsGroup is not allowed, pod: fsgroup-disallowed. Allowed fsGroup rule: MustRunAs, allowed fsGroup ranges: empty

maxsmythe (approved):
LGTM

JaydipGabani (approved):

PR #530, opened 2024-05-22
chore: adding cel for psp/volume and psp/selinux

JaydipGabani:

Fixes #541 


src/pod-security-policy/selinux/src.cel
```diff
@@ -0,0 +1,49 @@
...
+        !(has(variables.params.allowedSELinuxOptions.type) && has(c.securityContext.seLinuxOptions.type) && (c.securityContext.seLinuxOptions.type == variables.params.allowedSELinuxOptions.type)) ||
+        !(has(variables.params.allowedSELinuxOptions.user) && has(c.securityContext.seLinuxOptions.user) && (c.securityContext.seLinuxOptions.user == variables.params.allowedSELinuxOptions.user))
+      ) : (has(c.securityContext.seLinuxOptions.level) || has(c.securityContext.seLinuxOptions.role) || has(c.securityContext.seLinuxOptions.type) || has(c.securityContext.seLinuxOptions.user))
+      ) : false
+    ))
+validations:
+- expression: '(has(request.operation) && request.operation == "UPDATE") || variables.notViolatingSELinuxOptions'
+  messageExpression: '"SELinux options is not allowed, pod: " + object.metadata.name + ". Allowed options: " + variables.params.allowedSELinuxOptions'
+- expression: '(has(request.operation) && request.operation == "UPDATE") || size(variables.badContainers) == 0'
```
JaydipGabani, line 48: @maxsmythe @ritazh any idea how to get around this error? I tried this policy on CEL playground and it works fine. So I am bit confused why this error is getting thrown.

```
unexpected number of violations: got 1 violations but want none: got messages [expression '(has(request.operation) && request.operation == "UPDATE") || size(variables.badContainers) == 0' resulted in error: composited variable "badContainers" fails to evaluate: unsupported index type 'string' in list]
```

website/docs/validation/selinux.md
```diff
@@ -68,92 +68,150 @@ spec:
...
+                    (has(opts.level) ? has(c.securityContext.seLinuxOptions.level) && (c.securityContext.seLinuxOptions.level == opts.level) : !has(c.securityContext.seLinuxOptions.level)) &&
+                    (has(opts.role) ? has(c.securityContext.seLinuxOptions.role) && (c.securityContext.seLinuxOptions.role == opts.role) : !has(c.securityContext.seLinuxOptions.role)) &&
+                    (has(opts.type) ? has(c.securityContext.seLinuxOptions.type) && (c.securityContext.seLinuxOptions.type == opts.type) : !has(c.securityContext.seLinuxOptions.type)) &&
+                    (has(opts.user) ? has(c.securityContext.seLinuxOptions.user) && (c.securityContext.seLinuxOptions.user == opts.user) : !has(c.securityContext.seLinuxOptions.user))
+                  )
+                ) : false)
+                : false)
+              )
+          validations:
+          - expression: '(has(request.operation) && request.operation == "UPDATE") || variables.notViolatingSELinuxOptions'
```
maxsmythe, line 123: this seems wrong. This check appears to pass if there is any SE linux option is allowed. Really we want the check to fail if any disallowed option is used.

website/docs/validation/selinux.md
```diff
@@ -68,92 +68,150 @@ spec:
...
-          _matches_exemption(img, exemption) {
-              endswith(exemption, "*")
-              prefix := trim_suffix(exemption, "*")
-              startswith(img, prefix)
-          }
+      code:
+      - engine: K8sNativeValidation
+        source:
+          variables:
+          - name: notViolatingSELinuxOptions
```
maxsmythe, line 75: We should invert this to read `violatingSELinuxOptions`. The absence of violations is what satisfies this check, not the presence of any single allowed option.

maxsmythe (commented):
One comment on pod-level check

website/docs/validation/selinux.md
JaydipGabani: good catch! changed the code. Should be fixed now.

sozercan (approved):
Thanks! LGTM

mounicarajput (approved):

artifacthub/library/pod-security-policy/selinux/1.1.0/template.yaml
```diff
@@ -0,0 +1,202 @@
...
+                    description: "An SELinux user."
+  targets:
+    - target: admission.k8s.gatekeeper.sh
+      code:
+      - engine: K8sNativeValidation
+        source:
+          variables:
+          - name: usingAllowedSELinuxOptions
+            expression: |
+              has(variables.anyObject.spec.securityContext) && has(variables.anyObject.spec.securityContext.seLinuxOptions) ?
```
maxsmythe, line 65: This test looks wrong. Shouldn't we be assuming an option is disallowed unless specifically allowed by parameters?

As-written, it looks like undefined parameters => allow everything, which would seem to drift from the rego.

Adding gator tests for this would validate the drift.

maxsmythe (changes requested):

artifacthub/library/pod-security-policy/selinux/1.1.0/template.yaml
JaydipGabani: Good catch! I fixed it, it should be in parity with rego now.

library/pod-security-policy/selinux/template.yaml
```diff
@@ -56,89 +56,147 @@ spec:
...
+      code:
+      - engine: K8sNativeValidation
+        source:
+          variables:
+          - name: usingAllowedSELinuxOptions
+            expression: |
+              has(variables.anyObject.spec.securityContext) && has(variables.anyObject.spec.securityContext.seLinuxOptions) ? 
+                (has(variables.params.allowedSELinuxOptions) ? 
+                (
+                  size(variables.params.allowedSELinuxOptions) > 0 && variables.params.allowedSELinuxOptions.all(opts,
```
maxsmythe, line 68: I don't think .all() is correct here, the way the rego reads, at least one provided `allowedSELinuxOptions` must be satisfied.

As always, verify with testing.

library/pod-security-policy/selinux/template.yaml
```diff
@@ -56,89 +56,147 @@ spec:
...
+                    (has(opts.level) ? has(variables.anyObject.spec.securityContext.seLinuxOptions.level) && (variables.anyObject.spec.securityContext.seLinuxOptions.level == opts.level) : !has(variables.anyObject.spec.securityContext.seLinuxOptions.level)) &&
+                    (has(opts.role) ? has(variables.anyObject.spec.securityContext.seLinuxOptions.role) && (variables.anyObject.spec.securityContext.seLinuxOptions.role == opts.role) : !has(variables.anyObject.spec.securityContext.seLinuxOptions.role)) &&
```
maxsmythe, line 70: As mentioned in previous comment, this seems incorrect. It looks like you are disabling the check if the params side is undefined, as opposed to skipping if securityContext side is undefined.

Previous comment:

```
This test looks wrong. Shouldn't we be assuming an option is disallowed unless specifically allowed by parameters?

As-written, it looks like undefined parameters => allow everything, which would seem to drift from the rego.

Adding gator tests for this would validate the drift.
```

Please add the gator tests for this scenario, relying on reviewers to catch this drift consistently is brittle.

library/pod-security-policy/selinux/template.yaml
```diff
@@ -56,89 +56,147 @@ spec:
...
+                    (has(opts.type) ? has(variables.anyObject.spec.securityContext.seLinuxOptions.type) && (variables.anyObject.spec.securityContext.seLinuxOptions.type == opts.type) : !has(variables.anyObject.spec.securityContext.seLinuxOptions.type)) &&
+                    (has(opts.user) ? has(variables.anyObject.spec.securityContext.seLinuxOptions.user) && (variables.anyObject.spec.securityContext.seLinuxOptions.user == opts.user) : !has(variables.anyObject.spec.securityContext.seLinuxOptions.user))
+                  )
+                ) :  
+                true) 
+                : true
+          - name: containers
```
maxsmythe, line 77: I have not looked at containers code, but I'm guessing it will need to be modified to comply with pod level code.

maxsmythe (changes requested):

JaydipGabani:
Not stale

PR #533, opened 2024-05-25
feat: Update apparmor: add CEL, support securityContext

maxsmythe:
For context... here is how the applicable apparmor profile is derived:

https://github.com/kubernetes/kubernetes/blob/master/pkg/security/apparmor/helpers.go#L55-L76


maxsmythe:
Looks like there may be a bug in the Rego side... I'll take a look in a bit

maxsmythe:
Rego fixed ... pod-level security context retrieval did not take into account that it was a child of `spec`

JaydipGabani:
@maxsmythe do we want to update the policy version to `1.1.0` since we are updating rego and adding cel as well?

maxsmythe:
@JaydipGabani SGTM. Is there anything special we need to do for that? Or just update the annotation?

JaydipGabani:
> @JaydipGabani SGTM. Is there anything special we need to do for that? Or just update the annotation?

@maxsmythe Updating annotations and running `make generate-all` should work. It will generate files for artifact-hub as well.

maxsmythe:
bumped minor version

JaydipGabani:
Fixes #541

src/pod-security-policy/apparmor/src.cel
```diff
@@ -0,0 +1,49 @@
...
+        variables.podAppArmor.type == "Unconfined" ? "unconfined" : 
+          variables.podAppArmor.type == "Localhost" ? "localhost/" + variables.podAppArmor.localhostProfile : ""
+- name: appArmorByContainer
+  expression: |
+    variables.allContainers.map(container, [container.name,
+      has(container.securityContext) && has(container.securityContext.appArmorProfile) ?
+        (container.securityContext.appArmorProfile.type == "RuntimeDefault" ? "runtime/default" :
+          container.securityContext.appArmorProfile.type == "Unconfined" ? "unconfined" : 
+            container.securityContext.appArmorProfile.type == "Localhost" ? "localhost/" + container.securityContext.appArmorProfile.localhostProfile : "") :
+        has(variables.anyObject.metadata.annotations) && has(variables.anyObject.metadata.annotations) && ("container.apparmor.security.beta.kubernetes.io/" + container.name) in variables.anyObject.metadata.annotations ?
```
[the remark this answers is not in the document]
maxsmythe: good find!

JaydipGabani (approved):
LGTM on CEL

maxsmythe:
I'm guessing the k8s tests are failing because securityContext.appArmorProfile only exists as of k8s 1.30?

JaydipGabani:
@maxsmythe we are running tests with [kind: 0.17.0](https://github.com/kubernetes-sigs/kind/releases/tag/v0.17.0) and that seems to have k8s 1.25.3. So I agree, I think that would be the reason as well.

website/docs/validation/apparmor.md
```diff
@@ -54,61 +54,146 @@ spec:
...
+            expression: 'has(variables.anyObject.spec.securityContext) && has(variables.anyObject.spec.securityContext.appArmorProfile) ? variables.anyObject.spec.securityContext.appArmorProfile : null'
+          - name: canonicalPodAppArmor
+            expression: |
+              variables.podAppArmor == null ? "runtime/default" : 
+                variables.podAppArmor.type == "RuntimeDefault" ? "runtime/default" :
+                  variables.podAppArmor.type == "Unconfined" ? "unconfined" : 
+                    variables.podAppArmor.type == "Localhost" ? "localhost/" + variables.podAppArmor.localhostProfile : ""
+          - name: appArmorByContainer
+            expression: |
+              variables.allContainers.map(container, [container.name,
```
maxsmythe, line 79: is container.name unique across containers/ephemeralContainers/initContainers ? We should probably track these separately per-field.

JaydipGabani: Yes, container.name is unique across all containers.

maxsmythe: Thanks!

I think I still want to change this, as `gator` will not have the K8s API server enforcing that rule.

JaydipGabani: @maxsmythe I am not sure how this change makes any difference as the cel code it-self doesnt check for unique names across all types of containers. can you explain?

JaydipGabani: nvm, I think I got it.

artifacthub/library/pod-security-policy/apparmor/1.1.0/samples/psp-apparmor/example_allowed.yaml
```diff
@@ -0,0 +1,13 @@
+apiVersion: v1
+kind: Pod
+metadata:
+  name: nginx-apparmor-allowed
+  annotations:
+    # apparmor.security.beta.kubernetes.io/pod: unconfined # runtime/default
+    container.apparmor.security.beta.kubernetes.io/nginx: localhost/custom
```
sozercan, line 7: looks like this is deprecated, do we want to add `appArmorProfile` to this PR or create a v2 similar to seccomp?

```
Warning: metadata.annotations[container.apparmor.security.beta.kubernetes.io/nginx]: deprecated since v1.30; use the "appArmorProfile" field instead
```

sozercan: nevermind, i see we support both

sozercan (approved):
Thanks! LGTM

PR #537, opened 2024-05-30
chore: adding psp-users CEL policy

JaydipGabani:

Fixes #541 

Rego policy nehavior is - 

- `filed` in below list is reference to securityContext fields - `[ runAsUser, runAsGroup, fsGroup, supplementalGroup ]`
-------------------------------
- Input:

```
field:
      rule: MustRunAs
      ranges:
        - min: 100
          max: 200
```
- Behavior: if `field` is missing from object then throw missing violation, else throw violation is `field` is not in required range
 -------------------------------------
- Input:
```
runAsUser:
      rule: MustRunAsNonRoot
```
- Behavior: if `runAsUser` and `runAsNonRoot` both are missing from object then throw missing violation, else throw violation is `runAsUser == 0`
--------------------------------------------------
- Input:
```
field:
      rule: mayRunAs
      ranges:
        - min: 100
          max: 200
```
- Behavior: No missing field violation, but is field is present and violating the range then throw the violation
--------------------------------------------------
- Input:
```
field:
  rule: RunAsAny

```
- Behavior: No violations
--------------------------------------------------


library/pod-security-policy/users/template.yaml
```diff
@@ -132,158 +132,341 @@ spec:
...
+          - name: missingRequiredRunAsUserContainers
+            expression: |
+              variables.badContainers.filter(container, 
+                has(variables.params.runAsUser) && has(variables.params.runAsUser.rule) && (variables.params.runAsUser.rule == "MustRunAs") &&
+                !((has(container.securityContext) && has(container.securityContext.runAsUser)) || 
+                  (variables.podRunAsUser != null)))
+          - name: missingRequiredRunAsNonRootContainers
+            expression: |
+              variables.badContainers.filter(container, 
+                has(variables.params.runAsUser) && has(variables.params.runAsUser.rule) && (variables.params.runAsUser.rule == "MustRunAsNonRoot") && ((has(container.securityContext) && !(has(container.securityContext.runAsUser) || has(container.securityContext.runAsNonRoot))) ?
```
maxsmythe, line 176: We need to check that runAsNonRoot is both defined and equal to `true` (this is implied in the Rego check, since `not` will invert both "missing" and "false" to true.

Also, `has(container.securityContext.runAsUser) || has(container.securityContext.runAsNonRoot)` should be `has(container.securityContext.runAsUser) && has(container.securityContext.runAsNonRoot)` , since both conditions need to be satisfied.

library/pod-security-policy/users/template.yaml
```diff
@@ -132,158 +132,341 @@ spec:
...
+                  (has(variables.anyObject.spec.securityContext) && !(has(variables.anyObject.spec.securityContext.runAsUser) || has(variables.anyObject.spec.securityContext.runAsNonRoot))) : false
+                ))
+          - name: processedRunAsUserContainers
+            expression: (variables.missingRequiredRunAsNonRootContainers + variables.missingRequiredRunAsUserContainers).map(container, container.name)
+          - name: invalidRunAsUserContainers
+            expression: |
+              variables.badContainers.filter(container, 
+                !(container.name in variables.processedRunAsUserContainers) &&
```
maxsmythe, line 184: we should not assume that container names are globally unique, since tools like `gator` may not perform that kind of validation.

library/pod-security-policy/users/template.yaml
```diff
@@ -132,158 +132,341 @@ spec:
...
+                  (
+                    variables.params.runAsUser.rule == "MustRunAsNonRoot" ?
+                    (
+                      has(container.securityContext) && has(container.securityContext.runAsUser) ? (container.securityContext.runAsUser == 0) : 
+                        (variables.podRunAsUser == null) || variables.podRunAsUser == 0
+                    ) :
+                    (
+                      variables.params.runAsUser.rule == "MustRunAs" ?
+                      (
+                        has(container.securityContext) && has(container.securityContext.runAsUser) ? !variables.params.runAsUser.ranges.all(range, container.securityContext.runAsUser >= range.min && container.securityContext.runAsUser <= range.max) :
```
maxsmythe, line 197: I think this should be `exists()` instead of `all()`, since the constraint is satisfied as long as the user ID lies within at least one of the specified ranges.

library/pod-security-policy/users/template.yaml
```diff
@@ -132,158 +132,341 @@ spec:
...
+                !(variables.missingRequiredRunAsGroupContainers.exists(c, c.name == container.name)) &&
+                (
+                  has(variables.params.runAsGroup) && has(variables.params.runAsGroup.rule) ? 
+                  (
+                    variables.params.runAsGroup.rule == "RunAsAny" ? false :
+                    (
+                      (variables.params.runAsGroup.rule == "MustRunAs" || variables.params.runAsGroup.rule == "MayRunAs") && 
+                      (
+                        has(container.securityContext) && has(container.securityContext.runAsGroup) ? 
+                          !variables.params.runAsGroup.ranges.all(range, container.securityContext.runAsGroup >= range.min && container.securityContext.runAsGroup <= range.max) :
```
maxsmythe, line 228: Same comment about `all()` vs. `exists()`

library/pod-security-policy/users/template.yaml
```diff
@@ -132,158 +132,341 @@ spec:
...
+                (
+                  has(variables.params.runAsGroup) && has(variables.params.runAsGroup.rule) ? 
+                  (
+                    variables.params.runAsGroup.rule == "RunAsAny" ? false :
+                    (
+                      (variables.params.runAsGroup.rule == "MustRunAs" || variables.params.runAsGroup.rule == "MayRunAs") && 
+                      (
+                        has(container.securityContext) && has(container.securityContext.runAsGroup) ? 
+                          !variables.params.runAsGroup.ranges.all(range, container.securityContext.runAsGroup >= range.min && container.securityContext.runAsGroup <= range.max) : 
+                          variables.podRunAsGroup == null || !variables.params.runAsGroup.ranges.all(range, variables.podRunAsGroup >= range.min && variables.podRunAsGroup <= range.max)
```
maxsmythe, line 229: Same comment about `all()` vs. `exists()`

library/pod-security-policy/users/template.yaml
```diff
@@ -132,158 +132,341 @@ spec:
...
+                    variables.params.runAsUser.rule == "MustRunAsNonRoot" ?
+                    (
+                      has(container.securityContext) && has(container.securityContext.runAsUser) ? (container.securityContext.runAsUser == 0) : 
+                        (variables.podRunAsUser == null) || variables.podRunAsUser == 0
+                    ) :
+                    (
+                      variables.params.runAsUser.rule == "MustRunAs" ?
+                      (
+                        has(container.securityContext) && has(container.securityContext.runAsUser) ? !variables.params.runAsUser.ranges.all(range, container.securityContext.runAsUser >= range.min && container.securityContext.runAsUser <= range.max) :
+                          variables.podRunAsUser == null || !variables.params.runAsUser.ranges.all(range, variables.podRunAsUser >= range.min && variables.podRunAsUser <= range.max)
```
maxsmythe, line 198: Same comment about `all()` vs. `exists()`

library/pod-security-policy/users/template.yaml
```diff
@@ -132,158 +132,341 @@ spec:
...
+            expression: |
+              variables.badContainers.filter(container, 
+                has(variables.params.fsGroup) && has(variables.params.fsGroup.rule) && (variables.params.fsGroup.rule == "MustRunAs") &&
+                !((has(container.securityContext) && has(container.securityContext.fsGroup)) || 
+                  (variables.podRunAsFsGroup != null))
+              )
+          - name: invalidRunAsFsGroupContainers
+            expression: |
+              variables.badContainers.filter(container, 
+                !(variables.missingRequiredFsGroupContainers.exists(c, c.name == container.name)) &&
```
maxsmythe, line 251: as above, let's not rely on names being globally unique

library/pod-security-policy/users/template.yaml
```diff
@@ -132,158 +132,341 @@ spec:
...
+                (
+                  has(variables.params.fsGroup) && has(variables.params.fsGroup.rule) ? 
+                  (
+                    variables.params.fsGroup.rule == "RunAsAny" ? false :
+                    (
+                      (variables.params.fsGroup.rule == "MustRunAs" || variables.params.fsGroup.rule == "MayRunAs") && 
+                      (
+                        has(container.securityContext) && has(container.securityContext.fsGroup) ? 
+                          !variables.params.fsGroup.ranges.all(range, container.securityContext.fsGroup >= range.min && container.securityContext.fsGroup <= range.max) :
```
maxsmythe, line 260: Same comment about `all()` vs. `exists()`

library/pod-security-policy/users/template.yaml
```diff
@@ -132,158 +132,341 @@ spec:
...
+                (
+                  has(variables.params.fsGroup) && has(variables.params.fsGroup.rule) ? 
+                  (
+                    variables.params.fsGroup.rule == "RunAsAny" ? false :
+                    (
+                      (variables.params.fsGroup.rule == "MustRunAs" || variables.params.fsGroup.rule == "MayRunAs") && 
+                      (
+                        has(container.securityContext) && has(container.securityContext.fsGroup) ? 
+                          !variables.params.fsGroup.ranges.all(range, container.securityContext.fsGroup >= range.min && container.securityContext.fsGroup <= range.max) : 
+                          variables.podRunAsFsGroup == null || !variables.params.fsGroup.ranges.all(range, variables.podRunAsFsGroup >= range.min && variables.podRunAsFsGroup <= range.max)
```
maxsmythe, line 261: Same comment about `all()` vs. `exists()`

library/pod-security-policy/users/template.yaml
```diff
@@ -132,158 +132,341 @@ spec:
...
+                !(variables.missingRequiredSupplementalGroupsContainers.exists(c, c.name == container.name)) &&
+                (
+                  has(variables.params.supplementalGroups) && has(variables.params.supplementalGroups.rule) ? 
+                  (
+                    variables.params.supplementalGroups.rule == "RunAsAny" ? false :
+                    (
+                      (variables.params.supplementalGroups.rule == "MustRunAs" || variables.params.supplementalGroups.rule == "MayRunAs") && 
+                      (
+                        has(container.securityContext) && has(container.securityContext.supplementalGroups) ? 
+                          !variables.params.supplementalGroups.ranges.all(range, container.securityContext.supplementalGroups.all(gp, gp>= range.min && gp <= range.max)) :
```
maxsmythe, line 291: Same comment about `all()` vs. `exists()`

library/pod-security-policy/users/template.yaml
```diff
@@ -132,158 +132,341 @@ spec:
...
+                (
+                  has(variables.params.supplementalGroups) && has(variables.params.supplementalGroups.rule) ? 
+                  (
+                    variables.params.supplementalGroups.rule == "RunAsAny" ? false :
+                    (
+                      (variables.params.supplementalGroups.rule == "MustRunAs" || variables.params.supplementalGroups.rule == "MayRunAs") && 
+                      (
+                        has(container.securityContext) && has(container.securityContext.supplementalGroups) ? 
+                          !variables.params.supplementalGroups.ranges.all(range, container.securityContext.supplementalGroups.all(gp, gp>= range.min && gp <= range.max)) : 
+                          variables.podRunAsSupplementalGroups == null || !variables.params.supplementalGroups.ranges.all(range, variables.podRunAsSupplementalGroups.all(gp, gp >= range.min && gp <= range.max))
```
maxsmythe, line 292: Same comment about `all()` vs. `exists()`

maxsmythe (changes requested):

library/pod-security-policy/users/template.yaml
JaydipGabani: Processing each container type separately.

JaydipGabani:
@maxsmythe @ritazh @sozercan PTAL

library/pod-security-policy/users/template.yaml
```diff
@@ -132,158 +132,640 @@ spec:
...
+              variables.containers.filter(container, 
+                !(container.image in variables.exemptImages) &&
+                has(variables.params.runAsUser) && has(variables.params.runAsUser.rule) && (variables.params.runAsUser.rule == "MustRunAs") &&
+                !((has(container.securityContext) && has(container.securityContext.runAsUser)) || 
+                  (variables.podRunAsUser != null)))
+          - name: missingRequiredRunAsNonRootContainers
+            expression: |
+              variables.containers.filter(container, 
+                !(container.image in variables.exemptImages) &&
+                has(variables.params.runAsUser) && has(variables.params.runAsUser.rule) && (variables.params.runAsUser.rule == "MustRunAsNonRoot") ?
```
maxsmythe, line 184: I don't think you need a trinary operator (`?`) here

library/pod-security-policy/users/template.yaml
```diff
@@ -132,158 +132,640 @@ spec:
...
+                (
+                  (!has(container.securityContext) || (
+                    (!has(container.securityContext.runAsNonRoot) || !container.securityContext.runAsNonRoot) && (!has(container.securityContext.runAsUser) || container.securityContext.runAsUser == 0) 
+                  )) || variables.missingRunAsNonRootGlobal
```
maxsmythe, line 188: This reads like it will be true if either there is no global definition for running as non root OR if there is no local definition for running as non root.

This is incorrect. If local is defined but global is not, then there is no violation, as local definition supersedes.

library/pod-security-policy/users/template.yaml
```diff
@@ -132,158 +132,640 @@ spec:
...
+          - name: invalidRunAsUserContainers
+            expression: |
+              variables.containers.filter(container, 
+                !(container.image in variables.exemptImages) &&
+                !(container.name in variables.processedRunAsUserContainers) &&
+                has(variables.params.runAsUser) && has(variables.params.runAsUser.rule) ? 
+                (
+                  variables.params.runAsUser.rule == "RunAsAny" ? false :
+                  (
+                    variables.params.runAsUser.rule == "MustRunAsNonRoot" ?
```
maxsmythe, line 202: I think the code for this will be much simpler if we:

* have a variable for each rule type (`MustRunAs`, `MayRunAs`, etc.) that returns the containers violating the rule.
* if the rule type is not used, this variable will return an empty list.

This will allow you to:

* remove all code duplication (can go back to concatenating all containers again, since you are no longer required to filter based off container name)
* Reduce the nesting/branching for each variable.

Because this will significantly change/shorten the code, I'll stop reviewing here.

maxsmythe (changes requested):

library/pod-security-policy/users/template.yaml
JaydipGabani: Updated the code to correct this.

library/pod-security-policy/users/template.yaml
JaydipGabani: Thanks for the suggestion. Made an update to CEL code, it should now be much more compact. PTAL.

library/pod-security-policy/users/template.yaml
```diff
@@ -132,158 +132,640 @@ spec:
...
+            expression: 'has(variables.anyObject.spec.ephemeralContainers) ? variables.anyObject.spec.ephemeralContainers : []'
+          - name: exemptImagePrefixes
+            expression: |
+              !has(variables.params.exemptImages) ? [] :
+                variables.params.exemptImages.filter(image, image.endsWith("*")).map(image, string(image).replace("*", ""))
+          - name: exemptImageExplicit
+            expression: |
+              !has(variables.params.exemptImages) ? [] : 
+                variables.params.exemptImages.filter(image, !image.endsWith("*"))
+          - name: exemptImages
```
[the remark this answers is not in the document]
maxsmythe: this still needs fixing

library/pod-security-policy/users/template.yaml
```diff
@@ -132,158 +132,222 @@ spec:
...
+                variables.exemptImagePrefixes.exists(exemption, string(container.image).startsWith(exemption)))
+          - name: podRunAsUser
+            expression: |
+              variables.anyObject.kind == "Pod" && has(variables.anyObject.spec.securityContext) && has(variables.anyObject.spec.securityContext.runAsUser) ? variables.anyObject.spec.securityContext.runAsUser : null
+          - name: podSupplementalGroups
+            expression: |
+              variables.anyObject.kind == "Pod" && has(variables.anyObject.spec.securityContext) && has(variables.anyObject.spec.securityContext.supplementalGroups) ? variables.anyObject.spec.securityContext.supplementalGroups : null
+          - name: podRunAsGroup
+            expression: |
+              variables.anyObject.kind == "Pod" && has(variables.anyObject.spec.securityContext) && has(variables.anyObject.spec.securityContext.runAsGroup) ? variables.anyObject.spec.securityContext.runAsGroup : null
```
maxsmythe, line 166: Can we just put the test for kind == Pod into a `matchCondition`?

I don't think this constraint makes sense for anything other than Pods... we can add the same test to the Rego if necessary.

library/pod-security-policy/users/template.yaml
```diff
@@ -132,158 +132,222 @@ spec:
...
+            expression: |
+              has(variables.params.runAsUser) && has(variables.params.runAsUser.rule) && variables.params.runAsUser.rule == "MustRunAs" ? variables.nonExemptContainers.filter(container, (!has(container.securityContext) || !has(container.securityContext.runAsUser)) && variables.podRunAsUser == null).map(container, "Container " + container.name + " is attempting to run without a required securityContext/runAsUser") + variables.nonExemptContainers.filter(container, (has(container.securityContext) && has(container.securityContext.runAsUser) ? !variables.params.runAsUser.ranges.exists(range, container.securityContext.runAsUser >= range.min && container.securityContext.runAsUser <= range.max) : variables.podRunAsUser != null && !variables.params.runAsUser.ranges.exists(range, variables.podRunAsUser >= range.min && variables.podRunAsUser <= range.max))).map(container, "Container " + container.name + " is attempting to run as disallowed user. Allowed runAsUser: {ranges: [" + variables.params.runAsUser.ranges.map(range, "{max: " + string(range.max) + ", min: " + string(range.min) + "}").join(", ") + ", rule: " + variables.params.runAsUser.rule + "}") : []
```
maxsmythe, line 178: Can we use some string/block formatting to make the various declarations more legible? It's hard to parse the assertions without formatting. This is a general comment for all of these larger code blocks.

TBH I can't review without formatting improvements... the statements bleed together.

maxsmythe (changes requested):

library/pod-security-policy/users/template.yaml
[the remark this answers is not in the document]
JaydipGabani: Fixed it.

library/pod-security-policy/users/template.yaml
JaydipGabani: wdym by "put test for kind == Pod into a `matchCondition`?

I removed the condition `kind == Pod` from CEL and rego.

library/pod-security-policy/users/template.yaml
JaydipGabani: Appologies for putting the large code block in one line. Formatted all code large code blocks. Should be easier to parse and read.

library/pod-security-policy/users/template.yaml
maxsmythe: `matchCondition` is a field in VAP:

https://kubernetes.io/docs/reference/access-authn-authz/validating-admission-policy/#matching-requests-matchconditions

library/pod-security-policy/users/template.yaml
maxsmythe: thanks!

library/pod-security-policy/users/template.yaml
JaydipGabani: Updated CEL to add `matchCondition`.

JaydipGabani:
@maxsmythe PTAL.

library/pod-security-policy/users/template.yaml
```diff
@@ -132,158 +132,305 @@ spec:
...
+                  ", rule: " + variables.params.runAsUser.rule + "}"
+                ) : 
+                []
+          - name: violatingMustOrMayRunAsGroup
+            expression: |
+              has(variables.params.runAsGroup) && has(variables.params.runAsGroup.rule) && (variables.params.runAsGroup.rule == "MustRunAs" || variables.params.runAsGroup.rule == "MayRunAs") ? 
+                variables.nonExemptContainers.filter(container, 
+                  (!has(container.securityContext) || !has(container.securityContext.runAsGroup)) && variables.podRunAsGroup == null
+                ).map(container, 
+                  "Container " + container.name + " is attempting to run without a required securityContext/runAsGroup. Allowed runAsGroup: {ranges: [" +
```
maxsmythe, line 206: Isn't this fine for MayRunAs?

IIRC undefined is fine for MayRunAs

maxsmythe (changes requested):

JaydipGabani:
Not stale

PR #540, opened 2024-06-03
chore: adding cel for psp-seccomp policy

JaydipGabani:

Fixes #541 


src/pod-security-policy/seccomp/src.cel
```diff
@@ -0,0 +1,186 @@
...
+      !variables.hasPodAnnotations 
+    ).map(container, {
+      "container" : container.name,
+      "profile" : dyn("not configured"),
+      "file" : dyn(""),
+      "location" : dyn("no explicit profile found"),
+    })
+- name: allBadContainerProfiles
+  expression: |
+    variables.podAnnotationsProfiles + variables.containerAnnotationsProfiles + variables.podSecurityContextProfiles + variables.containerSecurityContextProfiles + variables.containerProfilesMissing
```
maxsmythe, line 131: Kubernetes no longer reads security contexts from the annotation, we should give constraint authors the ability to turn off relying on the annotation in order to avoid a security hole where pod runners "pretend" to have the correct security context.

I.e. we should add a parameter to turn off reading from the annotation.

Here is the k8s code for retrieving the seccomp profile:

https://github.com/kubernetes/kubernetes/blob/c6b5191c37f939d2d61e76de222a96ae5f5d9558/pkg/kubelet/kuberuntime/helpers.go#L270-L291

maxsmythe (changes requested):

src/pod-security-policy/seccomp/src.cel
JaydipGabani: Added v2 for seccomp with a variable to enable/disable reading from annotations.

src/pod-security-policy/seccomp/src.cel
```diff
@@ -0,0 +1,186 @@
+variables:
+- name: namingTranslation
```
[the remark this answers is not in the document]
JaydipGabani: @maxsmythe Modified the code to track one directional translation from sc to annotations, and added bits for canonicalization of profile read from the input objects.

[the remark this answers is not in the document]
JaydipGabani: Fixed some bugs and implemented suggested changes.

src/pod-security-policy/seccomp/src.cel
```diff
@@ -0,0 +1,186 @@
...
+- name: exemptImageExplicit
+  expression: |
+    !has(variables.params.exemptImages) ? [] : 
+      variables.params.exemptImages.filter(image, !image.endsWith("*"))
+- name: exemptImages
+  expression: |
+    (variables.containers + variables.initContainers + variables.ephemeralContainers).filter(container,
+      container.image in variables.exemptImageExplicit ||
+      variables.exemptImagePrefixes.exists(exemption, string(container.image).startsWith(exemption)))
+- name: badContainers
```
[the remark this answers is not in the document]
JaydipGabani: changes the variable name

src/pod-security-policy/seccomp/src.cel
maxsmythe: If we are creating "v2", we should just remove annotations from v2 altogether, since K8s no longer supports seccomp annotations

src/pod-security-policy/seccomp/src.rego
```diff
@@ -188,3 +152,18 @@ input_containers[container.name] = container {
 input_containers[container.name] = container {
     container := input.review.object.spec.ephemeralContainers[_]
 }
+
+canonicalize_seccomp_profile(profile) = out {
+    profile.type == "RuntimeDefault"
+    out := "runtime/default"
```
maxsmythe, line 158: Per existing code, this could also count as `docker/default`

src/pod-security-policy/seccomp/src.rego
```diff
@@ -7,15 +7,9 @@ container_annotation_key_prefix = "container.seccomp.security.alpha.kubernetes.i
 pod_annotation_key = "seccomp.security.alpha.kubernetes.io/pod"
 
 naming_translation = {
-    # securityContext -> annotation
```
maxsmythe, line 10: `naming_translation` can go away entirely

src/pod-security-policy/seccompv2/constraint.tmpl
```diff
@@ -0,0 +1,85 @@
...
+                It is recommended that users use the fully-qualified Docker image name (e.g. start with a domain name)
+                in order to avoid unexpectedly exempting images from an untrusted repository.
+              type: array
+              items:
+                type: string
+            allowedProfiles:
+              type: array
+              description: >-
+                An array of allowed profile values for seccomp on Pods/Containers.
+
```
maxsmythe, line 47: If we're doing a v2, since annotations are removed from K8s, we can make everyone's lives easier by only supporting the new syntax.

This will make the API match K8s's APIs 1:1.

It will also eliminate all need for translation logic, streamlining the code.

If users still need to enforce against annotations, they can keep using v1. Annotation CT will not bitrot, since that API no longer exists.

src/pod-security-policy/seccomp/src.rego
```diff
@@ -99,7 +70,7 @@ get_allowed_profiles[allowed] {
 # The simply translated profiles
 get_allowed_profiles[allowed] {
     profile := input.parameters.allowedProfiles[_]
-    not startswith(lower(profile), "localhost")
```
maxsmythe, line 102: Instead of keeping `naming_translation` we can:

* if there is a "/" in the name, use the name directly (assume its already in annotation format)
* If there is no "/" in the name, build a profile object (via a function call), then call canonicalize_seccomp_profile

This ensures we use the same translation logic everywhere.

src/pod-security-policy/seccomp/src.rego
JaydipGabani: Would it be better to translate from annotation style to SC style then, since `docker/default` and `runtime/default` both can be mapped to `RuntimeDefault`?

JaydipGabani: But this wouldn't work for `Localhost` type.

src/pod-security-policy/seccomp/src.rego
JaydipGabani: removed `naming_translation`

src/pod-security-policy/seccompv2/constraint.tmpl
JaydipGabani: Makes sense, updated v2 to remove reading from annotations altogether.

src/pod-security-policy/seccomp/src.rego
JaydipGabani: Had to implement a workaround in translation logic. But I think now the policy should be good.

src/pod-security-policy/seccompv2/src.cel
```diff
@@ -0,0 +1,122 @@
...
+  expression: |
+    !has(variables.params.allowedProfiles) ? [] : variables.params.allowedProfiles
+- name: allowedLocalhostFiles
+  expression: |
+    has(variables.params.allowedLocalhostFiles) ? variables.params.allowedLocalhostFiles : []
+- name: allowedProfilesTranslation
+  expression: |
+    (variables.inputAllowedProfiles.filter(profile,
+    profile != "Localhost").map(profile, profile == "Unconfined" ? "unconfined" : profile)) + 
+    (variables.inputAllowedProfiles.exists(profile, profile == "RuntimeDefault") ? ["runtime/default", "docker/default"] : [])
```
maxsmythe, line 39: No need to do the mapping for the v2 policy, that may simplify the code a bit.

JaydipGabani: Do you mean we should assume we have only SC style profile names in the input.params.allowedProfile since we are not using annotations in v2? 
if that is the case, then we would not need to canonicalize `obj.securityContext.seccompProfile.type` as well.

maxsmythe: Yep!

website/docs/validation/seccompv2.md
```diff
@@ -0,0 +1,496 @@
+---
+id: seccompv2
+title: Seccomp V2
+---
+
+# Seccomp V2
+
+## Description
+Controls the seccomp profile used by containers. Corresponds to the `seccomp.security.alpha.kubernetes.io/allowedProfileNames` annotation on a PodSecurityPolicy. For more information, see https://kubernetes.io/docs/concepts/policy/pod-security-policy/#seccompv2
```
sozercan, line 9: what's different between non-v2 and v2?

sozercan: looks like difference is the annotation. can we document it here?

artifacthub/library/pod-security-policy/seccompv2/1.0.0/artifacthub-pkg.yml
```diff
@@ -0,0 +1,22 @@
+version: 1.0.0
+name: k8spspseccompv2
+displayName: Seccomp V2
+createdAt: "2024-06-13T18:28:19Z"
+description: Controls the seccomp profile used by containers. Corresponds to the `seccomp.security.alpha.kubernetes.io/allowedProfileNames` annotation on a PodSecurityPolicy. For more information, see https://kubernetes.io/docs/concepts/policy/pod-security-policy/#seccompv2
```
sozercan, line 5: we should update the description too

JaydipGabani: Updated description.

website/docs/validation/seccompv2.md
JaydipGabani: Updated description.

sozercan (commented):
doc change request, otherwise LGTM

sozercan (approved):
Thanks! LGTM

JaydipGabani:
@maxsmythe @ritazh PTAL

library/pod-security-policy/seccompv2/template.yaml
```diff
@@ -0,0 +1,303 @@
...
+                in order to avoid unexpectedly exempting images from an untrusted repository.
+              type: array
+              items:
+                type: string
+            allowedProfiles:
+              type: array
+              description: >-
+                An array of allowed profile values for seccomp on Pods/Containers.
+
+                Can use the annotation naming scheme: `runtime/default`, `docker/default`, `unconfined` and/or
```
maxsmythe, line 39: We should get rid of mention/use of the annotation naming scheme for V2, since that is no longer a recognized paradigm for K8s

library/pod-security-policy/seccompv2/template.yaml
```diff
@@ -0,0 +1,303 @@
...
+            # The profiles explicitly in the list
+            get_allowed_profiles[allowed] {
+                allowed := input.parameters.allowedProfiles[_]
+            }
+
+            get_allowed_profiles[allowed] {
+                profile := input.parameters.allowedProfiles[_]
+                profile == "Localhost"
+                file := object.get(input.parameters, "allowedLocalhostFiles", [])[_]
+                allowed := sprintf("Localhost/%s", [file])
```
maxsmythe, line 239: rather than attempting string packing/unpacking, we can just mimic the structure of local host profiles, this will simplify comparison later on.

basically, `allowed := {"type": allowed, "localHostProfile": file}`

Similar for above.

Then `get_profile()` is also simplified, as is the comparison with localhost. We also are no longer dependent on our scratch string format being invalidated.

library/pod-security-policy/seccompv2/template.yaml
```diff
@@ -0,0 +1,303 @@
...
+                variables.exemptImagePrefixes.exists(exemption, string(container.image).startsWith(exemption)))
+          - name: unverifiedContainers
+            expression: |
+              (variables.containers + variables.initContainers + variables.ephemeralContainers).filter(container,
+                !variables.allowAllProfiles &&
+                !(container.image in variables.exemptImages))
+          - name: inputAllowedProfiles
+            expression: |
+              !has(variables.params.allowedProfiles) ? [] : variables.params.allowedProfiles
+          - name: derivedAllowedLocalhostFiles
```
maxsmythe, line 100: like the Rego code, we should avoid using string transformations here, just structs.

In this case, we can keep the `params.allowedLocalhostFiles` strings unmodified.

library/pod-security-policy/seccompv2/template.yaml
```diff
@@ -0,0 +1,303 @@
...
+          - name: allowAllLocalhostFiles
+            expression: |
+              variables.allowedProfiles.exists(profile, profile == "Localhost/*")
+          - name: hasPodSeccomp
+            expression: |
+              has(variables.anyObject.spec.securityContext) && has(variables.anyObject.spec.securityContext.seccompProfile)
+          - name: podLocalHostProfile
+            expression: |
+              variables.hasPodSeccomp && has(variables.anyObject.spec.securityContext.seccompProfile.localhostProfile) ? variables.anyObject.spec.securityContext.seccompProfile.localhostProfile : ""
+          - name: podSecurityContextProfile
```
maxsmythe, line 115: nit: `podSecurityContextProfileType`, that is what is actually being stored, and otherwise you are relying on users to read a plural (extra "s") to distinguish between two very different bits of data.

library/pod-security-policy/seccompv2/template.yaml
```diff
@@ -0,0 +1,303 @@
...
+              variables.allContainerProfiles.filter(badContainerProfile, 
+                  badContainerProfile.profile != "Localhost" &&
+                  !(badContainerProfile.profile in variables.allowedProfiles)
+              ).map(badProfile, "Seccomp profile '" + badProfile.profile + "' is not allowed for container '" + badProfile.container + "'. Found at: " + badProfile.location + ". Allowed profiles: " + variables.allowedProfiles.join(", "))
+          - name: badContainerProfilesWithFiles
+            expression: |
+              variables.allContainerProfiles.filter(badContainerProfile, 
+                badContainerProfile.profile == "Localhost" &&
+                !variables.allowAllLocalhostFiles &&
+                !((badContainerProfile.profile + "/" + badContainerProfile.file) in variables.allowedProfiles)
```
maxsmythe, line 165: as above, prefer field comparison to string manipulation

maxsmythe (changes requested):
Couple comments on v2, have not reviewed v1. Looking close!

library/pod-security-policy/seccomp/template.yaml
```diff
@@ -67,215 +67,349 @@ spec:
...
+          - name: allowSecurityContextLocalhost
+            expression: |
+              variables.inputAllowedProfiles.exists(profile, profile == "Localhost")
+          - name: derivedAllowedLocalhostFiles
+            expression: |
+              variables.allowSecurityContextLocalhost ? variables.params.allowedLocalhostFiles.map(file, "localhost/" + file) : []
+          - name: localhostWildcardAllowed
+            expression: |
+              variables.inputAllowedProfiles.exists(profile, profile == "localhost/*") || variables.derivedAllowedLocalhostFiles.exists(profile, profile == "localhost/*")
+          - name: allowedProfiles
```
maxsmythe, line 121: What about users who input `localhost/<some profile>`? It looks like that's getting dropped?

maxsmythe (changes requested):
v2 looks good modulo docstring comment.



v1 appears to have a bug WRT localhost

library/pod-security-policy/seccompv2/template.yaml
```diff
@@ -0,0 +1,301 @@
...
+              type: array
+              items:
+                type: string
+            allowedProfiles:
+              type: array
+              description: >-
+                An array of allowed profile values for seccomp on Pods/Containers.
+
+                Can use the securityContext naming scheme: `RuntimeDefault`, `Unconfined`
+                and/or `Localhost`. For securityContext `Localhost`, use the parameter `allowedLocalhostProfiles`
```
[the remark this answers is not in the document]
JaydipGabani: Updated the name. `allowedLocalHostFiles` is correct.

library/pod-security-policy/seccomp/template.yaml
JaydipGabani: I don't think so. I added the same in constraint - https://github.com/open-policy-agent/gatekeeper-library/pull/540/files#diff-58803b65572868312ed7364039674b1871f873f15cfd13aa29c8b9adb42e2901 - and added the examples to test `Localhost` profile. PTAL.

maxsmythe: Oops, missed it being added in a lower line

maxsmythe (changes requested):
One final nit, then LGTM

library/pod-security-policy/seccomp/template.yaml
```diff
@@ -67,215 +67,332 @@ spec:
...
+              })
+          - name: podLocalHostProfile
+            expression: |
+              variables.hasPodSeccomp && has(variables.anyObject.spec.securityContext.seccompProfile.localhostProfile) ? variables.anyObject.spec.securityContext.seccompProfile.localhostProfile : ""
+          - name: canonicalPodSecurityContextProfile
+            expression: |
+              has(variables.hasPodSeccomp) && has(variables.anyObject.spec.securityContext.seccompProfile.type) ? 
+                (variables.anyObject.spec.securityContext.seccompProfile.type == "RuntimeDefault" ? (
+                  variables.allowedProfiles.exists(profile, profile == "runtime/default") ? "runtime/default" : variables.allowedProfiles.exists(profile, profile == "docker/default") ? "docker/default" : "runtime/default") : 
+                variables.anyObject.spec.securityContext.seccompProfile.type == "Unconfined" ? "unconfined" : variables.anyObject.spec.securityContext.seccompProfile.type == "Localhost" ? "localhost/" + variables.anyObject.spec.securityContext.seccompProfile.localhostProfile : "")
```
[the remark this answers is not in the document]
JaydipGabani: Updated to use the variable.

library/pod-security-policy/seccomp/template.yaml
JaydipGabani: No worries! Good catch! I updated the code.

JaydipGabani:
@ritazh @maxsmythe Fixed the code, should be good now.

maxsmythe (approved):
LGTM

src/pod-security-policy/seccompv2/constraint.tmpl
```diff
@@ -0,0 +1,73 @@
+apiVersion: templates.gatekeeper.sh/v1
+kind: ConstraintTemplate
+metadata:
+  name: k8spspseccompv2
+  annotations:
+    metadata.gatekeeper.sh/title: "Seccomp V2"
+    metadata.gatekeeper.sh/version: 1.0.0
+    description: >-
+      Controls the seccomp profile used by containers. Corresponds to the
+      `securityContext.seccompProfile` field.
```
ritazh, line 10: might be good to explicitly mention:
```suggestion
      `securityContext.seccompProfile` field. Security contexts from the annotation is not considered as Kubernetes no longer reads security contexts from the annotation.
```

ritazh (approved):
LGTM

one nit

src/pod-security-policy/seccompv2/constraint.tmpl
JaydipGabani: Updated the description.

Issue #541, opened 2024-06-03
Add CEL code for PSP Policies in library


JaydipGabani:
**To-Do list for each policy:**

- [ ] Add `src.cel` file for the policy under `src/pod-security-policy/<name>/`

- [ ] Modify `constraint.tmpl` to add CEL engine and move rego under rego engine
```
  targets:
    - target: admission.k8s.gatekeeper.sh
      code: 
      - engine: K8sNativeValidation
        source:
{{ file.Read "src/pod-security-policy/<name>/src.cel" | strings.Indent 10 | strings.TrimSuffix "\n" }}
      - engine: Rego
        source:
          rego: |
{{ file.Read "src/pod-security-policy/<name>/src.rego" | strings.Indent 12 | strings.TrimSuffix "\n" }}
          libs:
            - |
{{ file.Read "src/pod-security-policy/<name>/lib_exempt_container.rego" | strings.Indent 14 | strings.TrimSuffix "\n" }}
```
- [ ] Bump minor version on `constraint.tmpl` by updating `metadata.gatekeeper.sh/version` annotation.
- [ ] Run `make generate-all` to generate all relavent files
- [ ] Run `make verify-gator-dockerized POLICY_ENGINE=cel && make verify-gator-dockerized POLICY_ENGINE=rego` to test changes in local

**PSP Policies list to track migration**

- [x] validation/allow-privilege-escalation
- [ ] validation/apparmor
- [ ] validation/capabilities
- [ ] validation/flexvolume-drivers
- [ ] validation/forbidden-sysctls
- [ ] validation/fsgroup
- [ ] validation/host-filesystem
- [ ] validation/host-namespaces
- [x] validation/host-network-ports
- [x] validation/privileged-containers
- [x] validation/proc-mount
- [x] validation/read-only-root-filesystem
- [ ] validation/seccomp
- [ ] validation/selinux
- [ ] validation/users
- [ ] validation/volumes

PR #542, opened 2024-06-03
chore: adding cel for psp-root-filesystem

JaydipGabani:

Fixes #541 


library/pod-security-policy/read-only-root-filesystem/template.yaml
```diff
@@ -37,67 +37,104 @@ spec:
...
+            expression: |
+              (variables.containers + variables.initContainers + variables.ephemeralContainers).filter(container,
+                !(container.image in variables.exemptImages) && 
+                (!has(container.securityContext) ||
+                !has(container.securityContext.readOnlyRootFilesystem) ||
+                container.securityContext.readOnlyRootFilesystem != true)
+              ).map(container, container.name)
+          validations:
+          - expression: '(has(request.operation) && request.operation == "UPDATE") || size(variables.badContainers) == 0'
+            messageExpression: '"only read-only root filesystem container is allowed: " + variabales.badContainers.join(", ")'
```
maxsmythe, line 73: `variables` is misspelled.

maxsmythe (changes requested):

library/pod-security-policy/read-only-root-filesystem/template.yaml
ritazh: why didnt CI/tests fail?

JaydipGabani: It seems that CEL was returning with the message for not recognizing the misspelled variable, but `gator` somehow treated it as violation (which it was expecting for disallow case). And since respective `suite,yaml` didn't specify expectation of specific message in case of violation, the error didn't get caught in CI.

JaydipGabani: I fixed the variable and added verification of violation message in `suite.yaml`

ritazh (approved):
LGTM

library/pod-security-policy/read-only-root-filesystem/template.yaml
maxsmythe: If there is an error executing `messageExpression`, VAP returns a default message instead.

maxsmythe (approved):
LGTM

PR #548, opened 2024-06-07
chore: Upgrade kind to v0.23.0; use k8s 1.30.0 for integration tests

maxsmythe:

We need k8s 1.30.0 so we can test constraints/templates against the newest apparmor behavior, where the apparmor profile is described in the securityContext as opposed to in an annotation.

Fixes #


library/pod-security-policy/proc-mount/samples/psp-proc-mount/example_allowed.yaml
```diff
@@ -1,10 +1,11 @@
 apiVersion: v1
 kind: Pod
 metadata:
-  name: nginx-proc-mount-disallowed
```
ritazh, line 4: I think you need to run 'make generate generate-website-docs generate-artifacthub-artifacts' to re-generate the templates and docs for this change

maxsmythe: Thanks! I thought I ran `make generate-all` I guess not :/

Hopefully this fixes the tests

maxsmythe: Tests are still failing

@sozercan @ritazh @JaydipGabani Any idea why the test workflow is not picking up my changes?

Here is the error:

```
# actual: The Pod "nginx-proc-mount-disallowed" is invalid: spec.containers[0].securityContext.procMount: Invalid value: "Unmasked": `hostUsers` must be false to use `Unmasked`
```

This should not be encountered given the changes made

maxsmythe: Made tests more verbose. Looks like the changes were taking effect.

Digging through the k8s code, it looks like UserNamespacesSupport feature gate needs to be enabled for `hostUsers` config to be honored.

ritazh (approved):
LGTM

PR #578, opened 2024-08-19
fix(k8sdisallowanonymous): correct no allowed roles

### LINKED ISSUE #575: Instance of k8sdisallowanonymous with no parameters yields no violations
With [this template](https://github.com/open-policy-agent/gatekeeper-library/blob/master/library/general/disallowanonymous/template.yaml) and the following constraint:

```
kind: K8sDisallowAnonymous
metadata:
  name: avoid-system-bindings
spec:
  match:
    kinds:
      - apiGroups: ["rbac.authorization.k8s.io"]
        kinds: ["ClusterRoleBinding"]
      - apiGroups: ["rbac.authorization.k8s.io"]
        kinds: ["RoleBinding"]
```

The following object does not generate a violation in my suite.yaml:

```
apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRoleBinding
metadata:
  name: test-crb
roleRef:
  apiGroup: rbac.authorization.k8s.io
  kind: ClusterRole
  name: test-crole
subjects:
- apiGroup: rbac.authorization.k8s.io
  kind: Group
  name: system:unauthenticated
```

If I then update the constraint to have an empty `allowedRoles`:

```
kind: K8sDisallowAnonymous
metadata:
  name: avoid-system-bindings
spec:
  match:
    kinds:
      - apiGroups: ["rbac.authorization.k8s.io"]
        kinds: ["ClusterRoleBinding"]
      - apiGroups: ["rbac.authorization.k8s.io"]
        kinds: ["RoleBinding"]
  parameters:
    allowedRoles: []
```
 It generates a violation.  This _should_ be a no-op, so it seems like the logic in the template has a bug.

apeabody:

Fixes #575


JaydipGabani (approved):
lgtm

artifacthub/library/general/disallowanonymous/1.0.1/samples/no-anonymous-bindings/constraint.yaml
```diff
@@ -0,0 +1,14 @@
...
+metadata:
+  name: no-anonymous
+spec:
+  match:
+    kinds:
+      - apiGroups: ["rbac.authorization.k8s.io"]
+        kinds: ["ClusterRoleBinding"]
+      - apiGroups: ["rbac.authorization.k8s.io"]
+        kinds: ["RoleBinding"]
+  parameters:
```
julianKatz, line 12: Should we also have a version of the constraint featured in the suite that specifies no `allowedRoles`, as that was the issue we are trying to fix?

julianKatz (approved):

artifacthub/library/general/disallowanonymous/1.0.1/samples/no-anonymous-bindings/constraint.yaml
apeabody: Hi @julianKatz - I added 4 new test with no `allowedRoles ` to the rego unit tests: https://github.com/open-policy-agent/gatekeeper-library/pull/578/files#diff-f0536bb898d20b404acf2fce2427efc3874d5ad60e3e1873583247139cc8de2e

I suppose we could also add another to the gator test suite, but I'm not sure we can exclude that from inclusion in the documentation.

maxsmythe (approved):
LGTM

PR #579, opened 2024-08-21
feat(k8sdisallowanonymous): allow disallowing system:authenticated

### LINKED ISSUE #561: Could we include `system:authenticated` in the set of disallowed subjects in k8sdisallowanonymous template?
Currently, the [`k8sdisallowanonymous` template](https://github.com/open-policy-agent/gatekeeper-library/blob/master/library/general/disallowanonymous/template.yaml) disallows bindings to the following two subjects:

https://github.com/open-policy-agent/gatekeeper-library/blob/625ca8b0eed3aead24939fc95adefad029fae275/library/general/disallowanonymous/template.yaml#L42-L48

**I believe `system:authenticated` belongs in this list as well.**

Rationale from https://orca.security/resources/blog/sys-all-google-kubernetes-engine-risk/:

_The loophole, which we dubbed Sys:All, stems from a likely widespread misconception that the system:authenticated group in Google Kubernetes Engine includes only verified and deterministic identities, whereas in fact, it includes any Google authenticated account (including outside the organization). This misunderstanding then creates a significant security loophole when administrators unknowingly bind this group with overly permissive roles._

Users that are seeking to prevent access by users outside their organization are unlikely to care if those users are logged-in to gmail (and thus are system:authenticated) or logged out (system:unauthenticated).

julianKatz:
Previously, the k8sdisallowanonymous would prevent bindings to the
system:anonymous and system:unauthenticated groups.

For Google Kubernetes Engine, the system:authenticated group is a
potential security threat.  This was described by Orca Security in this
blog post:

https://orca.security/resources/blog/sys-all-google-kubernetes-engine-risk/

This PR updates the k8sdisallowanonymous to prevent bindings to
`system:authenticated` if the parameters.disallowAuthenticated toggle is
set to true.  This will not break existing customers as the default
boolean value is false, leaving this functionality disabled.

Fixes #561


src/general/disallowanonymous/src.rego
```diff
@@ -2,18 +2,51 @@ package k8sdisallowanonymous
 
 violation[{"msg": msg}] {
   not is_allowed(input.review.object.roleRef, input.parameters.allowedRoles)
-  review(input.review.object.subjects[_])
+
+  disallowAuthenticated := object.get(input, ["parameters", "disallowAuthenticated"], false)
+  review(input.review.object.subjects[_], disallowAuthenticated)
+
   msg := sprintf("Unauthenticated user reference is not allowed in %v %v ", [input.review.object.kind, input.review.object.metadata.name])
```
apeabody, line 9: nit: stray space

```suggestion
  msg := sprintf("Unauthenticated user reference is not allowed in %v %v", [input.review.object.kind, input.review.object.metadata.name])
```

src/general/disallowanonymous/src_test.rego
```diff
@@ -222,6 +222,18 @@ test_allowed_multiple_role_multiple_subjects_mix_rolebinding {
...
+test_authenticated_with_disallow_authenticated {
+    inp := {"review": rolebinding([{"name": "system:authenticated", "kind": "Group"}, {"name": "system:anonymous", "kind": "User"}], "role-2"), "parameters": { "allowedRoles": [], "disallowAuthenticated": true }}
+    results := violation with input as inp
+    count(results) == 1
+}
+
+test_authenticated_with_disallow_authenticated_false {
+    inp := {"review": rolebinding([{"name": "system:authenticated", "kind": "Group"}, {"name": "system:anonymous", "kind": "User"}], "role-2"), "parameters": { "allowedRoles": [], "disallowAuthenticated": false }}
+    results := violation with input as inp
+    count(results) == 0
```
apeabody, line 234: This should still raise a violation due to the `system:anonymous` subject?

src/general/disallowanonymous/src_test.rego
```diff
@@ -222,6 +222,18 @@ test_allowed_multiple_role_multiple_subjects_mix_rolebinding {
     count(results) == 0
 }
 
+test_authenticated_with_disallow_authenticated {
+    inp := {"review": rolebinding([{"name": "system:authenticated", "kind": "Group"}, {"name": "system:anonymous", "kind": "User"}], "role-2"), "parameters": { "allowedRoles": [], "disallowAuthenticated": true }}
+    results := violation with input as inp
+    count(results) == 1
+}
+
+test_authenticated_with_disallow_authenticated_false {
```
apeabody, line 231: I would add a test with no `"disallowAuthenticated"` parameter at all.

src/general/disallowanonymous/src.rego
```diff
@@ -2,18 +2,51 @@ package k8sdisallowanonymous
 
 violation[{"msg": msg}] {
   not is_allowed(input.review.object.roleRef, input.parameters.allowedRoles)
```
apeabody, line 4: We should get this PR merged first: https://github.com/open-policy-agent/gatekeeper-library/pull/578

src/general/disallowanonymous/constraint.tmpl
```diff
@@ -24,6 +24,11 @@ spec:
               type: array
               items:
                 type: string
+            disallowAuthenticated:
```
apeabody, line 27: Optionally you could also include `default: false`, while also implemented in the Rego, makes it clearer to users the default behavior.

apeabody (commented):
Hey @julianKatz - Don't forget to also bump this to `1.1.0`: https://github.com/julianKatz/gatekeeper-library/blob/disallow-authenticated/src/general/disallowanonymous/constraint.tmpl#L7

src/general/disallowanonymous/src.rego
apeabody: Merged!

src/general/disallowanonymous/constraint.tmpl
julianKatz: Good call, included.

src/general/disallowanonymous/src_test.rego
julianKatz: There are other tests above that lack the disallowAuthenticated parameter.  I can dedup a bit to make that more obvious.

src/general/disallowanonymous/src_test.rego
julianKatz: The review object is the `authenticated_rolebinding`, which only has `system:authenticated` as a subject.

src/general/disallowanonymous/src.rego
```diff
@@ -2,18 +2,51 @@ package k8sdisallowanonymous
...
+
   msg := sprintf("Unauthenticated user reference is not allowed in %v %v ", [input.review.object.kind, input.review.object.metadata.name])
 }
 
 is_allowed(role, allowedRoles) {
   role.name == allowedRoles[_]
 }
 
-review(subject) = true {
+review(subject, disallowAuthenticated) = true {
```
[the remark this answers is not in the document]
julianKatz: Code has changed to make this obsolete

src/general/disallowanonymous/src.rego
julianKatz: Good catch, fixed.

src/general/disallowanonymous/src_test.rego
julianKatz: Added test case.

src/general/disallowanonymous/src.rego
```diff
@@ -2,18 +2,51 @@ package k8sdisallowanonymous
...
   subject.name == "system:unauthenticated"
 }
 
-review(subject) = true {
+review(subject, disallowAuthenticated) = true {
```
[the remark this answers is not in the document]
julianKatz: Changed code structure so this is obsolete.

src/general/disallowanonymous/src_test.rego
apeabody: To clarify I meant no parameter object at all, but with rebasing on https://github.com/open-policy-agent/gatekeeper-library/pull/578 that should also be covered.

website/docs/validation/disallowanonymous.md
```diff
@@ -16,7 +16,7 @@ metadata:
   name: k8sdisallowanonymous
```
apeabody, line 16: Running `make generate-all` will regenerate this and the other docs, etc.

artifacthub/library/general/disallowanonymous/1.0.1/artifacthub-pkg.yml
```diff
@@ -0,0 +1,22 @@
+version: 1.0.1
```
apeabody, line 1: FYI: I'm guessing this `1.0.1` folder is stale from earlier, and probably should be removed.

apeabody: I see you just did!

src/general/disallowanonymous/constraint.tmpl
```diff
@@ -24,6 +24,12 @@ spec:
...
+              description: >-
+                A boolean indicating whether `system:authenticated` should be
```
apeabody, line 29: Just a syntax suggestion (as this doesn't replace the other validations)

```suggestion
                A boolean indicating whether `system:authenticated` should also be
```

src/general/disallowanonymous/src.rego
```diff
@@ -1,19 +1,34 @@
...
-review(subject) = true {
-  subject.name == "system:unauthenticated"
+subject_is(subject, expected) {
+  subject.name == expected
 }
 
-review(subject) = true {
-  subject.name == "system:anonymous"
+message(group) := val {
+  val := sprintf("%v is not allowed as a subject name in %v %v", [group, input.review.object.kind, input.review.object.metadata.name])
```
apeabody, line 33: Just calling out that by including `group` we are slightly changing the original behavior.  As it's just the message I don't think it a breaking change, but wanted to note that is some cases it will increases the number of violations as they are now group specific.

library/general/disallowanonymous/suite.yaml
```diff
@@ -14,4 +14,16 @@ tests:
...
+    assertions:
+    - violations: no
+- name: disallow-authenticated
+  template: template.yaml
+  constraint: samples/no-authenticated/constraint.yaml
+  cases:
+    - name: authenticated-disallowed-with-parameter-true
+      object: samples/no-authenticated/example_disallowed.yaml
+      assertions:
+      - violations: yes
```
apeabody, line 29: You might want to consider also asserting the exact expected violation message(s) for higher test coverage.

```suggestion
      - violations: yes
      - message: "{message}"
         violations: 1
```

src/general/disallowanonymous/src.rego
```diff
@@ -1,19 +1,34 @@
 package k8sdisallowanonymous
 
+
```
apeabody, line 3: nit: causes extra line return in the template/docs

```suggestion
```

apeabody (approved):
Thanks for the contribution @julianKatz!

Made a few suggestions, and this will still need to be reviewed by another maintainer.

apeabody:
FYI: Probably a fresh `make generate-all` will resolve the CI generate error

src/general/disallowanonymous/src.rego
julianKatz: Good eye!

I confirmed with Max that going from 1 to multiple violations is not a breaking change.

My thinking was that before:

1. Someone with `system:anonymous` as the subject would get a message saying that `Unuthenticated` was the problem.

2. Someone with a binding to _both_ `system:anonymous` AND `system:unauthenticated` would get only a single violation.  Once they fix that violation, they'd then get _another_ violation for the other binding.  This UX is confusing.

apeabody: Yup, agreed, and makes sense to me!

library/general/disallowanonymous/suite.yaml
julianKatz: Nice!  Didn't know about this feature.  I've improved the suite test by confirming the exact messages that should be returned.

src/general/disallowanonymous/constraint.tmpl
julianKatz: Good catch ty

maxsmythe (approved):
LGTM

sozercan (approved):
LGTM

PR #583, opened 2024-08-26
fix(k8srequiredlabels): CEL broke when allowedRegex was empty

julianKatz:
The CEL implementation of this policy always expected the `allowedRegex` value to be set for a given `key` value in the contraint parameters.

Existing users of the template could (via rego) require only that a given label key exist, but assert nothing about the value of that label.

This change makes the CEL implementation uphold that same contract.


maxsmythe (approved):
LGTM

apeabody (approved):
Thanks for catching @julianKatz!

This highlights the value of the Rego engine unit tests (e.g. https://github.com/open-policy-agent/gatekeeper-library/blob/master/src/general/requiredlabels/src_test.rego), as they provide far more code coverage than just the universal integration tests(https://github.com/open-policy-agent/gatekeeper-library/blob/master/library/general/requiredlabels/suite.yaml).

JaydipGabani (approved):

PR #584, opened 2024-08-26
fix(k8spspreadonlyrootfilesystem): exemptImages CEL bug

julianKatz:
The CEL implementation of this policy's `exemptImages` variable was evaluated as a list of objects (containers), then being used later in the policy as if it were a list of strings.  This prevented the wildcard path from functioning.

This PR adds suite test cases that evaluate wildcards in `k8spspreadonlyrootfilesystem`.  These fail without the fix.
This PR then also updates the CEL implementation to return a list of strings for `exemptImages`.

Other templates with `K8sNativeValidation` implementations have this same problem. but will be addressed in separate PRs.

PR #588, opened 2024-08-28
fix(k8spspprocmount): exemptImages CEL bug

julianKatz:
I recently found (#584) that some K8sNativeValidation implementations of certain templates that iterate over and exempt containers by image had a bug preventing the exemption logic from working.

I've fixed that bug here by mapping from `container` struct to `container.image` string.  I've also added a suite test to verify this. That case fails without the change to the CEL logic.


artifacthub/library/pod-security-policy/proc-mount/1.1.1/samples/psp-proc-mount/example_disallowed_exempt_image.yaml
JaydipGabani, line 1: This should be renamed to `example_allowed_exempt_image.yaml` as we are not expecting it to generate violation. Renaming the file should also fix the CI.

JaydipGabani (changes requested):

artifacthub/library/pod-security-policy/proc-mount/1.1.1/samples/psp-proc-mount/example_disallowed_exempt_image.yaml
julianKatz: Updated this to be an `example_allowed` as you suggested

julianKatz:
Thanks for the review @JaydipGabani!  This is ready for another look.

JaydipGabani (approved):
lgtm

apeabody (approved):
LGTM

maxsmythe (approved):
LGTM

Markieta:
@julianKatz I just pulled this update and it seems to break functionality getting the following violations:

```
failed expression: (has(request.operation) && request.operation == "UPDATE") || size(variables.badContainers) == 0
```

PR #589, opened 2024-08-29
fix(k8spsphostnetworkingports): CEL fixes for hostNetwork variable and message

julianKatz:
My updates to the suite.yaml file yielded an unexpected failure due to an incorrect CEL expression:

```
        unexpected number of violations: got 1 violations but want none: got messages [failed expression: (has(request.operation) && request.operation == "UPDATE") ||
(!has(variables.params.hostNetwork) || !variables.params.hostNetwork ? (has(variables.anyObject.spec.hostNetwork) && !variables.anyObject.spec.hostNetwork) : true)]
```

By contrast, a run of `gator verify -v .
--enable-k8s-native-validation=false` yielded a fully passing suite.yaml.

This failed expression was actually failing due to its `messageExpression`, as non-primitive types cannot be combined with strings as in some interpreted languages (like rego).  Unfortunately the compiler does not indicate that the messageExpression is the source of the problem.

Once the message was fixed, I resolved the incorrect violation expression to fix the bug in the handling of params.hostNetwork.


library/pod-security-policy/host-network-ports/template.yaml
JaydipGabani, line 1: Update patch version to `1.1.3`

JaydipGabani (changes requested):

library/pod-security-policy/host-network-ports/template.yaml
julianKatz: Bumped the patch version and re-generated.  Ready for another look!

apeabody (approved):
Thanks @julianKatz!

maxsmythe (approved):
LGTM

JaydipGabani (approved):

PR #591, opened 2024-08-30
fix(k8spspprivilegedcontainer): exemptImages CEL bug

julianKatz:
I recently found (#584) that some K8sNativeValidation implementations of certain templates that iterate over and exempt containers by image had a bug preventing the exemption logic from working.

I've fixed that bug here by mapping from container struct to container.image string. I've also added a suite test to verify this. That case fails without the change to the CEL logic.


maxsmythe (approved):
LGTM

apeabody (approved):

ritazh (approved):
LGTM

sozercan (approved):

PR #547, opened 2024-06-05
chore: adding CEL for psp-host-filesystem

JaydipGabani:
**What this PR does / why we need it**:

**Which issue(s) does this PR fix** *(optional, using `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when the PR gets merged)*:
Fixes #541 

**Special notes for your reviewer**:

PR #595, opened 2024-09-04
fix(k8sPSPHostFilesystem): null-check on volumes

julianKatz:
A K8sNativeValidation implementation of this template was added in #547.

When testing it, I found that a Pod lacking the `volumes` field would yield a null-pointer style error on the CEL expression:

```
unexpected number of violations: got 1 violations but want none: got messages [expression '(has(request.operation) && request.operation == "UPDATE") || size(variables.badHostPaths) == 0' resulted in error: composited variable "badHostPaths" fails to evaluate: composited variable "volumes" fails to evaluate: no such key: volumes]
```

This PR adds a `has(` check to prevent that null pointer, and adds a suite test case that fails without the code change.


JaydipGabani (approved):
LGTM, Thanks!

maxsmythe (approved):
LGTM

PR #596, opened 2024-09-07
chore(k8spspcapabilities): double newlines can cause yamllint breakage

julianKatz:
The default value for newlines in the yamllint linter (a commonly used tool for linting yaml) is 2.  See
https://yamllint.readthedocs.io/en/stable/rules.html#module-yamllint.rules.empty_lines for more info.

This PR removes some unnecessary newlines in a rego file that trigger this linter rule.


apeabody (approved):

apeabody (approved):
Thanks @julianKatz!

I think this is good to merge once all the tests are green.

PR #599, opened 2024-09-19
fix: use newer seccompProfile spec in mutation

### LINKED ISSUE #195: Support spec.securityContext.seccompProfile
In Kubernetes [v1.19](https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG/CHANGELOG-1.19.md#seccomp-graduates-to-general-availability), seccomp graduated to GA which meant the seccomp annotations (`seccomp.security.alpha.kubernetes.io/pod: runtime/default`) were deprecated:

```
W0514 13:03:16.997735   41392 warnings.go:70] spec.template.metadata.annotations[container.seccomp.security.alpha.kubernetes.io/manager]: 
deprecated since v1.19; use the "seccompProfile" field instead
```

The "k8spspseccomp" mutating constraint should probably be updated to check both the annotation and the new field, at least until v1.25 comes out: https://github.com/open-policy-agent/gatekeeper-library/blob/259ad1bdd8945d6df451d3fc2867109d76b854e2/mutation/pod-security-policy/seccomp/samples/mutation.yaml

This issue and #188 makes me think we might need to a mechanism for creating constraints that do different things depending on the k8s version.

sathieu:
Older is ignored

Example mutation has no effect.

Fixes #195


maxsmythe (approved):
LGTM, thanks for the PR!

JaydipGabani (approved):
LGTM, Thanks!

ritazh (approved):

PR #618, opened 2024-12-27
chore: adding vap testing

### LINKED ISSUE #574: create a new set of tests with vap generation
existing tests only use gk for enforcement, we should create new tests that generate VAP. this should be scoped to CEL policies and k8s > 1.30+

JaydipGabani:

Fixes #574 


Makefile
```diff
@@ -36,9 +36,9 @@ deploy:
 ifeq ($(POLICY_ENGINE), rego)
```
ritazh, line 36: Can you add a comment here that only the first engine in the template gets used for evaluation UNLESS enableK8sNativeValidation=false which ensures the subsequent non-K8sNativeValidation engine gets used?

Makefile
```diff
@@ -36,9 +36,9 @@ deploy:
...
 	helm install -n gatekeeper-system gatekeeper gatekeeper/gatekeeper --create-namespace --version $(GATEKEEPER_VERSION) --set enableK8sNativeValidation=false
 else ifeq ($(POLICY_ENGINE), cel)
-ifneq ($(GATEKEEPER_VERSION), 3.15.1)
 	helm install -n gatekeeper-system gatekeeper gatekeeper/gatekeeper --create-namespace --version $(GATEKEEPER_VERSION) --set enableK8sNativeValidation=true
-endif
+else ifeq ($(POLICY_ENGINE), vap)
```
ritazh, line 40: For these, since Gatekeeper webhook is a fallback for VAP, how do we ensure the failure resulted from VAP instead of the GK webhook?

Makefile
ritazh, line 36: In general, I think because we are not actually testing the error message, it's hard to tell which engine caused the violation and which enforcement point caused the failure. Not sure how hard it is to add that, could be a follow up issue/PR if you want to track it.

JaydipGabani: Since GK does not have fallback between engines, violations are thrown through CEL engine if it is enabled and CT has CEL. Otherwise the source of violation is rego engine.

Makefile
JaydipGabani: We check the violation string to make sure it includes `ValidatingAdmissionPolicy` - https://github.com/open-policy-agent/gatekeeper-library/pull/618/files#diff-3b0266d06db240d302bf9febd17c09cf104dccce0bb2ee8e799e49c85f7082ffR96.

Makefile
ritazh: > Since GK does not have fallback between engines, violations are thrown through CEL engine if it is enabled and CT has CEL. Otherwise the source of violation is rego engine.

yea thats how it works today but if we ever expose priority like the issue you opened, then we need to test the actual violation message. maybe open an issue to track for future?

ritazh (approved):
LGTM

Makefile
JaydipGabani: In this case, we may want to modify the violation message to include engine information as well. Since as of now the violation message is the same regardless of the engine used to evaluate CTs. IMO, I don't think users apart from CT authors would care about which engine is being used to enforce policies. And CT authors would only care to verify the logic written for policy, which I think can be attained with `gator` to test CT.

maxsmythe (approved):
LGTM

JaydipGabani:
Not stale

PR #639, opened 2025-03-26
Rego v1 and Regal compliance

anderseknert:
This is rewrite of the entire Rego library for Rego v1 compliance, and the introduction of Regal to identify issues and enforce good coding standards.

To not be too disruptive, all policies were first copied to a `v1` directory under `src`, and no modifications to existing policies have been made. The Regal configuration is also set to ignore anything but the new v1 directory.

The result is about 1200 lines less Rego in the v1 directory compared to the previous versions.

There's still a bunch of practicalities to deal with here, I'm sure. Like how to integrate this in the project CI, publishing, and so on. The templates are currently just copied from the "old" directories, but should probably not follow the same versioning, and might need something to identify them as Rego v1 policies. And so on, and so forth :)

But I figured a PR would be the best way to start that discussion, and for me to get some directions for what to work on now.

The test.sh file have been updated to run the unit tests against the v1 policies too, and they're all passing. I have not yet played with the Gator CLI tests though, so that might be a good thing to look into as well.

### DIFF
--- a/.regal/config.yaml
+++ b/.regal/config.yaml
@@ -1,0 +1,42 @@
+rules:
+  custom:
+    prefer-value-in-head:
+      level: error
+      only-scalars: true
+  idiomatic:
+    directory-package-mismatch:
+      level: ignore
+    no-defined-entrypoint:
+      level: ignore
+  imports:
+    prefer-package-imports:
+      level: ignore
+  style:
+    avoid-get-and-list-prefix:
+      level: ignore
+    external-reference:
+      level: ignore
+    file-length:
+      ignore:
+        files:
+          - "*_test.rego"
+    line-length:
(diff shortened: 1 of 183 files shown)


ritazh:
Please update the constraint template to opt-in to rego v1 by explicitly adding:
```yaml
  targets:
      code:
      - engine: Rego
        source:
          version: v1
          rego: |
```

anderseknert:
Thanks @ritazh, I'll get to it first thing tomorrow 🙂

anderseknert:
Hi @JaydipGabani 👋 I kinda forgot about it 😅 But  I'mhappy to pick this up again when I'm back from vacation next week. Let's sync here or on Slack then.

PR #713, opened 2026-01-04
feat: Add NodePort namespace range policy

elhamafzalizadeh:
## Gatekeeper Policy: K8sNodePortNamespaceRange

This is a **Gatekeeper constraint** that restricts the NodePort ranges based on team namespaces. Its purpose is to prevent accidental or uncoordinated NodePort assignments in the cluster.

### Policy Details

- **apiVersion:** `constraints.gatekeeper.sh/v1beta1`  
- **kind:** `K8sNodePortNamespaceRange`  
- **metadata.name:** `enforce-nodeport-namespace-range`  
- **enforcementAction:** `deny` (any Service violating this policy will be denied)


JaydipGabani (changes requested):
@elhamafzalizadeh you will need to restructure and add the policy here - https://github.com/open-policy-agent/gatekeeper-library/tree/master/src/general/allowedreposv2, while keeping the examples and suits in the library/general. Please follow the structure similar to other policies.

PR #715, opened 2026-01-08
pod security profiles

### LINKED ISSUE #6: Organize PSP policies into standardized buckets
**Describe the solution you'd like**
There are quite a few PSP policies in the library. It might be confusing to users which ones to deploy and what constraints parameters to set.

sig-auth provides a set of standardized PSP definitions (https://docs.google.com/document/d/1d9c4BaDzRw1B5fAcf7gLOMZSVEvrpSutivjfNOwIqT0/edit#heading=h.ihgl7qlgtyzu) that we can align on, ranging from unrestricted to common best practices to restricted.  

We should create guidance and buckets for users to adopt PSP policies easily, and mention that these can be customized by users.

sozercan:

Fixes #6 

### DIFF
--- a/.github/workflows/generate-catalog.yaml
+++ b/.github/workflows/generate-catalog.yaml
@@ -83,10 +83,11 @@ jobs:
           gator policy update
           gator policy search ""
 
-          # Verify bundles are present
+          # Verify bundles are present in catalog
           echo ""
           echo "=== Bundles ==="
-          gator policy list --bundles
+          grep -n "^bundles:" catalog.yaml
+          grep -q "^bundles:" catalog.yaml
 
       - name: Upload catalog artifact
         uses: actions/upload-artifact@ea165f8d65b6e75b540449e92b4886f43607fa02 # v4.6.2
--- a/catalog.yaml
+++ b/catalog.yaml
@@ -1,7 +1,7 @@
 # Policy Catalog for Gatekeeper Library
 #
 # This file is auto-generated. Do not edit directly.
-# Generated at: 2026-02-12T19:15:38Z
+# Generated at: 2026-02-12T20:16:33Z
 #
 # To regenerate, run:
 #   gator policy generate-catalog --library-path=. --output=catalog.yaml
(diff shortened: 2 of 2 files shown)


src/pod-security-policy/seccomp/constraint.tmpl
```diff
@@ -4,7 +4,7 @@ metadata:
   name: k8spspseccomp
   annotations:
     metadata.gatekeeper.sh/title: "Seccomp"
-    metadata.gatekeeper.sh/version: 1.1.0
```
JaydipGabani: why is this being bumped? should bundle annotations be added here as well?

src/pod-security-policy/flexvolume-drivers/constraint.tmpl
JaydipGabani: This policy/volume type is not part of any PSS profile - `https://kubernetes.io/docs/concepts/security/pod-security-standards/#restricted`. We shold remove this from PSR bundle.

src/pod-security-policy/fsgroup/constraint.tmpl
JaydipGabani: This policy/fsgroup is not part of any PSS profile - https://kubernetes.io/docs/concepts/security/pod-security-standards/#restricted. We shold remove this from PSR bundle.

scripts/website/sidebars-template.js
JaydipGabani: how do we display PSP policies that are not part of PSS here? (e.g flexvolume, fsgroup)

artifacthub/library/pod-security-policy/apparmor/1.1.0/template.yaml
JaydipGabani: current template and version should not be updated, its only the next version that is added. Applies everywhere else too.

JaydipGabani (changes requested):
Thanks for raising this.

We are currently missing two PSS policies - 
- HostProcess (Windows)
- Host Probes/Lifecycle Hooks (v1.34+)

Both of these policies should be in both the bundle. Here is the pr for the same - https://github.com/open-policy-agent/gatekeeper-library/pull/717

